PR #57, opened 2016-10-12
Use bundler to manage CocoaPods and xcpretty

benasher44:

- Travis appears to be using CocoaPods 1.1.0.beta.2, which is missing some fixes for Xcode 8 (see [sample failing build](https://travis-ci.org/Instagram/IGListKit/jobs/166935850) from #51).
- This change will ensure that a consistent CocoaPods version is used by Travis
- In the added Gemfile, I picked the latest CocoaPods 1.1.0 RC (matches version in Podfile.lock) and the latest xcpretty.
- Changed `pod spec lint` to `pod lib lint` to verify local files instead of files from the version specified in the spec.

## Pull request checklist

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -48,3 +48,7 @@ playground.xcworkspace
 # Carthage/Checkouts
 
 Carthage/Build
+
+# Bundler
+.bundle
+vendor
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,7 +21,10 @@ env:
    - DESTINATION="OS=9.3,name=iPad Air"       SDK="$IOS_SDK" SCHEME="$IOS_SCHEME" RUN_TESTS="YES" BUILD_EXAMPLE="NO"  POD_LINT="NO"
    - DESTINATION="OS=10.0,name=iPhone 7"      SDK="$IOS_SDK" SCHEME="$IOS_SCHEME" RUN_TESTS="YES" BUILD_EXAMPLE="YES" POD_LINT="NO"
 
-install: bundle install --path vendor/bundle
+# Do this in before_install, so that gems are installed
+# before travis prints package version info and avoid a
+# warning about missing gems.
+before_install: bundle install --path vendor/bundle
 
 script:
 - if [ $POD_LINT == "YES" ]; then
--- a/.travis.yml
+++ b/.travis.yml
@@ -24,7 +24,7 @@ env:
 
 script:
 - if [ $POD_LINT == "YES" ]; then
-      bundle exec pod spec lint;
+      bundle exec pod lib lint;
   fi
 
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -29,13 +29,13 @@ script:
 
 
 - if [ $BUILD_EXAMPLE == "YES" ]; then
-      xcodebuild clean build -workspace Example/IGListKitExamples.xcworkspace -scheme IGListKitExamples -sdk "$SDK" -destination "$DESTINATION" ONLY_ACTIVE_ARCH=NO CODE_SIGNING_REQUIRED=NO | bundle exec xcpretty -c;
+      xcodebuild clean build -workspace Example/IGListKitExamples.xcworkspace -scheme IGListKitExamples -sdk "$SDK" -destination "$DESTINATION" ONLY_ACTIVE_ARCH=NO | bundle exec xcpretty -c;
   fi
 
 
 - if [ $RUN_TESTS == "YES" ]; then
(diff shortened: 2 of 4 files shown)


benasher44:
Hm. I signed the cla for #51, and it won't let me sign again.

rnystrom:
@benasher44 don't worry about the CLA bot, once you signed once you should be g2g from then on.

Since I'm a TravisCI noob I'm gonna ask @jessesquires to take a quick look at this.

benasher44:
Ah yep it responded shortly after. Thanks!

benasher44:
Looks like the output still contains this after moving `bundle install` to `before_install`
```
Could not find proper version of cocoapods (1.1.0.rc.3) in any of the sources
Run `bundle install` to install missing gems.
```
This build still passes, so this is mostly just annoying because `pod --version` will always output this. Going to try using the `gemfile` setting, which is also valid for objective-c projects.

jessesquires:
this LGTM except travis is still not happy

jessesquires:
Wow. still waiting on travis. First build has timed out.

Also:

```
Could not find proper version of cocoapods (1.1.0.rc.3) in any of the sources
```

😢

benasher44:
Yeah… I'm not sure what else to do to make that go away. It's trying to print `pod --version` before it runs `bundle install`, which seems like a Travis bug at this point (literally does it on the next line). It knows there's a Gemfile, so it really shouldn't try to print info about gems before they're installed. Maybe it doesn't expect that we're going to install a different version of the CocoaPods gem? In any case, printing versions of things before the `install` or `before_install` phases seems like a Travis bug. Filed travis-ci/travis-ci#6717.

As for the timeout, it looks like it's timing out during code signing. Code signing probably isn't needed here, since we just want to verify that it builds and tests pass. I'm going to try disabling that. I've tried this with travis before the Xcode 8 GM was released, and it caused issues. Maybe it's fixed now?

benasher44:
Luckily, even though that message is in red, it doesn't fail the build. Here's a passing one: https://travis-ci.org/Instagram/IGListKit/jobs/167060407

rnystrom:
Nice, I wonder if we can make any subscription updates to Travis to get higher pri queues...

@jessesquires think we're g2g here?

benasher44:
Are the timeouts a new issue in this PR or existing?

benasher44:
Ah looks like existing: https://travis-ci.org/Instagram/IGListKit/builds/167080863. This is good to go in terms of fixing `pod spec lint` flakes.

Quick question though. I just realized this is doing `pod spec lint` instead of `pod lib lint`. The former will go fully based on the spec (including pulling the spec's tag from git), whereas the latter always uses the files in the working directory. The latter seems like it be more useful for CI, since it'll verify new changes, whereas the former will always check the current version in the podspec. Should I change this?

benasher44:
Tried locally disabling code signing with CODE_SIGNING_ALLOWED=NO, and the tests failed to bootstrap every time. CODE_SIGNING_REQUIRED=NO seemed to work though, and it appeared to skip the signing step for tests. So maybe less hanging now?

benasher44:
Despite CODE_SIGNING_REQUIRED changes, there is still plenty of code signing happening. 🔥

This might be a red herring anyway, since the thing that happens right after code signing is starting the tests. It might just be hanging waiting for the simulator.

Anyway, I think I'm out of things I want to try in this PR.

jessesquires:
> Anyway, I think I'm out of things I want to try in this PR.

😆  Thanks for all the help so far @benasher44 ! 🙌 

I've been fighting with travis-ci timeouts since the upgrade to Xcode8. No idea whats going on, but all of my builds for other projects timeout for every single iOS 8 build. /shrug

jessesquires:
😕 
https://twitter.com/travisci/status/783389212825489408

benasher44:
No prob! There are a few tickets floating around about Xcode 8 Travis things, but I hadn't seen a response yet. So, good to see a response from Travis!

benasher44:
Early on, I found that only the iPhone SE destination worked reliably on Travis. This may or may not still be thee case. It might be worth adding that one in or replacing one of the existing destinations, so that if that or another seems consistent, at least you have one simulator that always works to keep PRs going.

benasher44:
There's still the unanswered question of `pod spec lint` vs `pod lib lint`. I could make this small change here as well (sort of related because ruby/CocoaPods?). I could also just file an issue for someone to dig into later.

benasher44:
Ah thought previous comment had a thumbs up, but it was the one before that. Let me know, and I'll revert.

jessesquires:
Sounds good to me! 👍

rnystrom:
Sweet, I think we're g2g here. We can resolve Travis infra timeout issues down the road if they come up.

benasher44:
Yup! I agree.

Issue #68, opened 2016-10-14
Conversion from value of type NSIndexPath* to incompatible type IGListMoveIndexPath*


valeriyvan:
XCode's static analyser shows issue:
/Users/someuser/somepath/IGListKit/Source/IGListDiff.mm:179:13: Conversion from value of type 'NSIndexPath *' to incompatible type 'IGListMoveIndexPath *'
## New issue checklist

- [*] I have reviewed the [`README`](https://github.com/Instagram/IGListKit/blob/master/README.md) and [documentation](http://instagram.github.io/IGListKit)
- [*] I have searched [existing issues](https://github.com/Instagram/IGListKit/issues) and this is not a duplicate

### General information

- Library version(s):
1.0.0
- iOS version(s):
Not relevant
- Devices/Simulators affected:
Not relevant
- Reproducible in the demo project? (Yes/No):
Not relevant
- Related issues:
None

rnystrom:
Interesting, I think this is a false-positive from the static analyzer.

- A move is generated (for an index _or_ index path) [here](https://github.com/Instagram/IGListKit/blob/master/Source/IGListDiff.mm#L245)
- There is a value for param `obj`, so it adds the object to the collection [here](https://github.com/Instagram/IGListKit/blob/master/Source/IGListDiff.mm#L175-L176)

It looks like we need to address the static analyzer though so we can have a clean analysis.

rnystrom:
Closed as of #69

PR #69, opened 2016-10-14
Simplifies block ridding of one if-else branch met only once.

valeriyvan:
Simplifies block ridding of one if-else branch met only once. 
Solves issue #68.

## Pull request checklist

- [*] All tests pass. Demo project builds and runs.
- [*] I have reviewed the [contributing guide](https://github.com/Instagram/IGListKit/blob/master/CONTRIBUTING.md)


rnystrom:
Awesome! I just put up a diff internally to do the same fix but didn't remove the last param from the block. This is 👌 great stuff!

valeriyvan:
That block was too dynamic even for Objective-C.

PR #72, opened 2016-10-16
Attempt to fix travis.yml

jessesquires:
We're still having travis issues. In other projects, I have issues with iOS 8 timeouts, but nothing else that we've seen here (`pod lint` problems, iOS 9 and 10 timeouts).

Theory: the *simulator* is the problem?

Using `iPhone 6/s` seems to be passing consistently. (like [here](https://travis-ci.org/Instagram/IGListKit/builds/167933436)). So let's try always using this sim.


jessesquires:
Wow. The builds started immediately.

jessesquires:
iOS 8.1 build failed 😢 

But it looks like the others are going to pass! 🎉 (At least they didn't timeout at the very start)

rnystrom:
I wonder if this is an Xcode 8 issue? I tried running the test suite on an 8.2 simulator and it just hangs. I can't even pause to see what is going on, seems like the whole thing is stuck.

jessesquires:
> I wonder if this is an Xcode 8 issue?

Yep, I think that's part of it actually. Someone mentioned this on twitter but I can't find the thread.

rnystrom:
@jessesquires I fired another tweet to @travis-ci but haven't heard anything yet https://twitter.com/_ryannystrom/status/787731537613774848

Should we lang this and carry on?

jessesquires:
Ok. 

So only 9.1, 9.2, 9.3 passed -- even though they reported 4 test failures. WTF.
https://travis-ci.org/Instagram/IGListKit/jobs/167979267

All other builds timed out.

"Continous integration". Something is continuous, clearly not "functionality".

rnystrom:
Found the reason for the failures (#78). Lessons learned when fixing up a PR internally while importing.

Though travis saying tests succeeded is scary.

jessesquires:
Made a few more changes:

- comment out iOS 8.x for now (since it just hangs)
- only build on iOS 9.x (don't run tests for each `.x`)
- set `CODE_SIGNING_REQUIRED=NO`

if we can get a least some of these to run consistently we can get *some signal*

jessesquires:
current status:

screen shot 2016-10-18 at 3 02 26 pm

awww yiissss

jessesquires:
Ok. So Xcode 8 hangs on running tests in iOS 8 simulators. (lolwut) 

Changes:

- We **only build** for iOS 8.x — not run tests.
- Build example for 8.1, 9.0, 10.0 only
- Run tests for 9.0 and 10.0 only

Seems ok to me, because:

- We're building all on supported iOS versions, so this should catch any compile errors and invalid API errors (deprecations, unavailable, etc.)
- Still running tests for iOS 9.0 and 10.0 -- should catch most things, I think. Would be nice to run on all versions but everything is 🔥  so 🤗

rnystrom:
Nice, I think that's fine.

So 8.1, 9.0, and 10.0 are still timing out on the sample app?

jessesquires:
does the sample app hang on building locally for you? @rnystrom

jessesquires:
Oh. The example project schemes weren't shared. That's why travis was failing on building the examples.

jessesquires:
Everything should pass now 🎉

jessesquires:
travis is green! 🎉 🌮

rnystrom:
🌮 🌮 🌮 🌮

PR #74, opened 2016-10-16
Makes objects function more swifty.

valeriyvan:
Makes objects function more swifty.

## Pull request checklist

- [x ] All tests pass. Demo project builds and runs.
- [ x] I have reviewed the [contributing guide](https://github.com/Instagram/IGListKit/blob/master/CONTRIBUTING.md)

### DIFF
--- a/Example/IGListKitExamples/ViewControllers/MixedDataViewController.swift
+++ b/Example/IGListKitExamples/ViewControllers/MixedDataViewController.swift
@@ -71,13 +71,8 @@ class MixedDataViewController: UIViewController, IGListAdapterDataSource {
     //MARK: IGListAdapterDataSource
 
     func objects(for listAdapter: IGListAdapter) -> [IGListDiffable] {
-        var items = [IGListDiffable]()
-        for obj in data {
-            if selectedClass == nil || type(of: obj) == selectedClass! {
-                items.append(obj as! IGListDiffable)
-            }
-        }
-        return items
+        guard selectedClass != nil else { return data.map { $0 as! IGListDiffable } }
+        return data.filter { type(of: $0) == selectedClass! }.map { $0 as! IGListDiffable }
     }
 
     func listAdapter(_ listAdapter: IGListAdapter, sectionControllerFor object: Any) -> IGListSectionController {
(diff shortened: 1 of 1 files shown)


rnystrom:
@jessesquires had some comments on the last swifty PR we did after I merged it, gonna let him take a quick look at this before we import.

Example/IGListKitExamples/ViewControllers/MixedDataViewController.swift
```diff
@@ -71,13 +71,8 @@ class MixedDataViewController: UIViewController, IGListAdapterDataSource {
...
 
     func objects(for listAdapter: IGListAdapter) -> [IGListDiffable] {
-        var items = [IGListDiffable]()
-        for obj in data {
-            if selectedClass == nil || type(of: obj) == selectedClass! {
-                items.append(obj as! IGListDiffable)
-            }
-        }
-        return items
+        guard selectedClass != nil else { return data.map { $0 as! IGListDiffable } }
```
jessesquires: nit: I think this would be easier to read if we put the `return` on a newline and the final `}` on a newline

Example/IGListKitExamples/ViewControllers/MixedDataViewController.swift
```diff
@@ -71,13 +71,8 @@ class MixedDataViewController: UIViewController, IGListAdapterDataSource {
...
+        return data.filter { type(of: $0) == selectedClass! }.map { $0 as! IGListDiffable }
```
jessesquires: actually, now i remember the issues I saw with this. scratch that first comment. 😄 

the previous version combined the 2 conditions in a single `if`. Now this is broken up and harder to follow, not to mention we've written `.map { $0 as! IGListDiffable }` twice. 😊 

why can't the initial filter contain both conditions? couldn't this be something like this? 

```swift
return data.filer { 
        if let class = selectedClass { 
            return type(of: $0) == class 
        } 
        return false
        }.map { $0 as! IGListDiffable }
```

I think this is more readable + more concise.

valeriyvan: Sure, initial filter could have both conditions. 
But it hides that if selectedClass is nil, function returns just it's input array with transformed elements.
Otherwise it filters input array and then transforms it's elements.
Version I am proposing keeps that clear.
Less condensed but more clear.

PR #81, opened 2016-10-17
Using array initialisation instead of for loop.

valeriyvan:
Using array initialisation instead of for loop.

## Pull request checklist

- [X] I have reviewed the [contributing guide](https://github.com/Instagram/IGListKit/blob/master/CONTRIBUTING.md)

### DIFF
--- a/Example/IGListKitExamples/ViewControllers/SingleSectionViewController.swift
+++ b/Example/IGListKitExamples/ViewControllers/SingleSectionViewController.swift
@@ -23,7 +23,7 @@ final class SingleSectionViewController: UIViewController, IGListAdapterDataSour
     
     let collectionView = IGListCollectionView(frame: CGRect.zero, collectionViewLayout: UICollectionViewFlowLayout())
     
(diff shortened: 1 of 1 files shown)


Example/IGListKitExamples/ViewControllers/SingleSectionViewController.swift
```diff
@@ -23,17 +23,13 @@ final class SingleSectionViewController: UIViewController, IGListAdapterDataSour
     
     let collectionView = IGListCollectionView(frame: CGRect.zero, collectionViewLayout: UICollectionViewFlowLayout())
     
-    var data = [Int]()
+    lazy var data: [Int] = { return Array(0..<20) }()
```
jessesquires: I don't think `lazy` buys us that much here. Let's remove that to simplify. 😊

valeriyvan: Agree. That was my first version, actually.

jessesquires:
LGTM! 🎉

PR #82, opened 2016-10-17
Using switch with is cases instead of if with multiple as?.

valeriyvan:
Using switch with is cases instead of if with multiple as?.

## Pull request checklist


jessesquires:
Thanks @valeriyvan ! 👍

Issue #39, opened 2016-10-02
Storyboard support


rnystrom:
I thought that storyboards would work out-of-the-box, but they don't. The adapter will attempt to register the cells, which probably writes over the cell registered as a prototype in the storyboard.

I think all we need to do is provide a dequeue method specifically for storyboards. This would skip checking and registering the cells. Something like:

```objc
- (__kindof UICollectionViewCell *)dequeueReusableCellFromStoryboardWithIdentifier:(NSString *)identifier forIndexPath:(NSIndexPath *)indexPath;
```

Note that we don't need to provide a cell class, unlike the current dequeue method, because the identifier is set in the storyboard.

zhubofei:
@rnystrom Shall we close this issue?

PR #90, opened 2016-10-20
Support supplementaryViews created from nibs

rawlinxx:
## Changes in this pull request

Summary: add dequeueReusableSupplementaryView with nib

PR #92, opened 2016-10-20
Add storyboard support #39

zhubofei:

This PR is for #39

## Pull request checklist


rnystrom:
Wo! This and #90 arriving at the same time are so awesome. Thanks for taking this on! 💪 

This solution is spot on, but we're going to need to add some tests. An example in the examples app would be awesome, but we could add that later. Take a look at #56 for an idea of what all we need:

- Unit tests creating section controllers from a storyboard
- _Optional_ example
- Add to `CHANGELOG.md`
- A new initializer for `IGListSingleSectionController`
  - See https://github.com/Instagram/IGListKit/blob/ce4b851da343940513a074342fab6f5a7e67125d/Source/IGListSingleSectionController.h#L74-L77
  - Note that we'll need unit tests for this too!

zhubofei:
@rnystrom Is a new  initializer for `IGListSingleSectionController` necessary?  
Something like
``` objective-c
- (instancetype)initWithStoryboardCollectionView:(UICollectionView *)collectionView
                         identifier:(NSString *)identifier
                 configureBlock:(IGListSingleSectionCellConfigureBlock)configureBlock
                      sizeBlock:(IGListSingleSectionCellSizeBlock)sizeBlock;
```
I'm not sure if sending in a weak ref of collectionView is a good idea. 🤔

rnystrom:
@zhubofei if we don't add a new init, how could we use `IGListSingleSectionController` with storyboards?

We shouldn't need to pass the collection view to the section controller, right? Something like

```objc
- (instancetype)initWithStoryboardIdentifier:(NSString *)identifier 
  configureBlock:(IGListSingleSectionCellConfigureBlock)configureBlock
  sizeBlock:(IGListSingleSectionCellSizeBlock)sizeBlock;
```

should be fine I think.

Also looks like there are some merge conflicts w/ the example project. Probably want to rebase on master.

zhubofei:
@rnystrom I think the collectionView could be prototyped in the storyboard as well? If we init the collectionView again in the SingleSectionController, we will lose the constraints attached to the original one, right?

rnystrom:
@zhubofei oh the collection view should _definitely_ be created from the storyboard too. Though you would just make the collection view subclass `IGListCollectionView` **in the storyboard**. Then just

```
@property @IBOutlet IGListCollectionView *collectionView;

// viewDidLoad
self.adapter.collectionView = self.collectionView;
```

That'll use the storyboard-generated collection view for everything then.

Make sense?

zhubofei:
Right! Sorry, I was just being silly 😂.

Source/IGListSingleSectionController.h
```diff
@@ -77,6 +77,22 @@ IGLK_SUBCLASSING_RESTRICTED
...
+ @param sizeBlock      A block that returns the size for the cell given the collection context.
+ 
+ @return A new section controller.
+ 
+ @warning Be VERY CAREFUL not to create retain cycles by holding strong references to: the object that owns the adapter
+ (usually "self") or the IGListAdapter. Pass in locally scoped objects or use weak references!
+ */
+- (instancetype)initWithStoryboardCellIdentifier:(NSString *)identifier
+                              configureBlock:(IGListSingleSectionCellConfigureBlock)configureBlock
+                                   sizeBlock:(IGListSingleSectionCellSizeBlock)sizeBlock;
```
rnystrom: nit: mind aligning the colons?

Source/IGListSingleSectionController.m
```diff
@@ -53,6 +54,20 @@ - (instancetype)initWithNibName:(NSString *)nibName
     return self;
 }
 
+- (instancetype)initWithStoryboardCellIdentifier:(NSString *)identifier
+                              configureBlock:(IGListSingleSectionCellConfigureBlock)configureBlock
+                                   sizeBlock:(IGListSingleSectionCellSizeBlock)sizeBlock {
+    IGParameterAssert(identifier != nil);
```
rnystrom: Let's assert `identifier.length > 0`

Source/IGListSingleSectionController.m
```diff
@@ -53,6 +54,20 @@ - (instancetype)initWithNibName:(NSString *)nibName
...
+    IGParameterAssert(configureBlock != nil);
+    if (self = [super init]) {
+        _identifier = identifier;
```
rnystrom: `_identifier = [identifier copy];`

Source/IGListSingleSectionController.m
```diff
@@ -53,6 +54,20 @@ - (instancetype)initWithNibName:(NSString *)nibName
     return self;
 }
 
+- (instancetype)initWithStoryboardCellIdentifier:(NSString *)identifier
+                              configureBlock:(IGListSingleSectionCellConfigureBlock)configureBlock
+                                   sizeBlock:(IGListSingleSectionCellSizeBlock)sizeBlock {
+    IGParameterAssert(identifier != nil);
+    IGParameterAssert(configureBlock != nil);
```
rnystrom: Need an assert for `sizeBlock` too

Source/IGListSingleSectionController.m
```diff
@@ -72,6 +87,10 @@ - (UICollectionViewCell *)cellForItemAtIndex:(NSInteger)index {
                                                           bundle:self.bundle
                                             forSectionController:self
                                                          atIndex:index];
+    } else if ([self.identifier length] > 0) {
```
rnystrom: This is fine for now, but we probably want to turn this into a state enum and avoid too many ifs. Don't worry about it here, just noting to myself.

Source/IGListSingleSectionController.m
```diff
@@ -72,6 +87,10 @@ - (UICollectionViewCell *)cellForItemAtIndex:(NSInteger)index {
...
+        cell = [collectionContext dequeueReusableCellFromStoryboardWithIdentifier:self.identifier
+                                            forSectionController:self
+                                                         atIndex:index];
```
rnystrom: nit: colon alignment

Tests/Objects/IGTestSingleStoryboardViewController.m
```diff
@@ -0,0 +1,43 @@
+//
+//  IGTestSingleStoryboardViewController.m
+//  IGListKit
+//
+//  Created by Bofei Zhu on 10/20/16.
+//  Copyright © 2016 Instagram. All rights reserved.
+//
```
rnystrom: Needs our copyright comments

Tests/Objects/IGTestStoryboardCell.h
```diff
@@ -0,0 +1,16 @@
+//
+//  IGTestStoryboardCell.h
+//  IGListKit
+//
+//  Created by Bofei Zhu on 10/20/16.
+//  Copyright © 2016 Instagram. All rights reserved.
+//
```
rnystrom: ++copyright comments

Tests/Objects/IGTestStoryboardCell.m
```diff
@@ -0,0 +1,13 @@
+//
+//  IGTestStoryboardCell.m
+//  IGListKit
+//
+//  Created by Bofei Zhu on 10/20/16.
+//  Copyright © 2016 Instagram. All rights reserved.
+//
```
rnystrom: ++

Tests/Objects/IGTestStoryboardCell.m
```diff
@@ -0,0 +1,13 @@
...
+
+#import "IGTestStoryboardCell.h"
+
+@implementation IGTestStoryboardCell
+
```
rnystrom: Probably need to init the label at least? 🤔

Source/IGListCollectionContext.h
```diff
@@ -109,6 +109,21 @@ NS_ASSUME_NONNULL_BEGIN
...
+ @param sectionController The section controller requesting this information.
+ @param index             The index of the cell.
+ 
+ @return A cell dequeued from the reuse pool or newly created.
+ 
+ @note This method uses a string representation of the cell class as the identifier.
+ */
+- (__kindof UICollectionViewCell *)dequeueReusableCellFromStoryboardWithIdentifier:(NSString *)identifier
+                                                              forSectionController:(IGListSectionController <IGListSectionType> *)sectionController
+                                                                           atIndex:(NSInteger)index;
```
jessesquires: I see this added to the `.h` file, but not `IGListCollectionContext.m`

... am I missing something?? 🤔

jessesquires:
Also need to update the `CHANGELOG.md` -- add an entry under the `2.0.0` release 😄

Source/IGListSingleSectionController.m
zhubofei: @rnystrom What about the other two init funcions `initWithCellClass:` & `initWithNibName:` Do you want me to add  assertions for sizeBlock to them as well?

Source/IGListCollectionContext.h
zhubofei: @jessesquires  There seems to be no implementation for IGListCollectionContext. 😶

rnystrom: Yup just a protocol

Source/IGListCollectionContext.h
```diff
@@ -109,6 +109,21 @@ NS_ASSUME_NONNULL_BEGIN
                                                           atIndex:(NSInteger)index;
 
 /**
+ Dequeues a cell from the UICollectionView reuse pool.
```
rnystrom: > Dequeues a **storyboard prototype** cell from the UICollectionView reuse pool.

Source/IGListSingleSectionController.h
```diff
@@ -77,6 +77,22 @@ IGLK_SUBCLASSING_RESTRICTED
...
+ @return A new section controller.
+ 
+ @warning Be VERY CAREFUL not to create retain cycles by holding strong references to: the object that owns the adapter
+ (usually "self") or the IGListAdapter. Pass in locally scoped objects or use weak references!
+ */
+- (instancetype)initWithStoryboardCellIdentifier:(NSString *)identifier
+                                  configureBlock:(IGListSingleSectionCellConfigureBlock)configureBlock
+                                       sizeBlock:(IGListSingleSectionCellSizeBlock)sizeBlock;
+
+/**
```
rnystrom: This is great!

Source/IGListSingleSectionController.m
```diff
@@ -53,6 +54,20 @@ - (instancetype)initWithNibName:(NSString *)nibName
     return self;
 }
 
+- (instancetype)initWithStoryboardCellIdentifier:(NSString *)identifier
+                                  configureBlock:(IGListSingleSectionCellConfigureBlock)configureBlock
+                                       sizeBlock:(IGListSingleSectionCellSizeBlock)sizeBlock {
+    IGParameterAssert(identifier != nil);
```
rnystrom: `identifier.length > 0`

Source/IGListStackedSectionController.m
```diff
@@ -208,6 +208,16 @@ - (UICollectionViewCell *)dequeueReusableCellWithNibName:(NSString *)nibName
                                                                                           atIndex:(index + offset)];
 }
 
+- (UICollectionViewCell *)dequeueReusableCellFromStoryboardWithIdentifier:(NSString *)identifier
+                                                     forSectionController:(IGListSectionController <IGListSectionType> *)sectionController
+                                                                  atIndex:(NSInteger)index
+{
```
rnystrom: nit: curly needs to be on the same line as the last parameter

Tests/Objects/IGTestStoryboardCell.m
rnystrom: ++

Tests/Objects/IGTestStoryboardCell.m
```diff
@@ -11,4 +11,10 @@
 
 @implementation IGTestStoryboardCell
 
+- (id)initWithCoder:(NSCoder *)aDecoder {
+    self = [super initWithCoder:aDecoder];
+    self.label.text = @"";
+    return self;
+}
```
rnystrom: Oh lol shoot, been living in code too much. I missed the `@IBOutlet`. Do you mind removing this? 😭

zhubofei:
@rnystrom There is a problem with Uint tests, though. It seems that the collectionView would ignore the call of `cellForItemAtIndexPath`. I think it should be a layout  problem under test environment rather than a bug with the source code, because, as you can see the single section example code runs just fine. I have tried to figure out what went wrong for hours without any luck.

Tests/Objects/IGTestStoryboardCell.m
zhubofei: Sure :)

zhubofei:
iOS Conf SG explains a lot! Actually, I was having the impression that engineers at Instagram work until 3AM. That's hardcore and inspiring. 😅

rnystrom:
I'll have to give the tests a try then and see what's going on. We def want everything supported and tested with storyboards that we can do today in code. Not sure exactly what the problem is tho.

zhubofei:
@rnystrom I finally figured out what goes wrong. It seems that the UIViewController has this lazy loading behavior. The root VC's view should be added to the UIWindow before layout subviews.

The only tests that fail are
``` objective-c
XCTAssertEqual(cell1.frame.size.width, 100);
XCTAssertEqual(cell2.frame.size.width, 100);
XCTAssertEqual(cell3.frame.size.width, 100);
```

I wonder if we should keep these tests or not. Because storyboard cells inherit the width from root VC. Or should we change 100 to viewController.view.frame.size.width?

rnystrom:
@zhubofei weird, ya let's not test the inner workings of storyboards. Feel free to remove those tests since they are covered already.

Source/IGListCollectionContext.h
jessesquires: lol. derp.

implemented in `IGListAdapter` 🙃

Example/IGListKitExamples/ViewControllers/SingleSectionStoryboardViewController.swift
```diff
@@ -0,0 +1,80 @@
...
+        super.viewDidLoad()
+        
+        view.addSubview(collectionView)
+        adapter.collectionView = collectionView
+        adapter.dataSource = self
+    }
+    
+    override func viewDidLayoutSubviews() {
+        super.viewDidLayoutSubviews()
+        collectionView.frame = view.bounds
```
JakeLin: If we use Storyboard, can we use Auto Layout or autoresizing? Then we don't need this line.

Example/IGListKitExamples/ViewControllers/SingleSectionStoryboardViewController.swift
```diff
@@ -0,0 +1,80 @@
...
+    }
+    
+    func listAdapter(_ listAdapter: IGListAdapter, sectionControllerFor object: Any) -> IGListSectionController {
+        let configureBlock = { (data: Any, cell: UICollectionViewCell) in
+            guard let cell = cell as? StoryboardCell, let number = data as? Int else { return }
+            cell.textLabel.text = "Cell: \(number + 1)"
+        }
+        
+        let sizeBlock = { (context: IGListCollectionContext?) -> CGSize in
+            guard let context = context else { return CGSize() }
```
JakeLin: To make it more specific, I think `return .zero` is better.

Tests/Objects/IGTestStoryboardCell.m
JakeLin: Since it is done in Storyboard, I think we don't need to call something like `[[UILabel alloc] initWithFrame:CGRectMake(0.0, 0.0, width, height)]` to init here.

Example/IGListKitExamples/ViewControllers/SingleSectionStoryboardViewController.swift
zhubofei: @JakeLin Nevermind. I find out I set a wrong constraint😂. Will definitely remove this line.

Example/IGListKitExamples/SectionControllers/DemoSectionController.swift
```diff
@@ -53,7 +53,17 @@ class DemoSectionController: IGListSectionController, IGListSectionType {
...
+        if object?.name == "Storyboard" {
+            let storyboard = UIStoryboard(name: "Demo", bundle: nil)
+            let controller = storyboard.instantiateViewController(withIdentifier: "demo") as! StoryboardViewController
+            controller.title = object?.name
+            viewController?.navigationController?.pushViewController(controller, animated: true)
+        } else if object?.name == "Single Section Storyboard" {
+            let storyboard = UIStoryboard(name: "Demo", bundle: nil)
+            let controller = storyboard.instantiateViewController(withIdentifier: "singleSectionDemo") as! SingleSectionStoryboardViewController
+            controller.title = object?.name
+            viewController?.navigationController?.pushViewController(controller, animated: true)
```
zhubofei: I thought about this for some time before putting it this way. Because it really makes no sense to give the cell identifier the same string as VC title. So if we combine this two together, I think we need a dictionary for title->identifier, right?

zhubofei: Or shall we add a cellIdentifier property to the DemoItem class?

JakeLin:
@zhubofei I like the idea to have `controllerIdentifier` in `DemoSectionController` 👍

Example/IGListKitExamples/SectionControllers/DemoSectionController.swift
```diff
@@ -53,7 +56,12 @@ class DemoSectionController: IGListSectionController, IGListSectionType {
     }
 
     func didSelectItem(at index: Int) {
-        if let controller = object?.controllerClass.init() {
+        if let identifier = object?.controllerIdentifier {
+            let storyboard = UIStoryboard(name: "Demo", bundle: nil)
```
rnystrom: Noting to myself: this means that any demo using a storyboard will have to put their VCs in the `Demo.storyboard` file. I think that's fine for now.

Example/IGListKitExamples/SectionControllers/DemoSectionController.swift
```diff
@@ -19,13 +19,16 @@ class DemoItem: NSObject {
 
     let name: String
     let controllerClass: UIViewController.Type
+    let controllerIdentifier: String?
 
     init(
         name: String,
-        controllerClass: UIViewController.Type
+        controllerClass: UIViewController.Type,
+        controllerIdentifier: String?=nil
```
rnystrom: ultra nit: spaces `String? = nil`

Example/IGListKitExamples/ViewControllers/SingleSectionStoryboardViewController.swift
```diff
@@ -0,0 +1,71 @@
...
+    }
+    
+    // MARK: - IGListSingleSectionControllerDelegate
+    
+    func didSelect(_ sectionController: IGListSingleSectionController) {
+        let section = adapter.section(for: sectionController) + 1
+        let alert = UIAlertController(title: "Section \(section) was selected 🎉", message: nil, preferredStyle: .alert)
+        alert.addAction(UIAlertAction(title: "Dismiss", style: .default, handler: nil))
+        present(alert, animated: true, completion: nil)
+    }
```
rnystrom: This is awesome

Example/IGListKitExamples/ViewControllers/SingleSectionStoryboardViewController.swift
```diff
@@ -0,0 +1,71 @@
...
+        let sectionController = IGListSingleSectionController(storyboardCellIdentifier: "cell",
+                                                              configureBlock: configureBlock,
+                                                              sizeBlock: sizeBlock)
+        sectionController.selectionDelegate = self
+        return sectionController
+    }
+    
+    func emptyView(for listAdapter: IGListAdapter) -> UIView? {
+        return nil
+    }
```
rnystrom: style nit: other VC uses `{ return nil }`

Source/IGListAdapter.m
```diff
@@ -727,6 +727,18 @@ - (__kindof UICollectionViewCell *)dequeueReusableCellOfClass:(Class)cellClass
     return [collectionView dequeueReusableCellWithReuseIdentifier:identifier forIndexPath:indexPath];
 }
 
+- (__kindof UICollectionViewCell *)dequeueReusableCellFromStoryboardWithIdentifier:(NSString *)identifier
+                                                              forSectionController:(IGListSectionController <IGListSectionType> *)sectionController
+                                                                           atIndex:(NSInteger)index {
+    IGAssertMainThread();
+    IGParameterAssert(sectionController != nil);
+    IGParameterAssert(identifier != nil);
```
rnystrom: This should be `IGParameterAssert(identifier.length > 0);`

zhubofei:
AWESOME!

jessesquires:
🎉

JakeLin:
Great PR to unlock Storyboard

PR #101, opened 2016-10-21
Uses Set instead of Array for uniqueness.

valeriyvan:
Uses Set instead of Array for uniqueness.

## Pull request checklist


rnystrom:
Ha, nice! Super simple.

valeriyvan:
This array initialisation is great use case for AnyGenerator. But what can write so far is much longer of what we already have. Unfortunately.

Issue #77, opened 2016-10-16
IGListKitCrashes when used in an iMessages Extension


ishabazz:
## New issue checklist
- [ x] I have reviewed the [`README`](https://github.com/Instagram/IGListKit/blob/master/README.md) and [documentation](http://instagram.github.io/IGListKit)
- [x ] I have searched [existing issues](https://github.com/Instagram/IGListKit/issues) and this is not a duplicate
### General information
- Library version(s):1.0.0
- XCode:8.0
- iOS version(s):10.0.x
- Devices/Simulators affected:iPhone 7Plus
- Reproducible in the demo project? (Yes/No):Yes

Running iMessage Apps which have IGListKit included via CocoaPods results in the app crashing on launch with the error like the following:  dyld: Library not loaded: @rpath/IGListKit.framework/IGListKit
  Referenced from: /Users/demo/Library/Developer/CoreSimulator/Devices/5C343568-A532-405F-8B65-52A049827328/data/Containers/Bundle/Application/AA18B7D3-48C4-4958-A14C-C056A309D6D8/IGListStickerDemo.app/PlugIns/MessagesExtension.appex/MessagesExtension
  Reason: image not found

Check out  https://github.com/ishabazz/IGListKitMessagesExtensionDemo  for a sample project that exhibits this error.

rnystrom:
I've definitely seen this error before, but that was using a non-Swift library in a Swift app using Buck for dep management. Might be something similar though.

jessesquires:
Not clear if this is:

1. a bug in IGListKit
2. a bug in CocoaPods
3. a issue with project setup

rnystrom:
Off the bat I get an Xcode warning in the sample project:

> ld: warning: linking against a dylib which is not safe for use in application extensions: /Users/rnystrom/Library/Developer/Xcode/DerivedData/IGListStickerDemo-cherwvpgwwdqmjhkvyhkxnqnciyx/Build/Products/Debug-iphonesimulator/IGListKit/IGListKit.framework/IGListKit

rnystrom:
After `$ pod install` I see some weird CocoaPods warnings:

```
[!] The `IGListStickerDemo [Debug]` target overrides the `ALWAYS_EMBED_SWIFT_STANDARD_LIBRARIES` build setting defined in `Pods/Target Support Files/Pods-IGListStickerDemo/Pods-IGListStickerDemo.debug.xcconfig'. This can lead to problems with the CocoaPods installation
[...]

PR #112, opened 2016-10-23
Added example schemes for Today/iMessage extensions

Sherlouk:

Offering this open to a bit of a discussion based on the comments at #77.
- Created two new schemes (IGListKitTodayExample, IGListKitMessageExample) in which respectively correspond to their app extension.
- Updated the Podfile to ensure they have IGListKit as a dependency
- Added a super simple example reusing the `LabelSectionController` from the main app

Done more as a 'yes it can be done' as opposed to 'look at what can be done' hence the basic functionality but I can see ways to improve if necessary!
## Pull request checklist


Sherlouk:
Just as a secondary note I'm not sure how many of these changes are because it's a new scheme, I have a different version of CocoaPods or I simply made a mistake - revise checking! (Not big with CP)

Also my justification for iOS 10 deployment target is because of the OS changes, Apple completely revamped both of them in iOS 10 so they're changed a fair bit (Can't even do iMessage extension in older versions, though strangely it's allowed?). However, if it is desired that they match the deployment target of the library then I'm happy to make these changes!

Sherlouk:
@rnystrom I'm not sure how you guys would like to treat the extra schemes, whether you want a demo chooser similar to the main app or whether you want one example which shows that it's possible (probably with a bit more functionality than just a few cells) and leave the alternative examples to the main app?

A bit of steer would be appreciated!

cc: @jessesquires

Sherlouk:
@rnystrom @jessesquires 
Sorry! Bumping in order to get a bit of direction on this (see previous comment)

Would like to finish what I started here before moving onto some other example pack stuff!

rnystrom:
@Sherlouk Ok finally got a chance to poke around at this, and WOW. This is so cool! 🚀 

Some organizational nits:

- `MessageExtension` dir should be `IGListKitMessageExample`
- Same renaming for the iMessage target (haven't made iMessage apps before, hopefully this is simple)
- Rebase and use the icons from 920aad6a698034a0ca0a4a7a3e3f6fbce9522a31
  - Idk if we need new assets, if so I have them all I can upload here
- Bundle display name to "IGListKit" for all of them

@jessesquires check this out. cray:

simulator screen shot oct 27 2016 12 36 00 pm

simulator screen shot oct 27 2016 12 36 25 pm

jessesquires:
that's awesome 🎉

jessesquires:
should we add this to the changelog?

rnystrom:
@jessesquires good call, let's do that. cc @Sherlouk

Sherlouk:
@rnystrom Thanks! ❤️ 
I've updated the display names to IGListKit, and refactored the scheme name & directory to match the target.

The "Today Extension" icon comes from the base app, but annoyingly the message app icon is completely different (they're slightly rectangular) so you can't use the same ones as the main app! Maybe a job for your design guys?

Just seen the message about the changelog, I'll update that now!

Sherlouk:
Wasn't sure whether to just add to the bottom (as I did) or to add a new section "Example Pack Changes" which can be specifically for changes to the example pack to distinguish against modifications to the actual library.

Happy to modify upon feedback!

Sherlouk:
There was something deep down inside me which meant I just had to move that storyboard once I saw it out of place! 😅 

Also RIP travis?

CHANGELOG.md
```diff
@@ -18,6 +18,7 @@ This release closes the [2.0.0 milestone](https://github.com/Instagram/IGListKit
 - Fixed `-[IGListAdapter reloadDataWithCompletion:]` not returning early when `collectionView` or `dataSource` is nil and `completion` is nil. [Ben Asher](https://github.com/benasher44) [(#51)](https://github.com/Instagram/IGListKit/pull/51)
 - Added `-isFirstSection` and `-isLastSection` APIs to `IGListSectionController`
 - Added support for cells created from storyboard. [Bofei Zhu](https://github.com/zhubofei) [(#92)](https://github.com/Instagram/IGListKit/pull/92)
+- Added examples for Today & iMessage extensions. [Sherlouk](https://github.com/Sherlouk) [(#112)](https://github.com/Instagram/IGListKit/pull/112)
```
jessesquires: 👍

jessesquires:
ha 😄  travis should be working ok (but will probably take awhile)

Sherlouk:
Are we happy with the overall complexity of these examples?

Started with very basic functionality, but not sure if I should do something a bit more exciting! Open to suggestions there, or just to leave it as is

jessesquires:
@Sherlouk - I think this is good for this PR. Feel free to open a new issue to discuss! 😄

rnystrom:
Simple is good! More a proof of concept that this is possible w/ fairly minimal setup.

The other thing this shows off is the fact that you can build a single section controller and share it between different targets/implementations! Pretty slick.

@jessesquires should we make the `LabelSectionController` class that all 3 examples use live in a shared lib? Or is having the message and today widget use a section controller from the main example app ok?

Sherlouk:
While in a full application I'd probably agree with your suggestion there Ryan, I actually think in an example pack like this it's much easier to follow the chain of target membership if you can simply see the section controller can be used in multiple targets.

Just my two cents :)

jessesquires:
yeah, agree with @Sherlouk 💯

Sherlouk:
Builds are done! 🎉 🎉

Sherlouk:
Not at all sure how I've increased unit tests 🤔 

@rnystrom We waiting on anything specific to progress this?

rnystrom:
@sherlouk nope! Landing internally right now

PR #117, opened 2016-10-24
Center and trim badges, add love to readme

rnystrom:

- Trimmed badges that I think are a little unnecessary
- Added a heart to the "made by"

cc @jessesquires thoughts?

## Pull request checklist


jessesquires:
I still think the other badges are useful. That's the info a lot of people want at a glance.

"Is this iOS only?"

"Is this MIT license?"

"Does this work with carthage?"

rnystrom:
I dig it. I'm leaving out the license one tho since that shows up in the header.

Issue #65, opened 2016-10-12
NSObject equality naming conflict


iwasrobbed:
## Issue

Since this project extends `NSObject` to make it conform to `IGListDiffable` and that protocol has a method with a similar `isEqual` signature as the standard `NSObject` one, it is creating naming conflicts with other, completely unrelated, objects in the project that use the vanilla `isEqual` comparison.

<img width="1082" alt="screen shot 2016-10-12 at 3 42 44 pm" src="https://cloud.githubusercontent.com/assets/317671/19330319/98fd04c8-9092-11e6-946f-c46e6a144075.png">

You should consider uniquely naming the `IGListDiffable` method, e.g.:

``` objc
- (BOOL)isEqualToDiffableObject:(nullable id<IGListDiffable>)object;
```

I looked into opening a PR for this, but honestly there are too many vague `isEqual` calls in the codebase so I'm not sure which should be the diffable equality check and which should be the vanilla `NSObject` ones.
## New issue checklist
- [x] I have reviewed the [`README`](https://github.com/Instagram/IGListKit/blob/master/README.md) and [documentation](http://instagram.github.io/IGListKit)
- [x] I have searched [existing issues](https://github.com/Instagram/IGListKit/issues) and this is not a duplicate
### General information
- Library version(s): 1.0.0
- ~~iOS version(s):~~
- ~~Devices/Simulators affected:~~
- Reproducible in the demo project? (Yes/No): N/A
- ~~Related issues:~~

jessesquires:
Hey @iwasrobbed ! 😄 

It looks like your error is with `override`. Removing this should fix that. However, `isEqual:` changed in Swift 3. It should be `isEqual(object: Any?) -> Bool`. (Note: **not** `AnyObject`.) Are you using Swift 2.3 or 3.0? (We're only supporting 3.0)

iwasrobbed:
This is Swift 2.3 code, but the same issue would possibly exist in a Swift 3.0 project unless they made compiler fixes for these sorts of ambiguity.

This is actually existing code, not something I added after adding IGListKit, and the class it's on inherits from `NSObject` so it's technically an override. The class in question has nothing to do with IGListKit whatsoever which strikes me as side effects which you wouldn't want to have in people's projects.

Removing that just gives another error `Method 'isEqual' with Objective-C selector 'isEqual:' conflicts with method 'isEqual' from superclass 'NSObject' with the same Objective-C selector`

jessesquires:
[...]

PR #120, opened 2016-10-25
Update README.md

jessesquires:
A few minor tweaks to clarify things

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,10 +22,17 @@ A data-driven `UICollectionView` framework for building fast and flexible lists.
 
 `IGListKit` is built and maintained by [Instagram engineering](https://engineering.instagram.com/), using the open source version for the Instagram app.
 
+## Requirements
+
+- Xcode 8.0+
+- iOS 8.0+
+- Interoperability with Swift 3.0+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -22,10 +22,17 @@ A data-driven `UICollectionView` framework for building fast and flexible lists.
 
 `IGListKit` is built and maintained by [Instagram engineering](https://engineering.instagram.com/), using the open source version for the Instagram app.
 
+## Requirements
+
+- Xcode 8.0+
+- iOS 8.0+
+- Interoperability with Swift 3.0+
```
Sherlouk: I personally didn't run into any issues running it in a Swift 2.3 codebase

jessesquires: Yeah -- everything should actually work here 😄 

However, all of our examples are Swift 3 and we're only testing on Swift 3. So I think we should only claim support for Swift 3+

cc @rnystrom - thoughts?

rnystrom: Ya I think we should call out that we only support 3.0. I think there _are_ problems on 2.3 tho (see #65).

PR #122, opened 2016-10-25
reword some assertions with more informations

romankl:
## Changes in this pull request
This commit changes some assertion texts. No behavior is changed; no assertions are added (just reworded).  Some of them were short or contained no real meaningful informations, where the error probably is.

Ref: #113

## Pull request checklist

- [x] All tests pass. Demo project builds and runs.
- [x] I added tests, an experiment, or detailed why my change isn't tested.
- [x] I have reviewed the [contributing guide](https://github.com/Instagram/IGListKit/blob/master/CONTRIBUTING.md)

PR #123, opened 2016-10-26
Support supplementaryViews created from storyboard

zhubofei:
Miss this in the previous PR #92. Will add unit test soon.


rnystrom:
@zhubofei nice catch! Before we land this, can we add some unit tests to make sure they are working? Looks like you added views to the test storyboard, just need like one test and we're g2g.

zhubofei:
@rnystrom I'm working on that right now and should finish by the end of the day. I do notice one thing, though. In the source code, all the *ItemControllers have been renamed to *SectionController. But in unit tests, they stay with *Item*. Do you keep their name on purpose?

rnystrom:
@zhubofei ha, leftover from naming refactor. Surprised those never showed up! Yes we should move them in a separate PR. #126

zhubofei:
A simple test with a lot of pain. Finally. 😂

zhubofei:
@rnystrom Should I test that annoying 0.07% in StackSectionController? 😂

Source/IGListAdapter.m
```diff
@@ -778,6 +778,21 @@ - (__kindof UICollectionReusableView *)dequeueReusableSupplementaryViewOfKind:(N
...
+                                                                             withIdentifier:(NSString *)identifier
+                                                                       forSectionController:(IGListSectionController <IGListSectionType> *)sectionController
+                                                                                    atIndex:(NSInteger)index {
+    IGAssertMainThread();
+    IGParameterAssert(elementKind.length > 0);
+    IGParameterAssert(identifier.length > 0);
+    IGParameterAssert(sectionController != nil);
+    IGParameterAssert(index >= 0);
+    UICollectionView *collectionView = self.collectionView;
+    IGAssert(collectionView != nil, @"Reloading adapter without a collection view.");
```
rnystrom: nit: We just updated our asserts in #122, can we make this assert read something like this?

https://github.com/Instagram/IGListKit/blob/master/Source/IGListAdapter.m#L720

rnystrom:
@zhubofei 😂  no use, don't worry about it.

Small nit then we're g2g on this!

rnystrom:
@zhubofei it looks like `IGListKit.xcodeproj/project.pbxproj` got out of sync. Mind rebasing and fixing that up?

zhubofei:
@rnystrom Just did. I think we're g2g.

PR #125, opened 2016-10-27
Improve Objective-C property attributes

JakeLin:
In this PR, we changed all properties with the type like `NSSet`, `NSString`, `NSArray` or Block to `copy`. 

## Pull request checklist

### DIFF
--- a/Source/IGListBatchUpdateData.h
+++ b/Source/IGListBatchUpdateData.h
@@ -34,22 +34,22 @@ IGLK_SUBCLASSING_RESTRICTED
 /**
  Clean section moves.
  */
-@property (nonatomic, strong, readonly) NSSet<IGListMoveIndex *> *moveSections;
+@property (nonatomic, copy, readonly) NSSet<IGListMoveIndex *> *moveSections;
 
 /**
  Clean item insert index paths.
  */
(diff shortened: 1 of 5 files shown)


jessesquires:
Nice! Thanks @JakeLin ! 🎉

rnystrom:
These are actually all designed w/out copy-on-write on purpose. The returned collections aren't copies anyways, and most of these properties are `readonly`, so the copy never actually happens.

The `[nibName copy]` and `IGKAlgorithm` typo is a good catch tho. The rest IMO we should leave alone.

jessesquires:
@rnystrom - Right, so that just means these `copy`s will always be no-ops...

JakeLin:
@rnystrom, thanks for your feedback, when I was working on it, I had similar concerns, most of them are readonly and in the .m file , if we implicitly call '_iVar = [var copy];' in 'init'', then we shouldn't have any problems. My preference is to make them 'copy' , then we won't forget to call copy in 'init' and we don't have to remember to call copy for mutable version of the type when using 'self.array = mutableArray' inside the implementation file.

I will close the PR and make a small one on the typo and '[nibName copy]'.

JakeLin:
close because of #131

Issue #97, opened 2016-10-20
How to deal with JSON data?


yusuftor:
Hi,
Firstly, I think this IGListKit is awesome and so useful!

This is more of a question than an issue, but I have JSON data returned from my servers that I want to populate my list with. Usually I would use [SwiftyJSON](https://github.com/SwiftyJSON/SwiftyJSON) to handle the data, and everytime `-cellForItemAt` is called, use something like `data[indexPath.row]["text"].stringValue` to retrieve the right part of the data and input it into the cell. However, I'm trying to figure out how to make this data work with IGListKit. As far as I can work out, I need to be able to some how separate each JSON object and combine it into an array that implements IGListDiffable to be returned in `-objects(for:)`. 

Would it be possible to provide me with some pointers, and possibly give an example of downloading and parsing JSON data from the web in the IGListKit example app? I think it would be super useful as I can see most people using this framework with data downloaded from online!

Thanks!

jessesquires:
Hey @yusuftor ! 😄 

I would definitely parse out the JSON into model objects. It sounds like you're just using raw JSON? That's probably not the best practice. 😅 

The flow should be something like this:

1. Network request
2. Request completes, gives JSON to a parser object
3. Parser returns array of parsed model objects (that conform to `IGListDiffable`)
4. Return this array in `-objects(for:)`
5. Then in your `SectionController`, you'll receive this model in `-didUpdateToObject:`

kwstasna:
@jessesquires Thanks a lot for your help and explaining. But would it be difficult for you, with all my respect, to show us a legit example with a simple json example and how can we make a model that conforms to `IGListDiffable` ?
Again, thanks a lot 😃

rnystrom:
@kwstasna Check out the latest demo that we added for an example custom model that conforms to `IGListDiffable`:

https://github.com/Instagram/IGListKit/blob/master/Example/IGListKitExamples/ViewControllers/DiffTableViewController.swift#L18-L39

For doing JSON-to-`Person` and diffing, imagine something like:

```swift
let response: [ [String: Any] ] = // something from the web
var people = [Person]()
for json in response {
  if let pk = json["pk"] as? Int,
     let name = json["name"] as? String {
    people.append(Person(pk: pk, name: name))
  }
}
let result = IGListDiffable(oldPeople, people, .equality)
// do something w/ the diff
```

Sherlouk:
[...]

PR #132, opened 2016-10-28
Created a mixed data example driven by JSON

Sherlouk:
- Created an example much like the 'Mixed Data' example but completely driven from a local JSON file.
## Justification

I had noticed someone raised the point of how to use JSON (#97) and while it's beyond this library with actually parsing it I do feel that a lot of people will want to parse JSON and then display it in some way (probably caching it in the middle somewhere).

I've subclassed the 'Mixed Data' example as to not duplicate the same collection view and all that functionality and instead I override the `data` variable.
## Pull request checklist


Sherlouk:
@rnystrom @jessesquires 

Would be great to get a stance on this as is, obviously there's a conflict and will fix that when the time comes

rnystrom:
Got around to looking at this. A few thoughts:

I'm struggling a bit understanding the value that this bring to demoing how to use IGListKit. I feel like this more shows how to do JSON parsing. Right?

Also, I want to avoid subclassing at all costs. In fact I'm going to add another task to `final` all our demo classes. Not only will this make maintenance of the example app trickier, it's a pattern we deter at Instagram and within IGListKit.

@jessesquires what are your thoughts?

jessesquires:
@rnystrom - agreed. this is really a "how to use JSON example"

We should probably direct people toward StackOverflow for out-of-scope questions like this in the future.

jessesquires:
However, thanks @Sherlouk for the time you put into this! 😄 

At least, there's a record here for people to check out.

Sherlouk:
Thanks for the feedback!

Will leave the branch on my fork for people to checkout if they really do get stuck and for future reference.

PR #135, opened 2016-10-29
Fix example where collectionview starts from wrong position

### LINKED ISSUE #133: Bug: Nested CollectionView
If you add more nested CollectionView's to Nested Adapter example, you will notice that the horizontal scroll will be broken. Pod installed from master branch (https://github.com/Instagram/IGListKit). 
- Xcode 8.0

Sherlouk:

- Reset the collectionView offset when the cell is reused
- Fixes #133 

Note: If you scroll the first collection view, then take it off screen, when you go back to the top the collection view offset will be back to the beginning. I believe this to be the correct behaviour unless we either stop reuse of cells, or store the offset of each section and set it once it loads.

## Pull request checklist


shuhrat10:
I think this fixes won't will be work on all cases, see my comments in #133

Sherlouk:
Closing pending bigger discussion around storage of offsets and native nesting capabilities.

PR #140, opened 2016-10-30
Create a flow layout for IGList

zhubofei:
Create a flow layout for IGList that does not newline sections #3
- [x] Minimum inter-item spacing
- [x] Minimum line spacing
- [x] Constant item size with constant layout time
- [x] Update layout on insert/delete/move
- [x] Unit Test


jessesquires:
@zhubofei -- just a quick note, we'll definitely need some tests here. 😄 

@rnystrom -- can we adjust settings on @coveralls ? that shit is noisy AF 🗣 🙉

rnystrom:
I'm going to do a little bit of a deeper review on this to make sure I understand what all is going on, but looking at the screenshot in #3 it looks pretty cool! Some things I want to make sure we support:

- Section insets
- Minimum inter-item spacing
- Minimum line spacing

That way you can make a grid layout w/ spacing between all the items (thinking of Instagram's photo grid here).

zhubofei:
@rnystrom I updated the initial comment to include a list of tasks. I'm not quite sure if I should create a separate repo for this.

zhubofei:
giphy
This looks promising. I think I only need to figure out a better update algorithm. Maybe with better animation.

zhubofei:
@rnystrom Also, since this layout only allows one item per section, do we really need section inset?

rnystrom:
@zhubofei this is a tricky problem. On one hand we don't want to overcomplicate the layout, on the other, if we provide something _too basic_ then people will reach its limitations quickly.

I'm almost starting to think that we should provide links to other open source layouts that do a good job of this as a recommendation.

What do you think? + @jessesquires

jessesquires:
> Also, since this layout only allows one item per section, do we really need section inset?

Yeah, I don't think so. At least, this isn't the right terminology.

We'll need:

1. Spacing between items:   ⬛️` |---|` ⬛️ 
2. Spacing between "rows":

⬛️ ⬛️ ⬛️ 
`---`
`   |   `
`---`
⬛️ ⬛️ ⬛️

jessesquires:
@rnystrom -- I could go either way actually. (1) include layout here, (2) have a new lib for the layout.

either way, we *do* need it.

zhubofei:
@rnystrom The only existing layout I found that can do "no newline" is https://github.com/rhodgkins/RDHCollectionViewGridLayout. You can only do fixed size item with it and the API is a little bit ...weird. I crafted the code a little bit to support fixed size section yesterday. So the only feature left is Scroll direction and maybe some optimization.

rnystrom:
@zhubofei sounds great! Let's push this forward then. We can make sure to document that this is only for single-item-section feeds for now. Some thoughts:

- Can we name this something explicit, like: "IGListGridCollectionViewLayout"?
- Don't bother w/ auto layout (you mean self-sizing cells, right?)
- What do you mean by "alignment"?
- Let's get rid of section insets then per @jessesquires 
- Scroll direction is 👍 

And as we mentioned before, if we're bundling this w/ IGListKit we will _definitely_ need to cover it w/ unit tests. I built a few of our layouts in Instagram w/ unit tests so I'm happy to help steer direction w/ those if you need it.

zhubofei:
@jessesquires @rnystrom What I am trying to do here is just copying all the behaviors `UICollectionViewFlowLayout` has with the same API.

zhubofei:
@rnystrom By alignment, I mean `UICollectionViewFlowLayout` will increase the inter-item spacing by default if item width+minimumInteritemSpacing cannot take up the whole line. This creates a somewhat "center alignment". Shall we just do the same or provide a choice to keep minimumInteritemSpacing which is more like a "left alignment" .

jessesquires:
@zhubofei -- I think the "center alignment" approach will be better.

Regarding scroll direction -- for a first version, I think it's ok if we are **vertical only**. We can support horizontal scroll later if needed. (and it would be an additive, non-breaking change) To my knowledge, we don't need horizontal scrolling grids anywhere in IG.

cc @rnystrom

zhubofei:
@rnystrom Some advice on how to do the unit test on layout would be extremely helpful. I actually don't have a clue about where to start.

zhubofei:
@jessesquires I am half way through the implementation of "center alignment" right now. But I think Instagram actually uses "left alignment" for the grid layout, right? Like this screenshot of my profile page.
img_0040

jessesquires:
@zhubofei Ha 😅 you are correct.

A couple of options here:

1. Let's just do left-alignment 
2. Make this configurable (probably not worth it at the moment -- this could also be an additive change later)

zhubofei:
@rnystrom I wrote two simple tests. Could you give me some advice on how to add more?

rnystrom:
@zhubofei finally getting back to this. For tests, here are some things that we could do to help the process out:

We use the following macro to make testing frames easier:

```objc
#define IGAssertEqualFrame(frame, x, y, w, h, ...) \
do { \
CGRect expected = CGRectMake(x, y, w, h); \
XCTAssertEqual(CGRectGetMinX(expected), CGRectGetMinX(frame)); \
XCTAssertEqual(CGRectGetMinY(expected), CGRectGetMinY(frame)); \
XCTAssertEqual(CGRectGetWidth(expected), CGRectGetWidth(frame)); \
XCTAssertEqual(CGRectGetHeight(expected), CGRectGetHeight(frame)); \
} while(0)
```

Also, I'm not sure we need to bother testing the integration of the layout+`IGListAdapter`. Instead, let's just test plain ol' `UICollectionView`. We can create a dummy datasource that returns `UICollectionViewCell`s.

The data for the `UICollectionViewDataSource` can just be an array of arrays. Then it's just:

```objc
- (NSInteger)numberOfSectionsInCollectionView:(UICollectionView *)collectionView {
    return [self.sections count];
}

- (NSInteger)collectionView:(UICollectionView *)collectionView numberOfItemsInSection:(NSInteger)section {
    return [self.sections[section] count];
}

- (UICollectionViewCell *)collectionView:(UICollectionView *)collectionView cellForItemAtIndexPath:(NSIndexPath *)indexPath {
    return [collectionView dequeueReusableCellWithReuseIdentifier:@"cell" forIndexPath:indexPath];
}
```

Your test can have a helper methods for setup and fetching cells:

```objc
- (void)prepareWithData:(NSArray<NSArray<IGTestSection *> *> *)data {
    self.dataSource.sections = data;
    [self.layout prepareLayout];
    [self.collectionView reloadData];
    [self.collectionView layoutIfNeeded];
}

- (UICollectionViewCell *)cellForSection:(NSInteger)section item:(NSInteger)item {
    return [self.collectionView cellForItemAtIndexPath:[NSIndexPath indexPathForItem:item inSection:section];
}
```

Then bring it all together and test final frame positions:

```objc
- (void)test_something {
    [self prepareWithData:@[
                            @[ @1, 2 ],
                            @[ @1 ]
                            ]];
    XCTAssertEqual(self.collectionView.contentSize.height, 120);
    IGAssertEqualFrame([self headerForSection:0].frame, 10, 10, 85, 10);
    IGAssertEqualFrame([self cellForSection:0 item:0].frame, 10, 20, 85, 10);
    IGAssertEqualFrame([self cellForSection:0 item:1].frame, 10, 40, 85, 20);
    IGAssertEqualFrame([self headerForSection:1].frame, 10, 75, 85, 10);
    IGAssertEqualFrame([self cellForSection:1 item:0].frame, 10, 85, 85, 30);
}
```

zhubofei:
@rnystrom Do we need the support of multiple items per section for this layout?

rnystrom:
@zhubofei let's keep it one-item-per-section for now to make it simple. But we should assert that `numberOfItems()` is never > 1

zhubofei:
@rnystrom Not sure where to put the assert though🤔.

rnystrom:
@zhubofei probably in `-prepareLayout`. It could even be surrounded by `#DEBUG` like we do [here](https://github.com/Instagram/IGListKit/blob/master/Source/IGListAdapter.m#L389-L393)

```objc
- (void)prepareLayout {
#if DEBUG
  for (NSInteger section in [self.collectionView numberOfSections]) {
  	IGAssert([self.collectionView numberOfItemsInSection:section] <= 1, @"explain");
  }
#endif
}
```

Also not sure if we should support 0 items in the section? Like if something were deleted? If that's simple it'd be great!

zhubofei:
@rnystrom Yes, I think with some modification we can support 0 items in the section but not simple. It should require the same work of supporting multiple items per section. I'm not sure if 0 item will ever occur, though. I think when we remove an object from IGList the section will be removed as well?

rnystrom:
> I'm not sure if 0 item will ever occur, though

If there's one thing I've learned w/ open source, it's that what seems impossible will always end up possible 😉 

As long as we assert that `numberOfItems() == 1` its fine to assume this though. We just can't fail silently when 0 is returned b/c we'll end up w/ weird behavior, and it'll be hard to figure out what's wrong.

zhubofei:
@rnystrom Great! IMHO, this layout is ready for some code review. Do you still think we should change the name, though? From `IGListGridCollectionViewLayout` or `IGListCollectionViewGridLayout`? I'm currently using `IGListCollectionViewFlowLayout`.

rnystrom:
@zhubofei awesome! re: naming ya let's go w/ `IGListGridCollectionViewLayout`. Can we make sure to note in the header docs that this only works w/ section controllers that return 1 item?

zhubofei:
@rnystrom Just did!

Source/IGListGridCollectionViewLayout.m
```diff
@@ -0,0 +1,422 @@
...
+
+/**
+ The line number for each item in order of index path.
+ */
+@property (nonatomic, copy, nullable) NSMutableArray *lineForItem;
+
+/**
+ The width of in collection view content.
+ */
+@property (nonatomic, assign) CGFloat contentWidth;
```
rnystrom: This should be `readonly` since the getter is overridden and we're not using ivars.

Source/IGListGridCollectionViewLayout.m
```diff
@@ -0,0 +1,422 @@
...
+
+/**
+ The height of in collection view content.
+ */
+@property (nonatomic, assign) CGFloat contentHeight;
```
rnystrom: same

Source/IGListGridCollectionViewLayout.m
```diff
@@ -0,0 +1,422 @@
...
+    } else {
+        NSInteger section = indexPath.section;
+        NSInteger lineNumber = section / self.itemPerLine;
+        NSInteger column = section - lineNumber * self.itemPerLine;
+        CGFloat x = column * (self.itemSize.width + self.minimumInteritemSpacing);
+        CGFloat y = lineNumber * (self.itemSize.height + self.minimumLineSpacing);
+        CGRect frame = CGRectMake(x, y, self.itemSize.width, self.itemSize.height);
+        UICollectionViewLayoutAttributes *attributes = [UICollectionViewLayoutAttributes layoutAttributesForCellWithIndexPath:indexPath];
+        attributes.frame = frame;
+        return attributes;
```
rnystrom: nit: `const` all of the primitives

Source/IGListGridCollectionViewLayout.m
```diff
@@ -0,0 +1,422 @@
...
+    }
+}
+
+- (BOOL)shouldInvalidateLayoutForBoundsChange:(CGRect)newBounds
+{
+    return NO;
```
rnystrom: What happens on landscape change? Should probably be:
```objc
const CGRect oldBounds = self.collectionView.bounds;
return CGRectGetWidth(oldBounds) != CGRectGetWidth(newBounds) 
  || CGRectGetHeight(oldBounds) != CGRectGetHeight(newBounds);
```

Source/IGListGridCollectionViewLayout.m
```diff
@@ -0,0 +1,422 @@
...
+{
+    self.itemPerLine = (int) ((self.contentWidth + self.minimumInteritemSpacing)
+                               / (itemSize.width + self.minimumInteritemSpacing));
+    self.lineNumber = (self.collectionView.numberOfSections + self.itemPerLine - 1) / self.itemPerLine;
+    self.interitemSpacing = (self.contentWidth - (self.itemSize.width * self.itemPerLine)) / (self.itemPerLine - 1);
+}
+
+@end
+
+#pragma mark - _IGListGridLayoutLine Implementation
```
rnystrom: remove

Source/IGListGridCollectionViewLayout.m
```diff
@@ -0,0 +1,422 @@
...
+
+@implementation _IGListGridLayoutLine
+
+- (id)initWithMinimumInteritemSpacing:(CGFloat)spacing
+                            headIndex:(NSInteger)headIndex
+                                frame:(CGRect)frame
+                      scrollDirection:(UICollectionViewScrollDirection)direction
```
rnystrom: Should we assert any state of these params? e.g. things that should be `> 0`, etc

Source/IGListGridCollectionViewLayout.m
```diff
@@ -0,0 +1,422 @@
...
+
+#pragma mark - Private API
+
+- (UICollectionViewLayoutAttributes *)attributesForItemAtIndexPath:(NSIndexPath *)indexPath withXOffset:(CGFloat)x
+{
+    NSInteger index = indexPath.section - self.headIndex;
+    CGSize itemSize = [self.itemSizes[index] CGSizeValue];
+    
+    // Center vertically
+    CGFloat y = (self.frame.size.height - itemSize.height) / 2;
```
rnystrom: This division by int `2` is interesting, not sure if rounding is intended or if we should divide by `2.0`

zhubofei: @rnystrom I'm not sure about this either. I'm not sure if we should use round number for all point values.

Example/IGListKitExamples/ViewControllers/StoryboardViewController.swift
```diff
@@ -39,8 +39,16 @@ final class StoryboardViewController: UIViewController, IGListAdapterDataSource
...
+        let sectionController = StoryboardLabelSectionController()
+        sectionController.delegate = self
+        return sectionController
     }
     
     func emptyView(for listAdapter: IGListAdapter) -> UIView? { return nil }
+    
+    func removeSectionControllerWantsRemoved(_ sectionController: StoryboardLabelSectionController) {
+        let section = adapter.section(for: sectionController)
+        words.remove(at: Int(section))
```
zhubofei: @rnystrom Yes, swift identifies `section` as `UInt`

Source/IGListGridCollectionViewLayout.h
```diff
@@ -0,0 +1,37 @@
...
+
+/**
+ This layout only works with section controllers that return 1 item.
+ */
+@interface IGListGridCollectionViewLayout : UICollectionViewLayout
+
+/**
+ The scroll direction of the grid.
+ */
+@property (nonatomic, assign) UICollectionViewScrollDirection scrollDirection;
```
zhubofei: Not sure about this. I plan to support changing scroll direction in the future. I saw UICollectionViewFlowLayout allows us to change direction after init.

zhubofei:
@rnystrom IMHO, at least for fixed-size cell we can get a pretty good performance w/out cache. Actually, for large datasource, it is better to compute attributes on the go. Because `layoutAttributesForElementsInRect` would be called a lot more than `prepareLayout` and cache also takes a lot of memory.

rnystrom:
@zhubofei oops looks like this branch got out of sync, once we're rebased we can land!

zhubofei:
@rnystrom Can I create a new PR instead? There are too many conflicts here. It's really hard for me to rebase 😂.

rnystrom:
@zhubofei haha sure thing! We can close this and open a new one

zhubofei:
@rnystrom Just opened #225. I'm gonna close this one.

otymartin:
@zhubofei So setting insets using 'IGListGridCollectionViewLayout' is currently not possible?? I'm able to create a grid but i would like some spacing on the left and right.

zhubofei:
@otymartin You can set Minimum inter-item spacing and Minimum line spacing instead.

otymartin:
@zhubofei 
**2 Issues:**
Im using storyboards so for whatever reason setting `minimumInterimSpace` && ` minimumLineSpacing` programmatically in my section controller does nothing. It only works via storyboards (interesting).

**Secondly (Under header 2)** is where I'm using the `IGListGridCollectViewLayout`. I have minimum line & inter-item at 4 each. As you can see, once the number of items in the row is determined, the excess spacing is shifted to the right instead of increasing the `inter-item spacing` which is the expected behaviour as shown in `Mixed Data GridItem Example` which uses the `UICollectionViewFlowLayout` &&[ apple docs](https://developer.apple.com/reference/appkit/nscollectionviewflowlayout/1402872-minimuminteritemspacing)

Let me know if perhaps im doing something wrong!😓

img_3248

zhubofei:
@otymartin I think we better open an issue for this. For issue 1 can you post some code sample so that I can look into it? As for issue 2, you can refer to #263. This grid layout uses left alignment by default and we have not yet implemented center alignment like UICollectionViewFlowLayout yet.

otymartin:
@zhubofei  I'll open an issue

jessesquires:
let's definitely move this into it's own issue thread.

let's start with https://github.com/Instagram/IGListKit/pull/140#issuecomment-267478925

PR #147, opened 2016-10-31
Provide object in IGListSingleSectionController size block

### LINKED ISSUE #144: Provide object in IGListSingleSectionController size block
Working on a tutorial, I found not having access to the `object` in the size block of a single section controller pretty limiting (e.g. dynamically sized text). We should include the object as a parameter in the block.

_This will be a breaking change_

zhubofei:
Fix #144

### DIFF
--- a/Example/IGListKitExamples/ViewControllers/SingleSectionStoryboardViewController.swift
+++ b/Example/IGListKitExamples/ViewControllers/SingleSectionStoryboardViewController.swift
@@ -44,7 +44,7 @@ final class SingleSectionStoryboardViewController: UIViewController, IGListAdapt
             guard let cell = cell as? StoryboardCell, let number = data as? Int else { return }
             cell.textLabel.text = "Cell: \(number + 1)"
         }
-        let sizeBlock = { (context: IGListCollectionContext?) -> CGSize in
+        let sizeBlock = { (data: Any, context: IGListCollectionContext?) -> CGSize in
             guard let context = context else { return .zero }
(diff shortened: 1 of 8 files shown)


Example/IGListKitExamples/ViewControllers/SingleSectionViewController.swift
```diff
@@ -52,7 +52,7 @@ final class SingleSectionViewController: UIViewController, IGListAdapterDataSour
             cell.textLabel.text = "Cell: \(number + 1)"
         }
         
-        let sizeBlock = { (context: IGListCollectionContext?) -> CGSize in
+        let sizeBlock = { (data:Any, context: IGListCollectionContext?) -> CGSize in
```
rnystrom: nit: should be `item`

Example/IGListKitExamples/ViewControllers/SingleSectionStoryboardViewController.swift
```diff
@@ -44,7 +44,7 @@ final class SingleSectionStoryboardViewController: UIViewController, IGListAdapt
             guard let cell = cell as? StoryboardCell, let number = data as? Int else { return }
             cell.textLabel.text = "Cell: \(number + 1)"
         }
-        let sizeBlock = { (context: IGListCollectionContext?) -> CGSize in
+        let sizeBlock = { (data: Any, context: IGListCollectionContext?) -> CGSize in
```
rnystrom: nit: `item`

PR #162, opened 2016-11-04
Support supplementaryViews created from nibs (continued)

rnystrom:
Continuing the work on #90. I don't believe I can push directly to that PR since the origin is `master` of a repo I don't have access to.

https://help.github.com/articles/checking-out-pull-requests-locally/

I went ahead and added another supplementary view test copying the old one we had.

cc @jessesquires in case there's something else I can do here. I believe this will still give @rawlinxx credit?


jessesquires:
@rnystrom - yep, LGTM! 👍

PR #172, opened 2016-11-07
Update changelog with missing revs

rnystrom:
I like that we ref the PRs in the changelog. Updating the log with missing revs from internal diffs (no PRs).


jessesquires:
nice! 🙌

Issue #156, opened 2016-11-03
Mark all example classes final


rnystrom:
We avoid subclassing on basically everything with IGListKit ([example](https://github.com/Instagram/IGListKit/blob/master/Source/IGListAdapter.h#L36-L37)) and want to encourage good patterns in our examples too. We should use Swift's `final` on all of our example classes, I think.

jessesquires:
👍 

lol, I've been meaning to do this

Sherlouk:
Sorreh :(

rnystrom:
@Sherlouk all good! I totally forgot that we can do `final` in Swift at all.

PR #173, opened 2016-11-07
Add final to each class in Example

zhubofei:
#156


zhubofei:
Not sure if we should `final` subclasses of `UICollectionViewCell`.

jessesquires:
Thanks @zhubofei !

PR #218, opened 2016-11-17
Test adapter proxy object

### LINKED ISSUE #186: Unit test IGListAdapterProxy invocation overrides
https://coveralls.io/builds/8776307/source?filename=Source%2FInternal%2FIGListAdapterProxy.m#L67

We had one or two issues with these early on. Basically testing when the targets deallocate (they are all `__weak` ivars). Also would be good to test scroll view vs collection view targeting for each method we proxy.

rnystrom:

Fixes #186

## Pull request checklist


jessesquires:
🎉

PR #232, opened 2016-11-20
Fix incorrect class in variable definitions

antons:

Fix incorrect class in variable definitions in IGListIndexPathResult. No change in functionality, nothing to test.

## Pull request checklist

### DIFF
--- a/Source/IGListIndexPathResult.m
+++ b/Source/IGListIndexPathResult.m
@@ -37,9 +37,9 @@
 }
 
 - (IGListIndexPathResult *)resultForBatchUpdates {
-    NSMutableSet<NSIndexPath *> *deletes = [self.deletes mutableCopy];
-    NSMutableSet<NSIndexPath *> *inserts = [self.inserts mutableCopy];
-    NSMutableSet<NSIndexPath *> *filteredUpdates = [self.updates mutableCopy];
+    NSMutableArray<NSIndexPath *> *deletes = [self.deletes mutableCopy];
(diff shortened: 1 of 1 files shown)


rnystrom:
Actually the reason we use a set is to make the object lookups `O(1)`, if we change these to arrays we'll end up with an `O(n^2)` method.

antons:
@rnystrom In that case it’s currently `O(n^2)`, because these three variables are mutable arrays, not sets.

rnystrom:
@antons good eye! Looks like this got lost in refactoring. Want to fix it so we're operating on sets w/in this algorithm?

antons:
@rnystrom Initially I thought it was a remnant of an old API, where ivars were sets, so I corrected the definition. I see two options to fix the bug:

```objc
NSMutableSet<NSIndexPath *> *deletes = [NSMutableSet setWithArray:self.deletes];
```
If the order of elements is important, ordered set could be a better option:
```objc
NSMutableOrderedSet<NSIndexPath *> *deletes = [NSMutableOrderedSet orderedSetWithArray:self.deletes];
```

rnystrom:
@antons let's go w/ the first one, order isn't important at all. The only reason we ended up w/ `NSArray` for the result public API was that its (a little) easier to work w/ for consumers.

Thanks for digging on this!

antons:
@rnystrom My pleasure, and thank you for open sourcing this! I’ll submit a new pull request for the bug fix.

PR #239, opened 2016-11-21
fix coveralls badge

jessesquires:
README badge had 91% even though coveralls says 95%.

Pasted badge code from coveralls.


rnystrom:
I think the badges get cached. If I see the badge is behind I shift-refresh in Chrome and its fixed.

jessesquires:
Yeah - GitHub caching is a problem sometimes. But it didn't have `?branch=master` .. so ?

PR #241, opened 2016-11-21
Update README.md

jessesquires:
Per comments in #1. 

Enough people asking about this to add it here.


david6p2:
Thanks for the update @jessesquires

PR #243, opened 2016-11-21
GridLayout refinements + fixes

jessesquires:
- Clean up GridLayout + docs
- Remove GridLayout scroll direction, not used/supported?
- Fix some file target membership issues
- Add GridLayout to umbrella header
- "Sort by name" on xcode project directories
- crtl-I auto format in xcode
- re-gen docs

PR #246, opened 2016-11-22
Cleanup layout

jessesquires:
Replaces #243 

---------------------

- Clean up GridLayout + docs
- Remove GridLayout scroll direction, not used/supported?
- Fix some file target membership issues
- Add GridLayout to umbrella header
- "Sort by name" on xcode project directories
- crtl-I auto format in xcode
- re-gen docs


zhubofei:
@jessesquires IMO, setting default itemSize to (0, 0) instead of (100, 100) may be a better choice. Because if itemSize is set to non-zero, the layout would assume that all the cells are fix-sized.

jessesquires:
@zhubofei Ahh.. ok. Docs didn't specify that. 😉 

After this merges, want to submit a PR to fix + add docs ? 😄

jessesquires:
Ah actually, that change broke tests. I'll fix this up.

zhubofei:
Thank you 😅. I'll be more careful with initialization next time.

PR #250, opened 2016-11-22
Fix TV example

rnystrom:

TV example build was failing (badge red). Fixed w/ `pod install`.

## Pull request checklist

### DIFF
--- a/Examples/Examples-tvOS/Pods/Pods.xcodeproj/project.pbxproj
+++ b/Examples/Examples-tvOS/Pods/Pods.xcodeproj/project.pbxproj
@@ -7,72 +7,78 @@
 	objects = {
 
 /* Begin PBXBuildFile section */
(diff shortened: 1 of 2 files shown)


rnystrom:
Waiting on CI ✅

Sherlouk:
Would it be worth having a shell script at `Examples/` level which pod installs for all examples (future proofing for what ever is to come, but currently does iOS and tvOS)?

rnystrom:
@Sherlouk I thought about that, only downside is that if we do that during CI builds it'll hide builds that are failing after a fresh `git clone` or download of the repo. IMO everything (workspace, tests, and examples) should be build-and-runnable immediately after download. If we don't check them w/ CI then we wont notice until someone files and issue.

jessesquires:
@rnystrom -- we should run pod install on the other examples too

PR #262, opened 2016-11-27
Focus in tvOS app

### LINKED ISSUE #261: Focus is not working properly in tvOS Example
- tvOS 9.2 horizontal focusing is not working
- tvOS 10.0 when scroll all 3 lines of the horizontal scroll to the end and then navigate between row up and down, horizontal scroll start moving to the left instead stay on the same position.

shuhrat10:
fixed #261


jessesquires:
Thanks @shuhrat10 ! 🎉 

Looks good to me, but I don't know tvOS very well.

cc @Sherlouk

Sherlouk:
@jessesquires Never done tvOS before this either haha, was very pleased to get it working as far as it did but if this makes it better then great!

Issue #240, opened 2016-11-21
Allow subclassing of IGListCollectionView? Remove IGLK_SUBCLASSING_RESTRICTED?


lucabartoletti:
In my project I have a CCTransparentTableView

It's a UITableView subclass that takes care to make the topMargin transparent to show what is behind.
I'm trying to substitute one of the CCTransparentTableView with a IGListCollectionView. The problem is that I'd like to reimplement the transparency logic in a CCTransparentCollectionView : IGListCollectionView and use it as IGListCollectionView. But IGLK_SUBCLASSING_RESTRICTED is blocking that.

Is IGLK_SUBCLASSING_RESTRICTED really necessary?

jessesquires:
Hey @lucabartoletti 😄 

In general, we've taken pretty firm stances against subclassing in the Insta codebase, which is why this macro exists.

1. Can you elaborate on what `CCTransparentCollectionView` does? Could this behavior be achieved *without* subclassing, (or by providing a custom `UICollectionViewLayout`)? 
2. I'll leave this open as a proposal. However, this is the first request we've gotten for this. If more people ask for this, we'll consider lifting the restrictions —  but I don't think subclassing `UICollectionView` is very common.

lucabartoletti:
Thanks for the reply @jessesquires 

On the long term I plan to implement a custom UICollectionViewLayout to achieve the same effect.
My need for subclassing was due to the fact that I was substituting a UITableView subclass with a UICollectionView that uses IGListKit, I don't have time to spent on implementing a custom layout and the fastest thing was to adopt the same code that we have in the UITableView subclass on a UICollectionView subclass. 

I understand your point about IGLK_SUBCLASSING_RESTRICTED and your internal policy against subclassing.

rnystrom:
@lucabartoletti Feel free to also fork the project and remove the subclassing restriction if that's a requirement.

Got any examples of the design you're trying to achieve with this? We're happy to help guide custom layout stuff 😃

jessesquires:
Going to close this for now, since we're passing on changing this at the moment.

Will re-open if there's more interest.

jessesquires:
Note, we're discussing this again in #409

PR #285, opened 2016-12-05
Don't set UICollectionView properties in IGListCollectionView

### LINKED ISSUE #277: Init overrides storyboard settings in IGListCollectionView. Should -[IGListCollectionView commonInit] be removed?
## New issue checklist

- [x] I have reviewed the [`README`](https://github.com/Instagram/IGListKit/blob/master/README.md) and [documentation](http://instagram.github.io/IGListKit)
- [x] I have searched [existing issues](https://github.com/Instagram/IGListKit/issues) and this is not a duplicate

### General information

- `IGListKit` version: 1.0.0
- iOS version(s): 10.1
- CocoaPods/Carthage version:
- Xcode version: 8.1
- Devices/Simulators affected:
- Reproducible in the demo project? (Yes/No): 
- Related issues: #92 

The next release of IGListKit will support storyboards, but at present [-[IGListCollectionView commonInit]](https://github.com/Instagram/IGListKit/blob/f2a9d93be84e7ea23ec85dddbb3e2421215a6eec/Source/IGListCollectionView.m#L15-L16) sets a couple of options which override those provided by the storyboard.

I can open a PR to remove these if appropriate. The IB default appears to be white, but the example ViewControllers would need updating to explicitly set a background color in code.

dcaunt:
Fixes #277

### DIFF
--- a/Source/IGListCollectionView.m
+++ b/Source/IGListCollectionView.m
@@ -11,25 +11,6 @@
 
 @implementation IGListCollectionView
 
-- (void)commonInit {
-    self.backgroundColor = [UIColor whiteColor];
-    self.alwaysBounceVertical = YES;
-}
-
-- (instancetype)initWithFrame:(CGRect)frame collectionViewLayout:(UICollectionViewLayout *)layout {
-    if (self = [super initWithFrame:frame collectionViewLayout:layout]) {
-        [self commonInit];
-    }
-    return self;
-}
-
-- (instancetype)initWithCoder:(NSCoder *)aDecoder {
-    if (self = [super initWithCoder:aDecoder]) {
-        [self commonInit];
-    }
-    return self;
-}
-
 - (void)layoutSubviews {
     /**
      UICollectionView will sometimes lay its cells out with an animation. This is especially noticeable on older devices
--- a/Source/IGListCollectionView.m
+++ b/Source/IGListCollectionView.m
@@ -11,6 +11,14 @@
 
 @implementation IGListCollectionView
 
+- (instancetype)initWithFrame:(CGRect)frame collectionViewLayout:(UICollectionViewLayout *)layout {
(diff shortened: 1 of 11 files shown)


Source/IGListCollectionView.m
```diff
@@ -11,25 +11,6 @@
...
 @implementation IGListCollectionView
 
-- (void)commonInit {
-    self.backgroundColor = [UIColor whiteColor];
-    self.alwaysBounceVertical = YES;
-}
-
-- (instancetype)initWithFrame:(CGRect)frame collectionViewLayout:(UICollectionViewLayout *)layout {
-    if (self = [super initWithFrame:frame collectionViewLayout:layout]) {
-        [self commonInit];
```
rnystrom: Can we move all of the `commonInit` into here and just get rid of `initWithCoder:`? I think its nice having a default white bg, saves a lot of boilerplate.

Thoughts @jessesquires?

dcaunt: FWIW the rationale for removal was that `IGListCollectionView` exists only to provide the compiler warnings but I'm more than happy to update the PR, I agree it's convenient.

jessesquires: actually, i'm fine either way. 

but i agree that white + bouncing is convenient. 

(also, IG implementations probably rely on these defaults right now?)

rnystrom: @jessesquires ya and w/out the compiler we will likely break some stuff w/out warning. @dcaunt let's update and maybe open a new issue for that?

I'd love to consider removing the dependence on `IGListCollectionView` but admit I **love** the compiler warnings for those methods.

jessesquires: agree, we should keep `IGListCollectionView` for those warnings.

but, we should consider #240 if more requests come up.

Issue #306, opened 2016-12-09
Cut 2.0.0 release


rnystrom:
# Launch Checklist

- [x] Run through documentation one last time checking header docs, formatting, and links (e.g. Guides)
- [x] PR [podspec version](https://github.com/Instagram/IGListKit/blob/master/IGListKit.podspec#L3) update to 2.0.0 against `master`
- [x] Add 2.0.0 release with changelog as notes
- [x] PR reseting changelog linking to release notes for 2.0.0 (include migration guide link)
  - [Example](https://github.com/jessesquires/JSQMessagesViewController/blob/develop/CHANGELOG.md#700)
- [x] Run `pod install` on examples
- [x] PR `master` against `stable` with latest commit https://github.com/Instagram/IGListKit/commit/b9ede32f5be71bed87b0d82c3cb7e50daa6444ea
- [x] Confirm `stable` is `0|0` [ahead/behind](https://github.com/Instagram/IGListKit/branches)
- [x] Publish GH release
- [x] `$ pod spec lint` locally and with CI
- [x] Push updated podspec with `$ pod trunk push IGListKit.podspec`
  - ~~@jessesquires do you have write access w/ IGListKit CP?~~
- [x] Verify 2.0.0 picked up on [Cocoapods](https://cocoapods.org/pods/IGListKit) including new documentation
- [ ] Test `$ pod install` on hobby weather app to see update
- [x] Send out tweestorm

rnystrom:
From 2.0.0 release notes draft:

> TODO:
> 
> Update and copy-pasta CHANGELOG.md here
> Update .xcodeproj and .podspec version numbers
> Gen and push docs
> Merge master into stable
> Close the milestone
> Click "publish release" here (from stable branch)
> run pod trunk push ... to update CocoaPods

rnystrom:
Release notes draft updated https://github.com/Instagram/IGListKit/releases/tag/untagged-81a98143190def1c7968

jessesquires:
@rnystrom closing this out.

One TODO left for you, testing your weather app 😊

PR #307, opened 2016-12-09
Update stable to 2.0.0

rnystrom:
Cut point https://github.com/Instagram/IGListKit/commit/b9ede32f5be71bed87b0d82c3cb7e50daa6444ea

Part of #306


rnystrom:
Missing diff that updates `.podspec` and constant version number

PR #308, opened 2016-12-09
Update podspec and project version to 2.0.0

rnystrom:
This needs to land and be part of #307 before that gets merged, right?

Part of #306


jessesquires:
yep!

PR #313, opened 2016-12-10
Update README.md

jessesquires:
Just a few tweaks, less verbose.

Also -- removed the "manual" installation. I don't think hardly anyone uses this. Plus, I don't think a simple drag + drop would work because of `OCMock`.


rnystrom:
I dig it! 👌

PR #314, opened 2016-12-10
Check for broken links in markdown files

### LINKED ISSUE #312: Markdown Link Verification
Since v1.0 a lot of files have moved around to make place for new functionality for example extra example packs or simply neatening up the file structure and this causes links to break.

While it seems the structure is a bit more stable now and links should continue to work I was wondering if we should consider some sort of check in travis?

By no means am I saying use this, but [this CLI tool](https://www.npmjs.com/package/markdown-link-check) could work by just checking each of the our MD files causing builds to fail if they produce broken links catching them before they get merged!

Sherlouk:

- Install `markdown-link-check` as a dependency
- Check each markdown file for links, and check them using aforementioned library

Closes #312 

A few points of note, I'm not sure if we can use the built-in caching capabilities to cache the library (at the moment it's installed for each build, even though it's only used for 1 -- seems a big waste!). Also a downside of the library is it doesn't really like relative links, we only had one -- so I changed it to be absolute.

## Pull request checklist

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -45,7 +45,7 @@ script:
 - set -o pipefail
 
 - if [ $CHECK_MARKDOWN == "YES" ]; then
-      find . -name "*.md" | markdown-link-check;
+      find . -name "*.md" -exec cat {} \; | markdown-link-check;
   fi
 
 - if [ $POD_LINT == "YES" ]; then
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,7 +21,7 @@ env:
    - EXAMPLE_SCHEME="IGListKitExamples"
 
    matrix:
-   - DESTINATION="OS=8.1,name=iPhone 6" SDK="$IOS_SDK" SCHEME="$IOS_SCHEME" RUN_TESTS="NO" BUILD_EXAMPLE="YES" RUN_UI_TESTS="NO" POD_LINT="YES"
+   - DESTINATION="OS=8.1,name=iPhone 6" SDK="$IOS_SDK" SCHEME="$IOS_SCHEME" RUN_TESTS="NO" BUILD_EXAMPLE="YES" RUN_UI_TESTS="NO" POD_LINT="YES" CHECK_MARKDOWN="YES"
    - DESTINATION="OS=8.2,name=iPhone 6" SDK="$IOS_SDK" SCHEME="$IOS_SCHEME" RUN_TESTS="NO" BUILD_EXAMPLE="NO"  RUN_UI_TESTS="NO" POD_LINT="NO"
    - DESTINATION="OS=8.3,name=iPhone 6" SDK="$IOS_SDK" SCHEME="$IOS_SCHEME" RUN_TESTS="NO" BUILD_EXAMPLE="NO"  RUN_UI_TESTS="NO" POD_LINT="NO"
    - DESTINATION="OS=8.4,name=iPhone 6" SDK="$IOS_SDK" SCHEME="$IOS_SCHEME" RUN_TESTS="NO" BUILD_EXAMPLE="NO"  RUN_UI_TESTS="NO" POD_LINT="NO"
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -39,15 +39,21 @@ env:
 
 before_install:
   - gem install slather --no-rdoc --no-ri --no-document --quiet
+  - npm install -g markdown-link-check
 
 script:
 - set -o pipefail
 
+- if [ $CHECK_MARKDOWN == "YES" ]; then
+      echo -en 'travis_fold:start:script.1\\r';
```
Sherlouk: It prints out a big block with ticks and crosses next to each URL, added a fold so it doesn't take over the entire job window.

CHANGELOG.md
```diff
@@ -51,7 +51,7 @@ You can find a [migration guide here](https://instagram.github.io/IGListKit/migr
 
 - Added support for cells created from nibs. [Sven Bacia](https://github.com/svenbacia) [(#56)](https://github.com/Instagram/IGListKit/pull/56)
 
-- Added an additional initializer for `IGListSingleSectionController` to be able to support single sections created from nibs. An example can be found [here](Examples/Examples-iOS/IGListKitExamples/ViewControllers/SingleSectionViewController.swift). [(#56)](https://github.com/Instagram/IGListKit/pull/56)
+- Added an additional initializer for `IGListSingleSectionController` to be able to support single sections created from nibs. An example can be found [here](https://github.com/Instagram/IGListKit/tree/master/Examples/Examples-iOS/IGListKitExamples/ViewControllers/SingleSectionViewController.swift). [(#56)](https://github.com/Instagram/IGListKit/pull/56)
```
jessesquires: 👍

jessesquires:
Thanks @Sherlouk !

I'd also like to see an example of a failure, if possible.

.travis.yml
jessesquires: actually, it's probably ok *not* to fold this. it's only going to run on the first build, right?

doesn't travis fold most steps automatically after completion?

Sherlouk:
@jessesquires Will wait for this build to go through fine (moved the `npm install`). Assuming that works I'll break a link and let that build. (then probably forget to fix it 😂)

.travis.yml
Sherlouk: Travis does to some auto-folding, but from the testing I've done it never seemed to fold this.
Yes it only gets triggered on the first build, I'll remove it for the next build and can see if you're happy to keep it unfolded after that!

.travis.yml
```diff
@@ -39,15 +39,21 @@ env:
 
 before_install:
   - gem install slather --no-rdoc --no-ri --no-document --quiet
+  - npm install -g markdown-link-check
```
Sherlouk: Moved the `npm install`, and I've googled around a lot with regards to the `cache:` and you can add a bit which is `directories:` so I tried specifying `node_modules` but I couldn't see this effecting anything.

Happy to take input though!

Sherlouk:
Glad we tested that, it seems the `set -o pipefail` was making it do the opposite of what we want it to do on this command. Rather than changing it for all the others, I've made it exit (not echo, as my testing instead had oops) immediately on fail.

Also for what it's worth the list of files this will check:
>./.github/CONTRIBUTING.md
./.github/ISSUE_TEMPLATE.md
./.github/PULL_REQUEST_TEMPLATE.md
./CHANGELOG.md
./Examples/Examples-iOS/LICENSE-examples.md
./Examples/Examples-tvOS/LICENSE-examples.md
./Guides/Getting Started.md
./Guides/IGListDiffable and Equality.md
./Guides/Migration.md
./LICENSE.md
./PATENTS.md
./Pods/OCMock/README.md
./README.md

Sherlouk:
Hurrah! So after eliminating multiple variables it turns out the reason a bad link passed the built was because of the folding I added... Intriguing!

Alas it is working now!

jessesquires:
Nice! 💯 Looks awesome. 🎉 Thanks @Sherlouk !

No need to squash, our syncing system will take care of that.

jessesquires:
Oh - let's fix that bad link 😄

rnystrom:
This is really awesome. TIL!

PR #319, opened 2016-12-12
Invalidate layout when using reloadData

### LINKED ISSUE #305: Cells not showing up after reloadData
## New issue checklist

- [x] I have reviewed the [`README`](https://github.com/Instagram/IGListKit/blob/master/README.md) and [documentation](http://instagram.github.io/IGListKit)
- [x] I have searched [existing issues](https://github.com/Instagram/IGListKit/issues) and this is not a duplicate

### General information

- `IGListKit` version: commit: https://github.com/Instagram/IGListKit/commit/8bfe2fefdd97ecc3cf8ecc96eaba5a91758a1b11
- iOS version(s): 10.1
- CocoaPods/Carthage version: 1.1.1
- Xcode version: 8.2 beta 2
- Devices/Simulators affected: iPhone 7
- Reproducible in the demo project? (Yes/No):
- Related issues:

Hey guys,

Firstly, I adore this library. Amazing job!
So I am probably doing something wrong, but here goes.

I have a toggle button that filters objects based on a Bool. Within the DataSource I have: 

```
public func objects(for listAdapter: IGListAdapter) -> [IGListDiffable] {
        guard let delegate = delegate else { return [] }
        
        var objects = !isFeatured ? delegate.model.posts as [IGListDiffable] : delegate.model.featuredPosts as [IGListDiffable]
```

model.featuredPosts is a simple filter:
```
var featuredPosts: [Post] {
        get {
            return posts.filter { $0.isFeatured }
        }
}
```

Within my delegate I am reloading that data with:
```
adapter.reloadData { (completed) in
    print(completed)
}
```

When I toggle to only show isFeatured posts, the collection view updates as expected and I see 3 items. When I toggle it again to show all posts (8 in total), I can see it requesting the sections and cells, however the collection view isn't updated. 

When I update the completion block as: 
```
adapter.reloadData(completion: { (completed) in
            self.adapter.performUpdates(animated: false, completion: nil)
            print(completed)
            print(self.adapter.collectionView?.numberOfSections)
 })
```

I can see the correct number of sections being printed (8), but not displayed. 

Any suggestions on what I may be doing wrong?

For what it's worth the collection view frame is being set:
```
override func viewDidLayoutSubviews() {
        super.viewDidLayoutSubviews()
        collectionView.frame = view.bounds
}
```

Thanks in advance,
Steve

rnystrom:

Always invalidate the `collectionViewLayout` when calling `reloadData`. Turns out there are inconsistencies when layout attributes are queries/stored using `UICollectionViewFlowLayout` and `estimatedItemSize`.

Reproduced in unit test, fixed the test.

Closes #305

## Pull request checklist

### DIFF
--- a/Source/IGListAdapterUpdater.m
+++ b/Source/IGListAdapterUpdater.m
@@ -84,6 +84,7 @@
 
     [delegate listAdapterUpdater:self willReloadDataWithCollectionView:collectionView];
     [collectionView reloadData];
+    [collectionView.collectionViewLayout invalidateLayout];
     [collectionView layoutIfNeeded];
     [delegate listAdapterUpdater:self didReloadDataWithCollectionView:collectionView];
 
--- a/Tests/IGListAdapterUpdaterTests.m
+++ b/Tests/IGListAdapterUpdaterTests.m
@@ -397,4 +397,45 @@
     XCTAssertEqual(inserts.count, 0);
 }
 
+- (void)test_whenCallingReloadData_withUICollectionViewFlowLayout_withEstimatedSize_thatSectionItemCountsCorrect {
+    UICollectionViewFlowLayout *layout = [UICollectionViewFlowLayout new];
+    // setting the estimated size of a layout causes UICollectionView to requery layout attributes during reloadData
+    // this becomes out of sync with the data source if the section/item count changes
+    layout.estimatedItemSize = CGSizeMake(100, 10);
+
+    UICollectionView *collectionView = [[UICollectionView alloc] initWithFrame:CGRectMake(0, 0, 100, 100) collectionViewLayout:layout];
+    IGListTestUICollectionViewDataSource *dataSource = [[IGListTestUICollectionViewDataSource alloc] initWithCollectionView:collectionView];
+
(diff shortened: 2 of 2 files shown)


rnystrom:
Will update w/ changelog after another PR that preps changelog for next version

Source/IGListAdapterUpdater.m
```diff
@@ -84,6 +84,7 @@ - (void)performReloadDataWithCollectionView:(UICollectionView *)collectionView {
 
     [delegate listAdapterUpdater:self willReloadDataWithCollectionView:collectionView];
     [collectionView reloadData];
+    [collectionView.collectionViewLayout invalidateLayout];
     [collectionView layoutIfNeeded];
```
jessesquires: because of the layout invalidation, i don't think we need `-layoutIfNeeded` anymore. want to check?

rnystrom: Will check, I'll need to spin through the places in Instagram.app that we call `reloadDataWithCompletion:` to make sure there are no weird side effects.

rnystrom:
Tested internally on a couple surfaces I remember having trouble before. I think we're g2g.

PR #235, opened 2016-11-20
macOS support

insidegui:
## Changes in this pull request

Added support for macOS. The macOS target includes only the core diffing functionality so `IGListDiff` can be used on macOS apps.

I have not added the tests because I don't know how you're setting them up, maybe the same tests can be used by adding some `#if os(...)`s to the test files.

I have tested integration on iOS, tvOS and macOS using both Carthage and Cocoapods.

PR #322, opened 2016-12-12
Update README.md

jessesquires:
@rnystrom - I know you removed this before, but I think we should add it back now that we support > 1 platform. 😄 

(and macOS is coming up in #235 )


jessesquires:
don't worry, it's still centered and pretty:

screen shot 2016-12-12 at 10 47 02 am

rnystrom:
👌  

Since macOS is coming I'm down w/ this. Good to show that we support all platforms.

Coming soon watchOS? 😂

Issue #318, opened 2016-12-12
Disable prefetchEnabled by default on IGListCollectionView


rnystrom:
Our experiments find that disabling this feature that IGListKit does not uses yields ~5% increase to scroll performance. `-[UICollectionView prefetchEnabled]` defaults to `YES` with the iOS 10 SDK.

svenbacia:
I started working on this [here](https://github.com/svenbacia/IGListKit/commit/54ae08cb005365bdf1cdbfd6f79fab759095911c). Currently the default value is only applied for collectionViews which are created from the `initWithFrame:collectionViewLayout:` initializer. Should the default value also be applied for collectionViews created from Storyboard for example by overriding `initWithCoder:`?

rnystrom:
@svenbacia yes definitely! We should make sure _every_ `IGListCollectionView` has the same behavior.

rnystrom:
nit on the branch before its PR'd:

Instead of checking OS version we can just check `[self respondsToSelector:@selector(setPrefetchingEnabled:)]`.

jessesquires:
Fixed in #323

PR #323, opened 2016-12-12
disables prefetchEnabled by default

svenbacia:
Disables `prefetchEnabled` by default on `IGListCollectionView` (#318).

## Pull request checklist

### DIFF
--- a/Source/IGListCollectionView.m
+++ b/Source/IGListCollectionView.m
@@ -19,7 +19,20 @@
             self.backgroundColor = [UIColor whiteColor];
         }
         
+        if ([self respondsToSelector:@selector(setPrefetchingEnabled:)]) {
+            self.prefetchingEnabled = NO;
+        }
+        
         self.alwaysBounceVertical = YES;
+    }
+    return self;
+}
+
+- (instancetype)initWithCoder:(NSCoder *)aDecoder {
+    if (self = [super initWithCoder:aDecoder]) {
+        if ([self respondsToSelector:@selector(setPrefetchingEnabled:)]) {
+            self.prefetchingEnabled = NO;
+        }
     }
     return self;
 }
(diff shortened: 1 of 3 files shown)


Source/IGListCollectionView.m
```diff
@@ -19,11 +19,25 @@ - (instancetype)initWithFrame:(CGRect)frame collectionViewLayout:(UICollectionVi
...
     }
     return self;
 }
 
+- (instancetype)initWithCoder:(NSCoder *)aDecoder {
+    if (self = [super initWithCoder:aDecoder]) {
+        [self updateIGListCollectionViewDefaultValues];
+    }
+    return self;
+}
```
jessesquires: ha 😅  we removed `initWithCoder:` previously so that we don't override storyboard settings.

let's remove, but let's keep your tests (asserts will need to be flipped)

Source/IGListCollectionView.m
```diff
@@ -19,11 +19,25 @@ - (instancetype)initWithFrame:(CGRect)frame collectionViewLayout:(UICollectionVi
...
+
+- (void)updateIGListCollectionViewDefaultValues {
```
jessesquires: let's rename: `updateWithDefaultValues`

jessesquires:
Ha, sorry. Just saw @rnystrom 's comments in #318.

jessesquires:
Actually, it feels bad to override storyboard settings. We probably shouldn't... 

Maybe we only do it for the prefetch property?

cc @rnystrom

jessesquires:
1 more thing - @svenbacia, can you also update the CHANGELOG ? 😄

rnystrom:
@svenbacia +1 _only_ changing the `prefetchingEnabled` value in init. Let's leave the rest of the settings alone. 

I took a look and storyboards do have this as an option. However I think it's still worth disabling since you can't actually use the feature.

@jessesquires maybe we should override:

```objc
-(id<UICollectionViewDataSourcePrefetching>)prefetchDataSource {
  return nil;
}
```

Maybe assert that it isn't used? Or maybe we add compiler warnings like we do with [other illegal methods](https://github.com/Instagram/IGListKit/blob/master/Source/IGListCollectionView.h#L31)?

rnystrom:
Awesome! @svenbacia just need to rebase then we can land.

PR #324, opened 2016-12-13
Add more missing unit tests

rnystrom:
Working on that coverage %. Hustle to 💯 

Seriously tho, nearly maxed the coverage that is realistic. I think we need to spend some time on the grid layout tho, there are parts untested that also are unnecessary. I should file a 3.0.0 issue to track work on that.

Waiting for CI ✅


rnystrom:
UI test failure again on iOS 10.1

Issue #270, opened 2016-11-30
macOS support follow-up work from #235


jessesquires:
We added basic macOS support in #235.

Once it's merged, we need to do a few more things:

- [x] Update changelog
- [x] Add a test target, `IGListKit-macOSTests` and existing tests that apply
- [x] Update `.travis.yml` to build and run the macOS targets on travis-ci
- [x] Add an example project in `Examples/Examples-macOS/`
- [x] Update README
- ~~Create a macro in `IGListMacros.h` for the following imports, so we don't have to copy-paste. Let's call it `IGLK_TARGET_OS_IMPORTS`.~~

```objc
#import <TargetConditionals.h>

#if TARGET_OS_EMBEDDED || TARGET_OS_SIMULATOR
    #import <UIKit/UIKit.h>
#else
    #import <Cocoa/Cocoa.h>
#endif
```

- [x] The `.podspec` currently lists files manually which is cumbersome. Let's move common files to `Source/Common/` so that `spec.osx.source_files` can become: `'Source/Common/*.{h,m,mm}'`. We'll also need to do the same for `Source/Internal/Common/`. 

---------------------------
EDIT:

#### Additional follow-ups for internal

- [x] Run `pod install` everywhere
- [x] Lower macOS target in podspec? (currently 10.11)

insidegui:
Let me know if you need help with the example project.

jessesquires:
Thanks @insidegui !

Because of how our syncing system works, we'll have to setup the basic project internally and push it. After that, we can have contributors help.

jessesquires:
@insidegui, just FYI #333 😄 

@Sherlouk 2 things up for grabs here:

1. The macro mentioned above. I had trouble getting this to work. My macro game is not strong. cc @rnystrom 
2. mac example, #333 (probably coordinate with @insidegui)

Sherlouk:
@jessesquires Cheers for the heads up!

Actually I think if I were to rebase correctly some of the changes from my branch are still valid against the original checklist -- or is that something you're still working on internally? (Referring to Common directory/Podspec update)

insidegui:
@jessesquires I don't know if a macro would be the best solution for this case. When porting [Keyframes](https://github.com/facebookincubator/Keyframes/pull/64) to macOS I created a separate header called `Compatibility` and imported it everywhere instead of adding the `#ifdef`s to each file individually.

I can do the sample :)

Sherlouk:
[...]

PR #328, opened 2016-12-14
macOS support follow-up work, #270

jessesquires:
Progress on #270. 

- Setup basic, empty example project
- Setup test target and add tests that work for macOS
- Update changelog
- Update travis-ci


jessesquires:
Once travis is green, I'll land, then run `pod install` on everything internally and push.

Issue #333, opened 2016-12-14
macOS example project


jessesquires:
Follow-up from #235 and #270.

As of 9600c23d151d8cf89dfef48397f3905542d210d7 we've got all the basics and infra setup. There's an empty macOS example app that does nothing.

https://github.com/Instagram/IGListKit/tree/master/Examples/Examples-macOS

Currently, this is valuable just to test CocoaPods integration.

It would be nice if someone wants to build this out with some more functionality, preferably something simple and lightweight.

Remember small, incremental PRs are best! 😄

jessesquires:
Basic project added in #337

PR #337, opened 2016-12-15
macOS example app

insidegui:
Related issue: #333 

I've made a very simple macOS example app. It's just a list of names which can be searched, shuffled or deleted.

I think this is a good starting point for anyone who wants to use this on macOS projects :)

iglistkitmac


jessesquires:
@insidegui that gif looks awesome! 🎉

rnystrom:
Omg this is great!

antons:
We’re using IGListKit with NSTableView, and I must mention an important point about moves. While UITableView expects to be told where each row was and where it was and where it is now, NSTableView requires moves to be reported incrementally. If moves are reported as they are now, multiple moves will often result in an inconsistent internal state and crashes. 

As far as I can see, NSCollectionView uses the newer behavior found in UITableView and UICollectionView.

I’ve added an option for reporting incremental moves to our internal copy of IGListDiff. Besides NSTableView, it’s useful for replaying changes on collections. Let me know if you’d like me to submit a pull request with it.

antons:
To give a concrete example, try replacing implementation of UsersViewController.shuffle() with:

```swift
var usersModified = users
let second = users[1]
usersModified.remove(at: 1)
usersModified.insert(second, at: 0)
users = usersModified
```

This produces a diff with:
```
<IGListIndexSetResult 0x600000073300; 0 inserts; 0 deletes; 0 updates; 2 moves>
▿ 2 elements
  - 0 : <IGListMoveIndex 0x6000000268c0; from: 1; to: 0;>
  - 1 : <IGListMoveIndex 0x60000002d720; from: 0; to: 1;>
```

Before: 

```
▿ 201 elements
  - 0 : "Aaron"
  - 1 : "Abigail"
…
```

After:
```
▿ 201 elements
  - 0 : "Abigail"
  - 1 : "Aaron"
…
```

Since NSTableView applies moves incrementally, it moves row at index 1 to 0, then moves row at index 0 (the row it just moved from index 1 in previous step) to 1. As a result, there’s no visible change, and the visible state does not match the data source.

insidegui:
@antons I didn't know that, thanks for reporting. I think a pull request would be really helpful, but let's see what @jessesquires and @rnystrom think about it :)

jessesquires:
@antons Thanks! 😄 I would definitely like to see a PR if you have the time 👍

jessesquires:
@antons Let's also open a new issue for this 😊

PR #342, opened 2016-12-19
Add incremental move reporting

antons:

Added incremental move reporting, as mentioned in #337. Diffs with incremental moves are necessary for NSTableView, and can be used to apply changes to collections.

## Pull request checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,8 +8,9 @@ The changelog for `IGListKit`. Also see the [releases](https://github.com/instag
 This release closes the [2.1.0 milestone](https://github.com/Instagram/IGListKit/milestone/2).
 
 ### Enhancements
-
 - Added support for macOS. Note: this is *only* for the Diffing components. There is **no support** for `IGListAdapter`, `IGListSectionController`, and other components at this time. [Guilherme Rambo](https://github.com/insidegui) [(#235)](https://github.com/Instagram/IGListKit/pull/235)
+
+- Added algorithm for generating incremental moves, which can be used to reorder a collection, and are required for moving rows with NSTableView. Available via `IGListDiffBehaviorIncrementalMoves` for `IGListDiffWithBehavior` and `IGListDiffPathsWithBehavior`. [Anton Sotkov](https://github.com/antons) [(#342)](https://github.com/Instagram/IGListKit/pull/342)
 
 - Added a [macOS example](https://github.com/Instagram/IGListKit/tree/master/Examples/Examples-macOS) project. [Guilherme Rambo](https://github.com/insidegui) [(#337)](https://github.com/Instagram/IGListKit/pull/337)
 
--- a/Source/Common/IGListDiff.h
+++ b/Source/Common/IGListDiff.h
@@ -34,7 +34,43 @@ typedef NS_ENUM(NSInteger, IGListDiffOption) {
  */
 typedef NS_OPTIONS(NSInteger, IGListDiffBehavior) {
     IGListDiffBehaviorDefault = 0,
+    /**
+     When used, moves are reported incrementally. Can be used to reorder a collection. O(n^2) complexity.
+     
+     By default, moves are reported with old and new positions. O(n) complexity.
+     */
+    IGListDiffBehaviorIncrementalMoves = 1 << 0
 };
+
+/**
+ Creates a diff using indexes between two collections.
+
+ @param oldArray The old objects to diff against.
+ @param newArray The new objects.
+ @param option   An option on how to compare objects.
+
+ @return A result object containing affected indexes.
+ */
+FOUNDATION_EXTERN IGListIndexSetResult *IGListDiff(NSArray<id<IGListDiffable>> *_Nullable oldArray,
+                                                   NSArray<id<IGListDiffable>> *_Nullable newArray,
+                                                   IGListDiffOption option);
+
+/**
+ Creates a diff using index paths between two collections.
+
+ @param fromSection The old section.
+ @param toSection   The new section.
+ @param oldArray    The old objects to diff against.
+ @param newArray    The new objects.
+ @param option      An option on how to compare objects.
+
+ @return A result object containing affected indexes.
+ */
+FOUNDATION_EXTERN IGListIndexPathResult *IGListDiffPaths(NSInteger fromSection,
+                                                         NSInteger toSection,
+                                                         NSArray<id<IGListDiffable>> *_Nullable oldArray,
+                                                         NSArray<id<IGListDiffable>> *_Nullable newArray,
+                                                         IGListDiffOption option);
 
 /**
  Creates a diff using indexes between two collections.
@@ -69,35 +105,5 @@ FOUNDATION_EXTERN IGListIndexPathResult *IGListDiffPathsWithBehavior(NSInteger f
                                                                      NSArray<id<IGListDiffable>> *_Nullable newArray,
                                                                      IGListDiffOption option,
                                                                      IGListDiffBehavior behavior);
-
-/**
- Creates a diff using indexes between two collections.
-
- @param oldArray The old objects to diff against.
- @param newArray The new objects.
- @param option   An option on how to compare objects.
-
- @return A result object containing affected indexes.
- */
-FOUNDATION_EXTERN IGListIndexSetResult *IGListDiff(NSArray<id<IGListDiffable>> *_Nullable oldArray,
-                                                   NSArray<id<IGListDiffable>> *_Nullable newArray,
-                                                   IGListDiffOption option);
-
-/**
- Creates a diff using index paths between two collections.
-
- @param fromSection The old section.
- @param toSection   The new section.
- @param oldArray    The old objects to diff against.
- @param newArray    The new objects.
- @param option      An option on how to compare objects.
-
- @return A result object containing affected indexes.
- */
-FOUNDATION_EXTERN IGListIndexPathResult *IGListDiffPaths(NSInteger fromSection,
-                                                         NSInteger toSection,
-                                                         NSArray<id<IGListDiffable>> *_Nullable oldArray,
-                                                         NSArray<id<IGListDiffable>> *_Nullable newArray,
-                                                         IGListDiffOption option);
 
 NS_ASSUME_NONNULL_END
--- a/Source/Common/IGListDiff.h
+++ b/Source/Common/IGListDiff.h
@@ -30,6 +30,47 @@ typedef NS_ENUM(NSInteger, IGListDiffOption) {
 };
 
 /**
+ Comparison behavior options.
+ */
+typedef NS_OPTIONS(NSInteger, IGListDiffBehavior) {
+    IGListDiffBehaviorDefault = 0,
+};
+
+/**
+ Creates a diff using indexes between two collections.
+
+ @param oldArray The old objects to diff against.
+ @param newArray The new objects.
+ @param option   An option on how to compare objects.
+ @param behavior Comparison behaviors.
+
+ @return A result object containing affected indexes.
+ */
+FOUNDATION_EXTERN IGListIndexSetResult *IGListDiffWithBehavior(NSArray<id<IGListDiffable>> *_Nullable oldArray,
+                                                               NSArray<id<IGListDiffable>> *_Nullable newArray,
+                                                               IGListDiffOption option,
+                                                               IGListDiffBehavior behavior);
+
+/**
+ Creates a diff using index paths between two collections.
+
+ @param fromSection The old section.
+ @param toSection   The new section.
+ @param oldArray    The old objects to diff against.
+ @param newArray    The new objects.
+ @param option      An option on how to compare objects.
+ @param behavior    Comparison behaviors.
+
+ @return A result object containing affected indexes.
+ */
+FOUNDATION_EXTERN IGListIndexPathResult *IGListDiffPathsWithBehavior(NSInteger fromSection,
+                                                                     NSInteger toSection,
+                                                                     NSArray<id<IGListDiffable>> *_Nullable oldArray,
+                                                                     NSArray<id<IGListDiffable>> *_Nullable newArray,
+                                                                     IGListDiffOption option,
+                                                                     IGListDiffBehavior behavior);
+
+/**
  Creates a diff using indexes between two collections.
 
  @param oldArray The old objects to diff against.
--- a/Tests/IGListDiffTests.m
+++ b/Tests/IGListDiffTests.m
@@ -306,6 +306,80 @@ static NSArray *sorted(NSArray *arr) {
     IGAssertContains(result.moves, [[IGListMoveIndex alloc] initWithFrom:1 to:3]);
 }
 
+
+- (void)test_whenMovingObjectsIncrementally_thatMovesAreCorrect {
+    NSArray *original = @[@0, @1, @2, @3, @4];
+    NSArray *tests = @[
+        @[@0, @1, @2, @3, @4],
+        @[@4, @3, @2, @1, @0],
+        @[@4, @2, @0, @1, @3],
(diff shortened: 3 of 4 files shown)


antons:
There are is one outstanding issue: Swift tests do not build. Xcode strips `IGListDiffOption` with `0` values. I could completely not replicate the behavior of `NSPointerFunctionsOptions`, where options with `0` values are the default, and are also available in Swift.

antons:
Here’s same problem discussed at [AsyncDisplayKit](https://github.com/facebook/AsyncDisplayKit/issues/2278). While they can replace `ASControlState` with `UIControlState` to piggyback on the exception used by the system frameworks, there’s no existing enum here.

Source/IGListDiff.mm
```diff
@@ -228,29 +237,64 @@ static id IGListDiffing(BOOL returnIndexPaths,
             addIndexToCollection(mInserts, toSection, i);
             runningOffset++;
         } else {
+            oldInsertOffsets[i - runningOffset + deleteOffsets[oldIndex]] = runningOffset;
```
rnystrom: Can you explain the theory behind this? Spent some time this morning wrapping my head around these changes and this is the only thing leaving me puzzled.

Source/IGListDiff.mm
```diff
@@ -228,29 +237,64 @@ static id IGListDiffing(BOOL returnIndexPaths,
...
+                for (NSInteger k = 0; k < newCount; k++) {
+                    currentIndexes[k + moveOffsets[k]] = k;
+                }
+                // update move offsets for moves
+                //  0  1  2 old
+                // +1 +1 -2 moveOffsets
+                //  2  0  1 new
+                // +2 ±0 -2 moveOffsets
+                //  2  1  0 final
+                for (NSInteger j = currentIndex + deltaToFinalIndex; j < currentIndex; j++) {
```
rnystrom: Just noting that these changes appear to make the algorithm _O(n^2)_ (2 for-loops, in fact). Since this behavior really only works for `NSTableView` (or doing your own incremental stuff) it wont impact the core of IGListKit, but its worth calling out.

We've actually stripped the original Least Common Subsequence part of the algorithm where we'd do an ascending and descending pass on the `newResultsArray` to do block move calculations since it didn't play nicely w/ `UICollectionView`. I tried to think up options this morning but came up a little short.

I'll let it marinade a little.

rnystrom:
Btw, its legit that you jumped in and started tackling this problem! Not super easy, but will be really valuable to unlock diffing on another platform. Not to mention I got all the unit tests working and they all passed. Really cool to see the ability to incrementally apply the changes to an array and get the correct results.

I'm STOKED to get this PR in!

jessesquires:
@rnystrom / @antons 

> The more I thought about it, these options aren't really mix-n-match. They are all binary options, so I think a new enum for move results would be more appropriate.

I agree that piggy-backing on `IGListDiffOption` (and making it an option set) doesn't make sense —not to mention, a breaking change. This move reporting doesn't have much to do with the *diff type* either. So I'd suggest a new enum for this:

```swift
enum IGListMoveOption {
    case IGListMoveOptionDefault
    case IGListMoveOptionIncremental
}
```

> Now the tricky part is how to add this without breaking everything. I think we should add a new IGListDiff*(...) function with a param for move options.

I agree, except I'm not sure I understand what @rnystrom suggested. 😅 I would say:

Provide 2 versions of the `IGListDiff*()` functions:

* One that *does not* include this new enum `IGListMoveOption` (i.e., the existing functions)
* One that adds a parameter for `IGListMoveOption`
* The first (original) functions will call the new ones and pass `IGListMoveOptionDefault`

This way, this is a purely additive change and we can put this in a `2.x` release.

If we later decide that this makes the API too large, then we can remove original functions and release `3.x`.

@antons - Also, don't forget to update the changelog. 😄

jessesquires:
@rnystrom - do we want to put this in `2.1.0` ? (probably should since this release adds macOS support)

Source/IGListDiff.mm
```diff
@@ -228,29 +237,64 @@ static id IGListDiffing(BOOL returnIndexPaths,
...
+                if ((oldIndex - deleteOffset + insertOffset) != i) {
+                    addMoveFromIndexToIndex(oldIndex, i);
+                }
+            }
+        }
+
+        addIndexToMap(toSection, i, newArray, newMap);
+    }
 
+    if ((option & IGListDiffMovesIncrementally) != 0) {
```
antons: Will do.

Tests/IGListDiffTests.m
```diff
@@ -306,6 +306,40 @@ - (void)test_whenMovingObjectShiftsOthers_thatMovesContainRequiredMoves {
...
+        @[@4, @0, @2, @3, @1],
+        @[@2, @1, @0, @4, @3],
+        @[@4, @0, @2, @1, @3],
+        @[@4, @2, @0, @1, @3],
+        @[@3, @2, @0, @1, @4],
+        @[@3, @2, @0, @1],
+        @[@0, @1, @2, @3, @4],
+	];
+    for (NSArray *expected in tests) {
+        IGListIndexSetResult *result = IGListDiff(original, expected, IGListDiffEquality|IGListDiffMoveIncrementally);
```
antons: Thanks, will update.

Source/IGListDiff.mm
rnystrom: @antons resurrected from old source. This is just the setup showing the ascend/descend, the actual block part (where `.index` is assigned) probably isn't relevant.

https://gist.github.com/rnystrom/0b78713ff77a74bccbe637a27d0d47cf

antons:
I think than one option set (as opposed to several enums) is the preferred system framework way adjusting the behavior of something (as we can see with `NSPointerFunctionsOptions`, `UIViewAnimationOptions`). It allows grouping and passing all options together in one parameter, and extending it later.

Several enums are a lot less extensible. On our internal copy I have also added IGListDiffNewUpdateLocations to be able to manually refresh the contents of updated cells after a batch update. This could be done with 3 enums, but that would require that another variant of the function would be required. With one option set it’s trivial to add the new option while keeping the same API.

```objc
typedef NS_OPTIONS(NSInteger, IGListDiffOption) {
    IGListDiffEquality = 0 << 0,
    IGListDiffPointerPersonality = 1 << 0,

    IGListDiffDefaultMoves = 0 << 4,
    IGListDiffIncrementalMoves = 1 << 4,
    
    IGListDiffOldUpdateLocations = 0 << 8,
    IGListDiffNewUpdateLocations = 1 << 8,
};
```

jessesquires:
@antons - In general, I agree. 😎  👍 

However, the issue with combining all of this into one option set is that we're conflating two very different concepts:

1. *How* diffing is accomplished, the definition of what equality means when diffing objects. This is a fundamental invariant of the algorithm. (either `-isEqual:` or pointer comparison)
2. The *update behavior* of moves (default vs incremental), old/new location reporting, etc. These are more "client-side API" features to request more granular data.

Furthermore, `DiffEquality & DiffPointerPersonality` is an **invalid** option set. These are mutually exclusive.

---------------------

Having said that, let's take my earlier suggestion and instead of introducing a new enum, let's introduce a new option set. Something like:

```objc
typedef NS_OPTIONS(NSInteger, IGListDiffBehavior) {
    IGListDiffBehaviorNone = 0 << 0,

    IGListDiffBehaviorIncrementalMoves = 0 << 4,
    
    IGListDiffBehaviorOldUpdatesLocations = 0 << 8,
    IGListDiffBehaviorNewUpdatesLocations = 1 << 8,
};
```

And then the original `IGListDiff*()` functions can default to `IGListDiffBehaviorNone`.

This will solve the Swift 3 issue, and give us an extensible interface for adding additional options in the future. It would be awesome if you want to submit a PR for the `Old/NewUpdatesLocations` options, too. 😄 

Thoughts?

rnystrom:
@jessesquires +1 +1 +1

antons:
@jessesquires `NSPointerFunctionsOptions` and `UIViewAnimationOptions` beg to differ. :-) Apple often combines mutually exclusive options in NS_OPTIONS in Foundation, AppKit, and UIKit. It’s not ideal, but there’s plenty of precedent.

I like your suggestion. Defaults for each behavior can be documented with each option.

```objc
typedef NS_OPTIONS(NSInteger, IGListDiffBehavior) {
    IGListDiffBehaviorDefault = 0,
    /// Report moves which can be applied incrementally. By default, old and new positions for each object are reported.
    IGListDiffBehaviorIncrementalMoves = 1 << 0,
    /// Report new locations for updated objects. By default, old locations are reported with updates.
    IGListDiffBehaviorNewUpdatesLocations = 1 << 1,
};
```

jessesquires:
lol not sure Apple is always the right place to look for good architecture/API decisions 😄 😅

Source/IGListDiff.mm
antons: Thanks! To begin with, I’ll set up the current algorithm, and will look into the LCS as time allows.

antons:
@rnystrom @jessesquires I separated the behavior enum and functions. Let me know what you think about the names, whether they should be separated into a new header, and whether you want experiment functions adjusted for the new parameter.

antons:
Rebased to latest changes.

Source/Common/IGListDiff.h
```diff
@@ -30,6 +30,17 @@ typedef NS_ENUM(NSInteger, IGListDiffOption) {
 };
 
 /**
+ Comparison behavior options.
+ */
+typedef NS_OPTIONS(NSInteger, IGListDiffBehavior) {
+    IGListDiffBehaviorDefault = 0,
```
jessesquires: I think naming as `None` will improve Swift inter-op

Source/Common/IGListDiff.h
```diff
@@ -30,6 +30,17 @@ typedef NS_ENUM(NSInteger, IGListDiffOption) {
...
+    /**
+     Report moves which can be applied incrementally. O(n^2) complexity. By default, old and new positions for each object are reported.
```
jessesquires: should be "**Reports** moves...."

> By default, old and new positions for each object are reported.

Can you elaborate on this more? I don't think it's clear what this option provides. 😊

Source/Common/IGListDiff.h
```diff
@@ -59,4 +70,38 @@ FOUNDATION_EXTERN IGListIndexPathResult *IGListDiffPaths(NSInteger fromSection,
...
+ Creates a diff using indexes between two collections.
+
+ @param oldArray The old objects to diff against.
+ @param newArray The new objects.
+ @param option   An option on how to compare objects.
+ @param behavior Comparison behaviors.
+
+ @return A result object containing affected indexes.
+ */
+FOUNDATION_EXTERN IGListIndexSetResult *IGListDiffWithBehavior(NSArray<id<IGListDiffable>> *_Nullable oldArray,
```
jessesquires: can we keep the same names for these new functions? `IGListDiff` and `IGListDiffPaths` ?

Source/Common/IGListDiff.h
antons: @jessesquires I don’t see any differences in the generated Swift interface whether the option is named `default` or `none`.

[Documentation](https://developer.apple.com/library/content/documentation/Swift/Conceptual/BuildingCocoaApps/InteractingWithCAPIs.html#//apple_ref/doc/uid/TP40014216-CH8-NoLink_29) never mentions `none` either:
> When importing C enumeration marked with the NS_OPTIONS macro, Swift marks any members that have a value of 0 as unavailable, because Swift uses an empty option set to specify no options.

Source/Common/IGListDiff.h
jessesquires: ugh, sorry. I'm getting my Swift and ObjC worlds mixed up. 😅 carry on.

antons: @jessesquires I thought so. :-)

antons:
@jessesquires I’ve updated the branch with your feedback. Changelog could be written better, but it’s getting late in Helsinki. I’ll take another look tomorrow. Feel free to edit it if you have any ideas.

antons:
@rnystrom Separated into three tests, let me know if you have an ideas on how to make them better.

Tests/IGListDiffTests.m
```diff
@@ -306,6 +306,64 @@ - (void)test_whenMovingObjectShiftsOthers_thatMovesContainRequiredMoves {
...
+    }];
+    [result.inserts enumerateIndexesUsingBlock:^(NSUInteger idx, BOOL * _Nonnull stop) {
+        [patched insertObject:expected[idx] atIndex:idx];
+    }];
+    for (IGListMoveIndex *move in result.moves) {
+        id object = [patched objectAtIndex:move.from];
+        [patched removeObjectAtIndex:move.from];
+        [patched insertObject:object atIndex:move.to];
+    }
+    XCTAssertEqualObjects(patched, expected);
```
rnystrom: This will make tracking the assert to the test a little trickier. Instead could this be a function that takes in an original array and diff result, then returns an array attempting to apply the changes on it? Then each test does `XCTAssertEqualObject(patched, expected)` and we'll be able to dial in on regressions quicker.

CHANGELOG.md
```diff
@@ -8,9 +8,12 @@ The changelog for `IGListKit`. Also see the [releases](https://github.com/instag
 This release closes the [2.1.0 milestone](https://github.com/Instagram/IGListKit/milestone/2).
 
 ### Enhancements
-
 - Added support for macOS. Note: this is *only* for the Diffing components. There is **no support** for `IGListAdapter`, `IGListSectionController`, and other components at this time. [Guilherme Rambo](https://github.com/insidegui) [(#235)](https://github.com/Instagram/IGListKit/pull/235)
 
+- Added ability to adjust Diffing behavior with `IGListDiffBehavior`, `IGListDiffWithBehavior`, and `IGListDiffPathsWithBehavior`.
+
+- Added algorithm for diffing with incremental moves which can be used with `IGListDiffBehaviorIncrementalMoves`.
```
rnystrom: Make sure to follow the other changelog formats and include a link to your profile (if you want it) and this PR

antons:
@rnystrom Added another test, and it’s failing. Looks like `oldInsertOffsets[i - runningOffset + deleteOffsets[oldIndex]] = runningOffset` is wrong in some cases with simultaneous deletion and insertion.

jessesquires:
Landing #368 now. After that, this is the only task left for `2.1.0`.

I think one last task here would be to update the mac example project. (i.e. Fix #371)

cc @rnystrom @antons

rnystrom:
@jessesquires maybe in the interim we can only claim support for `NSCollectionView` and refactor the example to use that instead? Leave `NSTableView` dependent on this PR.

jessesquires:
@rnystrom - cool with me. so you want to defer this to `2.2.0` ?

if so, I can push `2.1.0` anytime.

rnystrom:
@jessesquires let's defer this til 2.2 then and knock out 2.1

jessesquires:
Any updates here? What's the status of this PR?

jessesquires:
✨ friendly ping ✨ 

Any updates here? What's the status of this PR?

antons:
@jessesquires The issue I mentioned previously is still there. Looks like oldInsertOffsets[i - runningOffset + deleteOffsets[oldIndex]] = runningOffset is wrong in some cases with simultaneous deletion and insertion. I’m unlikely to have time for this for another few weeks. (Anybody who wants to is welcome to step in.)

rnystrom:
I've been thinking about this PR for a while now. IMO to get the changeset that we want requires a totally different diffing algorithm. I think that'd be fine to add to this project, but should probably be a new function like `IGListDiffIncremental(...)`.

Seeing as @antons is low on time (and THANK YOU for all your work here!) let's close this for now and revisit when there is need.

jessesquires:
opened a feature request to track this work: #546

PR #344, opened 2016-12-20
Fix broken links, fix travis

jessesquires:
Caught by travis.
https://travis-ci.org/Instagram/IGListKit/jobs/185367790

Many thanks to @Sherlouk for setting this up. 💯

### DIFF
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -15,7 +15,7 @@ please open an issue on GitHub.
 
 ## Experimental changes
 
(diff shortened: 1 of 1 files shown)


jessesquires:
@Sherlouk - is there any way to have `markdown-link-check` tell you the file? had to `grep` the URL to find out

Sherlouk:
@jessesquires I don't think we can do that as part of markdown-link-check, but the command we use simply iterates over all MD files and runs the link checker. We could probably have it print out the title of the page before it checks the page?

`echo "Checking file: README.md`

PR #374, opened 2016-12-28
Lowered macOS example deployment target to 10.10

Sherlouk:
Worth noting that while we support 10.8 and newer, Swift support is only 10.9 and newer. There were only a few more errors for 10.9 deployment target, but I'm not familiar with how it all works.

Using an `@available` to disable row actions for users on 10.10 (only available for 10.11 and newer).


Sherlouk:
Coveralls being a tad excessive 🤔 

Need to run `pod install` for the macOS examples, but I'll leave that to be done internally as per Jesse's previous comments.

Not sure why it failed then, seems tad random

jessesquires:
After the things I outlined above, let's push a commit with `pod install`. Sometimes this only updates a few files -- as long as it doesn't touch `.xcconfig` and a few other types of files, our importer will work.

If `pod install` touches more than just project files, then we can revert and I'll run `pod install` internally.

Sherlouk:
`pod install` is making changes to all the `.lock` files (which you've mentioned are bad news bears)

Still want me to push?

jessesquires:
@Sherlouk - ah, nope! i'll run internally 😄

PR #379, opened 2016-12-30
Make -[IGListAdapter updater] Public, Read-Only

Adlai-Holler:
Hey IGListKit folks! Would you be willing to expose this, so that other objects that are given a list adapter can inspect its updater configuration?

## Pull request checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,8 @@ This release closes the [2.1.0 milestone](https://github.com/Instagram/IGListKit
 
 - Added `allowsBackgroundReloading` flag (default `YES`) to `IGListAdapterUpdater` so users can configure this behavior as needed. [Adlai-Holler](https://github.com/Adlai-Holler) [(#375)](https://github.com/Instagram/IGListKit/pull/375)
 
+- `-[IGListAdapter updater]` is now public (read-only). [Adlai-Holler](https://github.com/Adlai-Holler) [(#379)](https://github.com/Instagram/IGListKit/pull/379)
+
 ### Fixes
 
 - Avoid `UICollectionView` crashes when queueing a reload and insert/delete on the same item as well as reloading an item in a section that is animating. [Ryan Nystrom](https://github.com/rnystrom) [(#325)](https://github.com/Instagram/IGListKit/pull/325)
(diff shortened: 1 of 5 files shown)


jessesquires:
I'm fine with this change. What's your use case?

> /cc: @rnystrom ?

Adlai-Holler:
I'd like to be able to assert that a list adapter passed into an API does not use `IGListReloadDataUpdater` and that the updater is configured with `allowsBackgroundReloading=NO`. I could modify the API to require the user to keep track of the updater and pass it in alongside the list adapter, but it seems like unnecessary hassle.

jessesquires:
Hmm... but you have to initialize the adapter with an `id<IGListUpdatingDelegate>` couldn't you enforce this then?

Adlai-Holler:
Since the component that receives the adapter is not the one who initializes it, the adapter's owner would have to keep track of the updater separately and provide them both into this API. So `@property ASCollectionNode.listAdapter` would have to become `-[ASCollectionNode setListAdapter:updater:]`.

CHANGELOG.md
```diff
@@ -19,6 +19,8 @@ This release closes the [2.1.0 milestone](https://github.com/Instagram/IGListKit
 
 - Added CocoaPods subspec for diffing, `IGListKit/Diffing` and an [installation guide](https://instagram.github.io/IGListKit/installation.html). [Sherlouk](https://github.com/Sherlouk) [(#368)](https://github.com/Instagram/IGListKit/pull/368)
 
+- `IGListAdapter.updatingDelegate` is now public (read-only). [Adlai-Holler](https://github.com/Adlai-Holler) [(#379)](https://github.com/Instagram/IGListKit/pull/379)
```
rnystrom: Super nit: all of our changelog is in ObjC. Can we change the first part to `-[IGListAdapter updatingDelegate]`

jessesquires:
@rnystrom - want to just import this and forward-fix internally?

Adlai-Holler:
@rnystrom The internal (readwrite) version of the property unfortunately needs to stay, because the tests use it:

screen shot 2017-01-03 at 3 07 40 pm

Adlai-Holler:
Cool. I've renamed `updatingDelegate -> updater` and pushed ✨

Issue #348, opened 2016-12-21
IGListStackedSectionController's children need to know numberOrItems before didUpdate is called


jeffbailey:
## New issue checklist

- [X] I have reviewed the [`README`](https://github.com/Instagram/IGListKit/blob/master/README.md) and [documentation](http://instagram.github.io/IGListKit)
- [X] I have searched [existing issues](https://github.com/Instagram/IGListKit/issues) and this is not a duplicate

### General information

- `IGListKit` version:
- iOS version(s): 10.2
- CocoaPods/Carthage version: master
- Xcode version: 8.2
- Devices/Simulators affected:
- Reproducible in the demo project? (Yes/No): NO
- Related issues:

IGListStackedSectionController is incredibly useful for decomposing section controllers, but I'm running into an issue that appears to be a bug when one of the child section controllers has a variable number of cells, based on the object injected into it.

A child section controller's
`    func numberOfItems() -> Int
`
method is getting called before
`    func didUpdate(to object: Any)
`
That results in the section controller having to know how many items it contains BEFORE it has its data.

I tracked the issue down to the stacked section controller constructor calling it's reloadData method:
```
- (instancetype)initWithSectionControllers:(NSArray <IGListSectionController<IGListSectionType> *> *)sectionControllers {
    if (self = [super init]) {
        for (IGListSectionController<IGListSectionType> *sectionController in sectionControllers) {
            sectionController.collectionContext = self;
            sectionController.viewController = self.viewController;
        }

        _visibleSectionControllers = [[NSCountedSet alloc] init];
        _sectionControllers = [NSOrderedSet orderedSetWithArray:sectionControllers];

        self.displayDelegate = self;
        self.scrollDelegate = self;

        [self reloadData];
    }
    return self;
}
```
The `reloadData` method is asking its children for the number of items, but the children don't have their data injected yet.

I can easily workaround the issue by explicitly setting the data (i.e. view model in my case) on the child section controller when I build up stacked section controller, but that seems like a hack unless I'm missing something.

Let me know what you think.  If you agree its a bug, I'd be happy to create a unit test and fix the issue, but I probably need a little guidance as to the appropriate fix.

rnystrom:
@jeffbailey doing some cleanup, if you're still interested in tackling this, that'd be awesome! As far as guidance goes, my bet is that we:

- Remove [reloadData](https://github.com/Instagram/IGListKit/blob/master/Source/IGListStackedSectionController.m#L61) from the init
[...]

PR #390, opened 2017-01-06
Fix for #348 - IGListStackedSectionController's children need to know numberOrItems before didUpdate is called

jeffbailey:
The reloadData method in IGListStackedSectionController was being called too soon.  It was moved from the constructor to the `didUpdateToObject` method.  Tests were updated to reflect the change.

## Pull request checklist


jessesquires:
Nice. Thanks @jeffbailey ! 💯

Tests/IGListStackSectionControllerTests.m
```diff
@@ -85,8 +85,9 @@ - (void)test_whenInitializingStack_thatNumberOfItemsMatches {
     section4.items = 1;
 
     IGListStackedSectionController *stack = [[IGListStackedSectionController alloc] initWithSectionControllers:@[section1, section2, section3, section4]];
-
-    XCTAssertEqual([stack numberOfItems], 6);
+    [stack performBatchAnimated:false updates:^{} completion:^(BOOL finished) {
+        XCTAssertEqual([stack numberOfItems], 6);
```
rnystrom: These tests use [IGListAdapterUpdater](https://github.com/Instagram/IGListKit/blob/master/Tests/IGListStackSectionControllerTests.m#L57) which asynchronously collect updates during a runloop turn, so this assert wont execute w/out an expectation+wait setup. Here's an [example](https://github.com/Instagram/IGListKit/blob/master/Tests/IGListStackSectionControllerTests.m#L766-L776). Same goes for all the asserts in this change that got moved into the completion block.

A really easy way to check that the block executes is:

```objc
__block BOOL executed = NO;
[adapter performUpdates:^{
  executed = YES;
}];
XCTAssertTrue(executed);
```

jeffbailey: Yep, it should have dawned on me I made the changed tests asynchronous :).  The issue with these particular tests is that they don't follow the normal pattern of the other tests because they don't use the adapter.  I don't know without digging further why they don't use the adapter or if they can be changed easily.  Also, my proposed solution wouldn't work anyway since the completion block never gets called because there's no CollectionContext set on the section controller.  One easy way to use to fix the broken tests would be to use the following pattern:

```
- (void)test_whenInitializingStack_thatNumberOfItemsMatches {
    IGListTestSection *section1 = [[IGListTestSection alloc] init];
    section1.items = 2;
    IGListTestSection *section2 = [[IGListTestSection alloc] init];
    section2.items = 3;
    IGListTestSection *section3 = [[IGListTestSection alloc] init];
    section3.items = 0;
    IGListTestSection *section4 = [[IGListTestSection alloc] init];
    section4.items = 1;

    IGListStackedSectionController *stack = [[IGListStackedSectionController alloc] initWithSectionControllers:@[section1, section2, section3, section4]];
    [stack performSelectorOnMainThread:@selector(reloadData) withObject:nil waitUntilDone:TRUE];
    XCTAssertEqual([stack numberOfItems], 6);
}
```

Basically just call the private reloadData method using performSelector.  Let me know what you think the best way to proceed would be.

jeffbailey: @rnystrom - let me know what you think about just using:
`    [stack performSelectorOnMainThread:@selector(reloadData) withObject:nil waitUntilDone:TRUE];
`
to get the unit tests working.

rnystrom: @jeffbailey oh! I missed that. Good call.

Instead of doing the `performSelector` hack, how about we expose `reloadData` in the internal header and call it directly?

https://github.com/Instagram/IGListKit/blob/master/Source/Internal/IGListStackedSectionControllerInternal.h

The header is already imported to the test.

jessesquires:
Friendly ping ✨ any updates here?

@rnystrom if this is good, we can follow up with a changelog entry separately.

jeffbailey:
The internal header is a great option.  I'll get this done today.  Thanks @rnystrom

rnystrom:
@jeffbailey looks great! Very last thing is a quick note in the `CHANGELOG.md` file about this fix and we're g2g. It should go under the `2.2.0` list.

PR #401, opened 2017-01-09
Test for section controller retain cycles

rnystrom:

Saw our [retain cycle detector](https://github.com/facebook/FBRetainCycleDetector) fire for the `_objectTransitionBlock` in the adapter (t15355965) so I added a test. The test passes, so I don't think that is what's going on, but figured I'd submit the test anyways.

## Pull request checklist


jessesquires:
💯

PR #410, opened 2017-01-13
Fix possible retain cycle during update

rnystrom:

Fixes a retain cycle introduced in 4cc91a25c8b262953e4f2d8e5dc78ee15c6265b2 when a data source is deallocated after queuing an update. The cycle happens when:

- An object holding a strong ref to an `IGListAdapter` is also the data source
- The object is deallocated after calling `performUpdatesAnimated:completion:`

A strong ref to the `dataSource` (e.g. the object w/ a strong ref to the adapter) is captured in the [objectTransitionBlock](https://github.com/Instagram/IGListKit/blob/master/Source/IGListAdapter.m#L271). The adapter holds a strong ref to the updater, which stores the block with a [strong ref](https://github.com/Instagram/IGListKit/blob/master/Source/Internal/IGListAdapterUpdaterInternal.h#L38).

The cycle [is broken](https://github.com/Instagram/IGListKit/blob/master/Source/IGListAdapterUpdater.m#L312) as soon as the update is performed, but if the data source is deallocated, we might never get here.

I'm not totally crazy about this fix as it complicates the API as its only the least invasive. However we have to bail the update if the data source is deallocated since we can't get the correct state or section controllers. Otherwise the `UICollectionView` will throw (what 4cc91a25c8b262953e4f2d8e5dc78ee15c6265b2 intended to fix).

This diff breaks the cycle by removing the [strongly held data source in the block](https://github.com/Instagram/IGListKit/blob/master/Source/IGListAdapter.m#L271).

Found w/ our retain cycle detector, flagged in tasks t15355965, t15271777, t15271776

## Pull request checklist

### DIFF
--- a/Source/IGListUpdatingDelegate.h
+++ b/Source/IGListUpdatingDelegate.h
@@ -24,8 +24,10 @@ typedef void (^IGListUpdatingCompletion)(BOOL finished);
  A block to be called when the adapter applies changes to the collection view.
 
  @param toObjects The new objects in the collection.
+
+ @return A flag indicating if the transition was applied.
  */
-typedef void (^IGListObjectTransitionBlock)(NSArray *toObjects);
+typedef BOOL (^IGListObjectTransitionBlock)(NSArray *toObjects);
 
 /// A block that contains all of the updates.
 typedef void (^IGListItemUpdateBlock)();
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,8 @@ This release closes the [2.2.0 milestone](https://github.com/Instagram/IGListKit
 
 - Fix bug where emptyView's hidden status is not updated after the number of items is changed with `insertInSectionController:atIndexes:` or related methods. [Peter Edmonston](https://github.com/edmonston) [(#395)](https://github.com/Instagram/IGListKit/pull/395)
 
+- Fixes a possible retain cycle when the data source is deallocated after queueing an update. [Ryan Nystrom](https://github.com/rnystrom) (tbd)
+
 2.1.0
 -----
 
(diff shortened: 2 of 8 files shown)


rnystrom:
CI working, UI tests died on one of the builds

rnystrom:
So the cycle exists, but I've confirmed that it is **always** broken via:

- [Update queued](https://phabricator.intern.facebook.com/diffusion/FBS/browse/master/fbobjc/VendorLib/IGListKit/src/Source/IGListAdapterUpdater.m;daea734c4206a303f2df6c896054243ba76e7c3e$407)
- [Block dispatched](https://phabricator.intern.facebook.com/diffusion/FBS/browse/master/fbobjc/VendorLib/IGListKit/src/Source/IGListAdapterUpdater.m;daea734c4206a303f2df6c896054243ba76e7c3e$336-345,335)
- Both batch updates and reload methods [cleanup their state](https://phabricator.intern.facebook.com/diffusion/FBS/browse/master/fbobjc/VendorLib/IGListKit/src/Source/IGListAdapterUpdater.m;daea734c4206a303f2df6c896054243ba76e7c3e$312)

I'm going to think more about how this cycle could be explicitly broken, but it does serve its purpose of not letting the data source disappear when an update has to happen. Though this diff solves that by just stopping the update.

Source/IGListUpdatingDelegate.h
```diff
@@ -24,8 +24,10 @@ typedef void (^IGListUpdatingCompletion)(BOOL finished);
  A block to be called when the adapter applies changes to the collection view.
 
  @param toObjects The new objects in the collection.
+
+ @return A flag indicating if the transition was applied.
  */
-typedef void (^IGListObjectTransitionBlock)(NSArray *toObjects);
+typedef BOOL (^IGListObjectTransitionBlock)(NSArray *toObjects);
```
rnystrom: Good point, it is.

rnystrom:
Talked w/ @jeremycohen offline, going to abandon this since the cycle is deliberate and always cleaned up.

PR #411, opened 2017-01-13
didDeselect for IGListAdapter and IGListSingleSectionControllerDelegate

danielrhodes:
This implements the `didDeselectItemAt:` delegate call from `UICollectionViewDelegate`. It mostly does the exact same thing as the `didSelect` calls which already exist. Tests are also based on the select tests.

- Implements the `UICollectionViewDelegate` `didDeselectItemAt:` delegate call in `IGListAdapter`
- Adds `didDeselectSectionController:withObject:` delegate call to `IGListSingleSectionControllerDelegate`
- Tests for changes
- Updates example projects

## Pull request checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,8 @@ This release closes the [3.0.0 milestone](https://github.com/Instagram/IGListKit
                         withObject:(id)object;
 ```
 
+- Added `didDeselectItemAtIndex` delegate call in `IGListSectionType` and `didDeselect` delegate call in `IGListSingleSectionControllerDelegate`. [Daniel Rhodes](https://github.com/danielrhodes) [(#411)](https://github.com/Instagram/IGListKit/pull/411)
+
 2.2.0
 -----
 
--- a/Examples/Examples-iOS/IGListKitExamples/SectionControllers/DemoSectionController.swift
+++ b/Examples/Examples-iOS/IGListKitExamples/SectionControllers/DemoSectionController.swift
@@ -80,5 +80,7 @@ final class DemoSectionController: IGListSectionController, IGListSectionType {
             viewController?.navigationController?.pushViewController(controller, animated: true)
         }
     }
+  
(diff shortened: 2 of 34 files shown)


rnystrom:
So we actually used to have this API but removed it b/c nobody ever used it. Can I ask why we'd want to add this back?

Another thing we've discussed, since this method will rarely ever get used, it'll add a lot of boilerplate (especially once we land internally). We've talked about adding a `selectionDelegate` instead and moving the `didSelectItemAtIndex:` method there along w/ a deselect method.

Thoughts?

danielrhodes:
Well personally I need it to store the selected state on a cell so I don't have to maintain state elsewhere (imagine a list with checkmarks). I also need to be able to set a cell as selected/unselected, which would be ideally done through the section controller based on object. Open to other patterns that might solve this, but using what is built in seems to be the cleanest way. More generally to your point, this library is wrapping UICollectionView with a better API, but I don't necessarily think that means it should have a reduced feature set.

I am fine with a `selectionDelegate` since Swift is not friendly towards optional protocol methods. For `IGListSingleSectionControllerDelegate`, I would rather have something like `sectionController:withObject:selected:` to keep it to one method.

rnystrom:
Using `UICollectionView` internals to track data selection states is a #GreatDebate here at Instagram, but generally we've found there to be too many complexities w/ the inner workings of UICV to be able to fully trust it.

The `didSelectItemAtIndex:selected:` or something API is really interesting though. 🤔

jessesquires:
✨ friendly ping ✨ 

what's the status of this @rnystrom ?

if we don't want to accept, we should close.

danielrhodes:
@jessesquires After the convo above, I took another look at my own code, which this pull request was supposed to solve, and made some changes so that I no longer needed this patch. So if you want to close the PR, I am indifferent.

jessesquires:
Thanks @danielrhodes ! 💯

Issue #168, opened 2016-11-06
Remove IGListSectionType protocol and absorb methods into IGListSectionController superclass


yusuftor:
It's not possible to access the numberOfItems() function on an IGListSectionController unless it is type cast to a specific IGListSectionController subclass. Is there a reason for this?

For example:

`let sectionController = adapter.sectionController(for: content.last!)`
`let numberOfItems = sectionController.numberOfItems()` <- Doesn't work

`let numberOfItems = (sectionController as! SectionControllerSubclass).numberOfItems()` <- Works

rnystrom:
@yusuftor I wonder if Swift interop is messing up here. The interface is declared as:

```objc
- (__kindof IGListSectionController <IGListSectionType> * _Nullable)sectionControllerForObject:(id)object;
```

I wonder if the `__kindof` is where it breaks. Maybe we remove that?

yusuftor:
I'm not strong on Objective-C, but I guess it must be a Swift interop problem if it works fine in Objective-C. The other protocol functions aren't accessible either, but the properties seem to be accessible, e.g. scrollDelegate, workingRangeDelegate.

jessesquires:
@rnystrom -- I'm gonna say this is a bug. Let's see if we can adjust the return type to be more Swift-friendly.

zhubofei:
@rnystrom Seems that a type can't be both a protocol and a concrete type at the same time in Swift. Is it possible to expose `IGListSectionType` as a class instead of a protocol (don't know if that make sense at all)?

jessesquires:
Ah, right. @zhubofei is correct.

Currently, Swift does not have a way to express "concrete type + protocol". 

I think we'll have to live with this for now.

jessesquires:
I'm going to leave this open as a bug for now for future reference.

jessesquires:
Actually, going to close this since there's nothing actionable.

This won't be expressible in Swift for awhile. We can revisit then -- if ever.

zhubofei:
@rnystrom @jessesquires I came across this [issue](https://github.com/spotify/HubFramework/issues/158) when I went through the code of Spotify's HubFramework today. I was wondering if we can use the same approach by merging  `IGListSectionType` into `IGListSectionController` class?

jessesquires:
@zhubofei yeah, it's an interesting idea.

Our goals with this design were:

1. Avoid having a bunch of empty methods in a superclass with asserts (gross)
[...]

PR #412, opened 2017-01-13
hacky WIP proof-of-concept for #168

jessesquires:
## ⚠️ Do not merge ⚠️ 

See discussion at https://github.com/Instagram/IGListKit/issues/168#issuecomment-272355012

This is quick and dirty. 💩 

Some notes:

1. This changes the way APIs bridge over to Swift (so it's massively breaking for Swift code). ([example](https://github.com/Instagram/IGListKit/pull/412/files#diff-577b9e71dc7472a72e6d9c335420a9cbR63))
2. One issue with this new protocol approach is handling the *old* internal header for [IGListSectionController](https://github.com/Instagram/IGListKit/blob/master/Source/Internal/IGListStackedSectionControllerInternal.h). Now that we have a public `IGListSectionControllerProtocol`, we need an equivalent private protocol that allows us to set `collectionContext`, `viewController`, `isFirstSection`, `isLastSection`. Not sure how to do this at the moment.


Examples/Examples-iOS/IGListKitExamples/ViewControllers/DisplayViewController.swift
```diff
@@ -27,6 +27,11 @@ class DisplayViewController: UIViewController, IGListAdapterDataSource {
         view.addSubview(collectionView)
         adapter.collectionView = collectionView
         adapter.dataSource = self
+
+
+        let x = adapter.sectionController(for: self)
+        let _ = x.numberOfItems() // works!
```
jessesquires: here's where the ✨ magic ✨ happens

Issue #402, opened 2017-01-10
`-[IGListIndexSetResult hasChanges]` should instead be a readonly property


mdznr:
`-[IGListIndexSetResult hasChanges]` should instead be a readonly property.

IGListIndexSetResult.h:46 is currently:
```objc
- (BOOL)hasChanges;
```

it should instead be:
```objc
@property (nonatomic, assign, readonly) BOOL hasChanges;
```

rnystrom:
I'm down with this. Not breaking for ObjC, breaking for Swift tho. But I think it makes more sense as a computed variable for Swift rather than a method. @jessesquires on board?

jessesquires:
@rnystrom - Sounds good.

We have other breaking changes coming for 3.0.

mdznr:
Thanks for the quick response!

PR #415, opened 2017-01-13
[IGListIndexSetResult hasChanges]` should instead be a readonly property #402

AdamRobertsEF:
Changed method to read only property.

### DIFF
--- a/Source/Common/IGListIndexSetResult.h
+++ b/Source/Common/IGListIndexSetResult.h
@@ -29,7 +29,7 @@ NS_ASSUME_NONNULL_BEGIN
 @property (nonatomic, strong, readonly) NSIndexSet *deletes;
 
 /**
- The indexes in the new collection that need updated.
+ The indexes in the old collection that need updated.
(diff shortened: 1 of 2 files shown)


Source/Common/IGListIndexSetResult.h
```diff
@@ -43,7 +43,7 @@ NS_ASSUME_NONNULL_BEGIN
 
  @return `YES` if the result has changes, `NO` otherwise.
  */
-- (BOOL)hasChanges;
+@property (nonatomic, assign, readonly) BOOL hasChanges; // Fix For Git Issue #402
```
rnystrom: Can you remove the code comment?

AdamRobertsEF: Yes

jessesquires:
this is breaking

AdamRobertsEF:
@rnystrom I have some github app issue and reforked. I didn't expect it to close the commit, since you already imported.

PR #419, opened 2017-01-15
Short circuit diffing if old array or new array is empty

rjchatfield:

Avoid calculating moves and updates.
- If `oldArray` is empty, return insert new rows.
- If `newArray` is empty, return deletes for old rows

Note: This refactor defines local properties and blocks earlier in function. But no logic was changed in these were changed.

## Pull request checklist

- [X] I added an entry to the `CHANGELOG.md` for any breaking changes, enhancements, or bug fixes.


rnystrom:
Interesting! @rjchatfield what's the inspiration behind this? Did you find a bottleneck when diffing old/new arrays?

CHANGELOG.md
```diff
@@ -20,6 +20,10 @@ This release closes the [3.0.0 milestone](https://github.com/Instagram/IGListKit
                         withObject:(id)object;
 ```
 
+### Enhancements
+
+- Short circuit diffing if old array or new array is empty.
```
jessesquires: this should go under 2.2.0

rjchatfield: Fixed!

rjchatfield:
@rnystrom We had our own diffing implementation which included this optimistic short circuit. It was a pretty common case to go from an empty list to a full list (init state -> loading cached state). So it just made sense for us to improve it, and avoid looping over every item and looking it up in a lookup table.

jessesquires:
✨ friendly ping ✨ 

what's the status of this @rnystrom ?

if we don't want to accept, we should close.

rnystrom:
Going to close for now since IMO we should have as little path-forking as possible. If someone can prove me wrong w/ some profiling on how this improves performance please do!

PR #442, opened 2017-01-22
Added `-[IGListAdapter visibleCellsForObject:]` API

### LINKED ISSUE #437: Add -[IGListAdapter visibleCellsForObject:] API
Got a request internally for this API. It's doable now, but more work than necessary:

- Get the section of an object with `sectionForObject:`
- Create an `NSIndexPath`
- Get the cell with `cellForRowAtIndexPath:`

Would be nice to just grab all visible cells that exist for a given object.

Sherlouk:

- Added `-[IGListAdapter visibleCellsForObject:]` API

## Approach

1. Get all visible cells
2. Get the section for the object
3. Filter cells where indexPath.section of cell is equal to the object
4. Return

## Pull request checklist

- [x] I added test(s), an experiment, or detailed why my change isn't tested.

-------------

Purely additive, can go in 2.2.0 release? Hold off on changelog until agreed.
Wanting to learn some more ObjC so gave this a shot, would appreciate some nit picking/feedback

Closes #437

### DIFF
--- a/Source/IGListAdapter.h
+++ b/Source/IGListAdapter.h
@@ -197,6 +197,8 @@ IGLK_SUBCLASSING_RESTRICTED
 /**
  An unordered array of the currently visible cells for a given object.
  
+ @param item An object in the list
+ 
  @return An array of collection view cells.
  */
 - (NSArray<UICollectionViewCell *> *)visibleCellsForObject:(id)item;
--- a/Source/IGListAdapter.h
+++ b/Source/IGListAdapter.h
@@ -197,11 +197,11 @@ IGLK_SUBCLASSING_RESTRICTED
 /**
  An unordered array of the currently visible cells for a given object.
  
- @param item An object in the list
+ @param object An object in the list
  
  @return An array of collection view cells.
  */
-- (NSArray<UICollectionViewCell *> *)visibleCellsForObject:(id)item;
+- (NSArray<UICollectionViewCell *> *)visibleCellsForObject:(id)object;
 
 /**
  Scrolls to the sepcified object in the list adapter.
--- a/Source/IGListAdapter.h
+++ b/Source/IGListAdapter.h
@@ -195,6 +195,13 @@ IGLK_SUBCLASSING_RESTRICTED
 - (NSArray *)visibleObjects;
 
 /**
+ An unordered array of the currently visible cells for a given object.
+ 
+ @return An array of collection view cells.
+ */
+- (NSArray<UICollectionViewCell *> *)visibleCellsForObject:(id)item;
+
+/**
  Scrolls to the sepcified object in the list adapter.
 
  @param object             The object to which to scroll.
--- a/Source/IGListAdapter.m
+++ b/Source/IGListAdapter.m
@@ -414,6 +414,23 @@
     return [visibleObjects allObjects];
 }
 
+- (NSArray<UICollectionViewCell *> *)visibleCellsForObject:(id)item {
+    IGAssertMainThread();
+    IGParameterAssert(item != nil);
+    
+    NSArray<UICollectionViewCell *> *visibleCells = [self.collectionView visibleCells];
+    NSInteger section = [self.sectionMap sectionForObject:item];
+    IGAssert(section != NSNotFound, @"Section not found for object %@", item);
+    
+    NSPredicate *controllerPredicate = [NSPredicate predicateWithBlock:^BOOL(UICollectionViewCell* cell, NSDictionary* bindings) {
+        NSIndexPath *indexPath = [self.collectionView indexPathForCell:cell];
+        return indexPath.section == section;
+    }];
+    
+    NSArray<UICollectionViewCell *> *filteredCells = [visibleCells filteredArrayUsingPredicate:controllerPredicate];
+    return filteredCells;
+}
+
 
 #pragma mark - Layout
 
--- a/Source/IGListAdapter.m
+++ b/Source/IGListAdapter.m
@@ -414,21 +414,23 @@
     return [visibleObjects allObjects];
 }
 
-- (NSArray<UICollectionViewCell *> *)visibleCellsForObject:(id)item {
+- (NSArray<UICollectionViewCell *> *)visibleCellsForObject:(id)object {
     IGAssertMainThread();
-    IGParameterAssert(item != nil);
+    IGParameterAssert(object != nil);
+    
+    const NSInteger section = [self.sectionMap sectionForObject:object];
+    if (section == NSNotFound) {
+        return [NSArray new];
+    }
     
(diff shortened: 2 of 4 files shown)


Sherlouk:
Turns out there's already a method for `-[IGListCollectionContext visibleCellsForSectionController:]` 

https://instagram.github.io/IGListKit/Protocols/IGListCollectionContext.html#/c:objc(pl)IGListCollectionContext(im)visibleCellsForSectionController:

https://github.com/Instagram/IGListKit/blob/master/Source/IGListAdapter.m#L777-L788

Considering you can quite easily get the section controller from object using `-[IGListAdapter sectionControllerForObject:]` it doesn't seem too difficult to do this already?

@rnystrom Suggestions?

rnystrom:
@Sherlouk doh. Ya, maybe we should just recommend people use that API instead...

Sherlouk:
@rnystrom Was your API request, if you think the need is no longer necessary then close this and that issue 😄

rnystrom:
Lol I know, forgot that API existed. Might still be valuable to cut out the middle-layer steps. Could also return early if `sectionController` is `nil`.

Sherlouk:
Sorry, lost track here...

We saying in the old API to return early?
What steps do we want to take here

(Also pardon the test name 😂, only just realised I didn't change it)

rnystrom:
The old API will end up [asserting](https://github.com/Instagram/IGListKit/blob/master/Source/IGListAdapter.m#L340) if the section controller is nil. If you used this method by first getting the object, then the SC (in ObjC) you might assert:

```objc
id controller = [self.adapter sectionControllerForObject:object]; // can be nil
NSArray *cells = [self.adapter visibleCellsForSectionController:controller]; // assert if nil
```

Which means all your callsites have to look like:

```objc
id controller = [self.adapter sectionControllerForObject:object]; // can be nil
if (controller != nil) {
  NSArray *cells = [self.adapter visibleCellsForSectionController:controller]; // assert if nil
}
```

Which is kind of gross. Swift will be just as bad. Options are:

- Remove the assert
- Add `visibleCellsForObject:` method that returns early if a section controller isn't found for the object

We should also add a unit test that sends `nil` to `visibleCellsForSectionController:` even tho the param marked `nonnull`, in ObJC-land that can still happen and the compiler wont help you.

rnystrom:
@Sherlouk ok thought about it more. Let's do option 2 and move forward w/ this for 2.2 if you think that's doable 😄

Source/IGListAdapter.h
```diff
@@ -195,6 +195,15 @@ IGLK_SUBCLASSING_RESTRICTED
 - (NSArray *)visibleObjects;
 
 /**
+ An unordered array of the currently visible cells for a given object.
+ 
+ @param item An object in the list
+ 
+ @return An array of collection view cells.
+ */
+- (NSArray<UICollectionViewCell *> *)visibleCellsForObject:(id)item;
```
rnystrom: nit: `object` as param name

Source/IGListAdapter.m
```diff
@@ -414,6 +414,23 @@ - (NSArray *)visibleObjects {
...
+- (NSArray<UICollectionViewCell *> *)visibleCellsForObject:(id)item {
+    IGAssertMainThread();
+    IGParameterAssert(item != nil);
+    
+    NSArray<UICollectionViewCell *> *visibleCells = [self.collectionView visibleCells];
+    NSInteger section = [self.sectionMap sectionForObject:item];
+    IGAssert(section != NSNotFound, @"Section not found for object %@", item);
+    
+    NSPredicate *controllerPredicate = [NSPredicate predicateWithBlock:^BOOL(UICollectionViewCell* cell, NSDictionary* bindings) {
+        NSIndexPath *indexPath = [self.collectionView indexPathForCell:cell];
```
rnystrom: nit: calling `self.collectionView` inside this block will execute an extra `objc_msgSend(...)` every execution. Instead let's capture the collection view in a local variable that the block uses:
```objc
UICollectionView *collectionView = self.collectionView;
NSPredicate *predicate = [... ^{
  NSIndexPath *path = [collectionView indexPathForCell:cell];
}];
```

Source/IGListAdapter.m
```diff
@@ -414,6 +414,23 @@ - (NSArray *)visibleObjects {
     return [visibleObjects allObjects];
 }
 
+- (NSArray<UICollectionViewCell *> *)visibleCellsForObject:(id)item {
+    IGAssertMainThread();
+    IGParameterAssert(item != nil);
+    
+    NSArray<UICollectionViewCell *> *visibleCells = [self.collectionView visibleCells];
+    NSInteger section = [self.sectionMap sectionForObject:item];
+    IGAssert(section != NSNotFound, @"Section not found for object %@", item);
```
rnystrom: Actually let's remove this. A use might be holding a ref to an expired object. I think it's ok to return an empty array in that case.

Source/IGListAdapter.m
```diff
@@ -414,6 +414,23 @@ - (NSArray *)visibleObjects {
     return [visibleObjects allObjects];
 }
 
+- (NSArray<UICollectionViewCell *> *)visibleCellsForObject:(id)item {
+    IGAssertMainThread();
+    IGParameterAssert(item != nil);
+    
+    NSArray<UICollectionViewCell *> *visibleCells = [self.collectionView visibleCells];
+    NSInteger section = [self.sectionMap sectionForObject:item];
```
rnystrom: nit: `const`

Sherlouk:
Option 2 is good, want to raise a new issue to track progress on amending old functionality?
Will add another unit test as well testing if an object is not found that it returns empty

Source/IGListAdapter.m
```diff
@@ -414,6 +414,23 @@ - (NSArray *)visibleObjects {
...
+    IGAssert(section != NSNotFound, @"Section not found for object %@", item);
+    
+    NSPredicate *controllerPredicate = [NSPredicate predicateWithBlock:^BOOL(UICollectionViewCell* cell, NSDictionary* bindings) {
+        NSIndexPath *indexPath = [self.collectionView indexPathForCell:cell];
+        return indexPath.section == section;
+    }];
+    
+    NSArray<UICollectionViewCell *> *filteredCells = [visibleCells filteredArrayUsingPredicate:controllerPredicate];
+    return filteredCells;
```
rnystrom: nit: IMO just `return [visibleCells filteredArrayUsingPredicate:controllerPredicate];`

Sherlouk:
All valid comments, will work on it tonight!

Sherlouk:
@rnystrom Just updated, should address all of your points!

Not sure if this is the best code to return an empty array, but it does seem to work?
```objc
return [NSArray new];
```

PR #443, opened 2017-01-22
Add APIs for invalidating single and many items in section controller

### LINKED ISSUE #360: IGListCollectionContext API to invalidate layout of a section controller
We need an API to:

- Invalidate all cells in a section controller
- Invalidate _specific_ cells in a section controler

See #346 for more details.

Sherlouk:

- Added `-[IGListCollectionContext invalidateLayoutForIndexPaths]`
- Added `-[IGListCollectionContext invalidateLayoutForSectionController]`

Unsure the best way to test these new methods, so a suggestion would be welcomed!

Closes #360 

## Pull request checklist

### DIFF
--- a/Source/IGListAdapter.m
+++ b/Source/IGListAdapter.m
@@ -1016,17 +1016,17 @@
 }
 
 - (void)invalidateLayoutForSectionController:(IGListSectionController<IGListSectionType> *)sectionController {
-    IGAssertMainThread()
+    IGAssertMainThread();
     IGParameterAssert(sectionController != nil);
     
-    NSInteger *itemCount = [sectionController numberOfItems];
+    NSInteger itemCount = [sectionController numberOfItems];
     NSIndexSet *indexPathRange = [NSIndexSet indexSetWithIndexesInRange:NSMakeRange(0, itemCount)];
     NSArray *indexPaths = [self indexPathsFromSectionController:sectionController indexes:indexPathRange adjustForUpdateBlock:NO];
     [self invalidateLayoutForIndexPaths:indexPaths];
 }
 
 - (void)invalidateLayoutForIndexPaths:(NSArray<NSIndexPath *> *) indexPaths {
-    IGAssertMainThread()
+    IGAssertMainThread();
     IGParameterAssert(indexPaths != nil);
     
     UICollectionViewLayoutInvalidationContext *context = [UICollectionViewLayoutInvalidationContext new];
--- a/Source/IGListAdapter.m
+++ b/Source/IGListAdapter.m
@@ -1015,19 +1015,12 @@
     [self.collectionView scrollToItemAtIndexPath:indexPath atScrollPosition:scrollPosition animated:animated];
 }
 
-- (void)invalidateLayoutForSectionController:(IGListSectionController<IGListSectionType> *)sectionController {
+- (void)invalidateLayoutForSectionController:(IGListSectionController<IGListSectionType> *)sectionController
+                                   atIndexes:(NSIndexSet *)indexes {
     IGAssertMainThread();
     IGParameterAssert(sectionController != nil);
     
-    NSInteger itemCount = [sectionController numberOfItems];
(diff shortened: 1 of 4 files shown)


Sherlouk:
Was building against macOS scheme so Xcode wasn't flagging errors, should be fixed now.

Source/IGListCollectionContext.h
```diff
@@ -253,6 +253,20 @@ NS_ASSUME_NONNULL_BEGIN
...
+ @param sectionController The section controller.
+ */
+- (void)invalidateLayoutForSectionController:(IGListSectionController<IGListSectionType> *)sectionController;
+
+/**
+ Invalidates the collection view layout for the items at the given index paths.
+ 
+ @param indexPaths The index path array to invalidate.
+ */
+- (void)invalidateLayoutForIndexPaths:(NSArray<NSIndexPath *> *) indexPaths;
```
rnystrom: Since section controllers shouldn't know about index paths, can we change this to:
```objc
- (void)invalidateLayoutForSectionController:(IGListSectionController<IGListSectionType> *)sectionController indexes:(NSIndexSet *)indexes;
```

I wonder, maybe we should just have a single API for this?

Source/IGListAdapter.m
```diff
@@ -1015,6 +1015,25 @@ - (void)scrollToSectionController:(IGListSectionController<IGListSectionType> *)
     [self.collectionView scrollToItemAtIndexPath:indexPath atScrollPosition:scrollPosition animated:animated];
 }
 
+- (void)invalidateLayoutForSectionController:(IGListSectionController<IGListSectionType> *)sectionController {
+    IGAssertMainThread();
+    IGParameterAssert(sectionController != nil);
+    
+    NSInteger itemCount = [sectionController numberOfItems];
```
rnystrom: nit: `const`

Source/IGListAdapter.m
```diff
@@ -1015,6 +1015,25 @@ - (void)scrollToSectionController:(IGListSectionController<IGListSectionType> *)
...
+    [self invalidateLayoutForIndexPaths:indexPaths];
+}
+
+- (void)invalidateLayoutForIndexPaths:(NSArray<NSIndexPath *> *) indexPaths {
+    IGAssertMainThread();
+    IGParameterAssert(indexPaths != nil);
+    
+    UICollectionViewLayoutInvalidationContext *context = [UICollectionViewLayoutInvalidationContext new];
+    [context invalidateItemsAtIndexPaths:indexPaths];
+    [self.collectionView.collectionViewLayout invalidateLayoutWithContext:context];
```
rnystrom: Noting to myself that this is an iOS 8+ API, so we're covered w/ our min API

Source/IGListAdapter.m
Sherlouk: Don't mind adding this, but for the sake of learning -- Is putting const here the same as using 'let' in Swift? "A non mutable version"

rnystrom: yup!

Source/IGListCollectionContext.h
Sherlouk: Will reduce down to just a single API, if a user wants to reset the whole section then it's not that difficult to create an index set for the entire range

Source/IGListAdapter.m
Sherlouk: Removed the need for this variable by having a single API, though thanks for explaining it to me!

Sherlouk:
Will work on the tests ASAP, as you say I think a bit more work is needed just to create a different sort of test object that is usable for more complicated tests.

Amateur hour over here, but fun nonetheless!

rnystrom:
@Sherlouk no hurry on this 😄 we don't need to bundle it w/ 2.2, take your time. 

> fun nonetheless!

🙌

Sherlouk:
Going to close this PR for the time being until I get a chance to work on this piece, will reopen then!

Issue #288, opened 2016-12-05
Bug when a section reload is queued while batch updating


rnystrom:
I discovered this working on my sample weather app. Here's the conditions that I noticed to reproduce this:

- Add a `Dispatch.async { ... }` inside `didUpdate(to object:)`
- In the block, call `collectionContext.reload(section: self)`
  - Note **do not** use batch updating

Result: reload will never be called. 

This is because `-[UICollectionView performBatchUpdates:completion:]`, is asynchronous in that it executes over several runloop turns (aggregation and animations).

We use a [flag](https://github.com/Instagram/IGListKit/blob/cdc796746adf95d9a9ae78379a29cddf50c9f738/Source/Internal/IGListAdapterUpdaterInternal.h#L44) to monitor when batch updating, setting it to [YES](https://github.com/Instagram/IGListKit/blob/8ea5e088809b6ea3fe0335d3fdd59f8c77301574/Source/IGListAdapterUpdater.m#L280-L281) _before_ we enter the update block, then setting it to [NO](https://github.com/Instagram/IGListKit/blob/8ea5e088809b6ea3fe0335d3fdd59f8c77301574/Source/IGListAdapterUpdater.m#L289) in the **completion block** of `performBatchUpdates:completion:`.

If a reload is added while this flag is set to `YES`, [we assume](https://github.com/Instagram/IGListKit/blob/master/Source/IGListAdapterUpdater.m#L481) it is part of unloading our blocks of batched operations.

The problem is anyone can queue a reload from anywhere. When the current batch updates finish, we [queue](https://github.com/Instagram/IGListKit/blob/master/Source/IGListAdapterUpdater.m#L203) another update to clear anything we might have collected in the meanwhile. However [before we update](https://github.com/Instagram/IGListKit/blob/master/Source/IGListAdapterUpdater.m#L328) we check if `hasChanges` is true.

Unfortunately the bug lies in that we [don't check](https://github.com/Instagram/IGListKit/blob/master/Source/IGListAdapterUpdater.m#L48-L510) reloads in `hasChanges`.

Some options for fixing:

- Check `self.reloadSections.count > 0` in `hasChanges`
  - Quickest fix
- Replace `batchUpdateOrReloadInProgress` with a "mode" enum to model states of `performBatchUpdates:completion:`:
  - Just before calling
  - Enter/exit update block
  - Enter/exit completion block
  - Finished (technically on completion block exit)

[...]

PR #418, opened 2017-01-13
Add item-level moves to IGListCollectionContext

rnystrom:
## Changes in this pull request

Adding an API to do item-level (cell) moves on the collection view. This complicates things a little bit because of all the issues that moving sections have while in batch updates (e.g. simultaneous animation UICV bugs). Thankfully we use pretty strict types so the compiler does most of the work for us.

Closes #145 

## TODO

- [x] Tests build and pass
- [x] Add `IGListBatchUpdateData` tests to check moves during
  - [x] ~~Moving within a reloaded section (no op)~~ can't reload sections
  - [x] Moving within a deleted section (no op)
  - [x] Moving within a moved section (convert section ops to delete+insert)
  - [x] Moving an index path that is also reloaded (convert to delete+insert path)
- [x] Add move unit tests to `IGListAdapterUpdater`
- [x] Add move unit tests to `IGListReloadDataUpdater` (mostly for code coverage...)
- [x] Add move unit tests to `IGListStackedSectionController`
- [x] Add `CHANGELOG.md` entry for 3.0.0
- [x] Test moving without batch updates

Issue #392, opened 2017-01-07
RFC: Remove item-level updates from IGListCollectionContext, require them done in batch updates


rnystrom:
It is way too easy to execute an item-level update out-of-sync with coalesced batch updates and break things. We've observed this in Instagram. Although rare, at our scale its still a measurable problem.

By item-level updates, I'm specifically talking about:

- [Reload items](https://github.com/Instagram/IGListKit/blob/master/Source/IGListCollectionContext.h#L176-L183)
- [Insert items](https://github.com/Instagram/IGListKit/blob/master/Source/IGListCollectionContext.h#L185-L192)
- [Delete items](https://github.com/Instagram/IGListKit/blob/master/Source/IGListCollectionContext.h#L194-L201)
- [Reload](https://github.com/Instagram/IGListKit/blob/master/Source/IGListCollectionContext.h#L203-L208) the whole section controller

## The Problem

The preferred way to use them is to [batch update](https://github.com/Instagram/IGListKit/blob/master/Source/IGListCollectionContext.h#L210-L240) your **state** (e.g. w/e powers `numberOfItems`) and **item** changes:

```objc
[self.collectionContext performBatchAnimated:YES updates:^{
  NSMutableArray *models = [self.models mutableCopy];
  [models removeObjectAtIndex:2];
  self.models = models;
  [self.collectionContext deleteInSectionController:self 
    atIndexes:[NSIndexSet indexSetWithIndex:2]];
} completion: nil]
```

However I've seen lots of examples not-using the batch update block:

```objc
NSMutableArray *models = [self.models mutableCopy];
[models removeObjectAtIndex:2];
self.models = models;
[self.collectionContext deleteInSectionController:self 
  atIndexes:[NSIndexSet indexSetWithIndex:2]];
```

We even do this in one of [the examples](https://github.com/Instagram/IGListKit/blob/master/Examples/Examples-iOS/IGListKitExamples/SectionControllers/ExpandableSectionController.swift#L45-L46).

The problem w/ this is that if a batch update is in progress, you might update while the `UICollectionViewData` (private structure) is changing the state of the `UICollectionView`. While this is all main-thread affined, the whole operation happens over multiple runloop turns, so it should be treated as an async operation (e.g. racing).

## Proposal

I propose we remove all of the item-update operations from `IGListCollectionContext` so they cannot be done w/out batch updates. Instead, we provide a new `IGListBatchUpdateContext` object as a parameter to the update block that has all of the above batch update methods.

[...]

Issue #184, opened 2016-11-11
Multi-selection


rnystrom:
Been asked this quiet a lot internally. It's easy to do but different than traditional `UICollectionView`. The whole "deselect" API is _not_ what we want to be using for this.

cdoncarroll:
I obviously don't have the proper context into why we don't want to use `UICollectionView`'s deselection API. Can you elaborate?

rnystrom:
@cdoncarroll It's not a bad API, but `UICollectionView` does a lot of (private) state preservation for you, and all the state it saves revolves around "index paths". 

I whipped up a sample project illustrating a few of these ideas: [UICollectionViewDeselection.zip](https://github.com/Instagram/IGListKit/files/586510/UICollectionViewDeselection.zip)

Where you have `-[UICollectionViewCell selected]`, we could have `-[IGListSectionController selected]` but what if only 1 of 2 cells in the section controller were actually selected? There's also `-[UICollectionView indexPathsForSelectedItems]`. Again we could do `[IGListAdapter selectedObjects]` but we have the same "1 of 2 cells in the section selected" problem.

Then there's `reloadData` which nukes selection state. I'm not sure what we do there. Maybe we internally track cell selection and restore it when cells are used? Possible. However if we do that, now we're out of sync w/ `-[UICollectionView indexPathsForSelectedItems]`, so we replace that w/ our own tracking.

Also you have to override `-[UICollectionViewCell setSelected:]` to change a cell's appearance to use custom hide/show selectors. You can use `-[UICollectionView selectedBackgroundView]` but I don't imagine that'll be super useful.

Putting this all together, we end up reimplementing the selection API and introduce some weird `UICollectionView` behavior (e.g. in the example, select a cell, tap "Reload", now try deselecting the cell. No select or deselect methods are called. Dunno why.).

In the end we do a lot of complicated things and possibly introduce some weird behavior all to save clients tracking state themselves, which shouldn't be hard.

kronik:
@rnystrom so what's the right way of managing selected cells (setting a state, getting a list of selected) for IGListCollectionView?

rnystrom:
@kronik I'll make sure to get an example up, but for now probably manually tracking selection in the section controllers. The trickier part is then getting a list of selected stuff. There are two options that I've seen work:

- At the VC level, iterate all of your section controllers. Check for your selectable SC type, then check if the SC is marked as selected (again using your custom property)
[...]

PR #449, opened 2017-01-27
WIP: Auto diffing section controller

rnystrom:
⚠️⚠️⚠️ **Do Not Merge** ⚠️⚠️⚠️

Started work on the plane to get this moving since #418  is up and ready to land. We'll likely need to spend some time fleshing out the API of this, and I think I'll split it up into a couple different PRs once ready for review. Putting this up now to get early feedback.

## What is this?

This adds an auto-diffing section controller as outlined in #38. There are several key parts:

- Subclass a new section controller `IGListAutoSectionController` (naming wip)
- Connect a data source
- Implement the data source methods that do 3 things:
  - Given a top-level object, transform it into an array of **diffable** view models
  - Given a view model, return a cell
  - Given a view model, return a size for a cell
- A new protocol for the cell `IGListBindable` so that we can control when the cell is updated w/ the view model.
  - The most important part of this is that it unlocks moving and reloading a cell, which you can't do w/ `UICollectionView`

## TODO

- [ ] Unit test `reloadObjects:`
- [x] Add `didSelectItemAtIndex:` support
- [ ] Verify all headers use FB copyright
- [ ] Create PR for `IGListAdapterUpdater` changes
  - [ ] Include fix + tests for #288 
- [ ] Create PR for new section controller w/ unit tests
- [ ] Create PR for example
- [ ] Add header docs

### IGListAdapter

I had to make some changes to `IGListAdapterUpdater` in order for this to work. The biggest challenge is when the section controller receives a new object, it has to split it up into view models and then apply changes.

However, `didUpdateToObject:` will be called _inside_ `performBatchUpdates:`, so we need to diff and fire off the changes on the collection view immediately. A few issues:

- You couldn't use `-[IGListCollectionContext performBatchAnimated:updates:completion:]` inside `didUpdateToObject:` and have it update immediately (async updates)
- There are [other ways](https://github.com/Instagram/IGListKit/blob/master/Source/IGListAdapter.m#L325) that `didUpdateToObject:` gets called, so we can't just insert/delete/move
- We're planning to remove non-batched updates entirely in #392

### Diffing top-level models

One super important piece to getting this right is having top-level (the ones given to `IGListAdapter` and handed to the SC in `didUpdateToObject:` always equal itself, even if its values change. I solved this in the unit tests w/ an object that implements diffing like:

```objc
@interface IGTestObject: NSObject <IGListDiffable>
@property NSString *key;
@end

@implementation IGTestObject
- (id)diffIdentifier {
  return self.key;
}

- (BOOL)isEqualToDiffableObject:(id)object {
  return [self.key isEqual:[object key]];
}
@end
```

If you use a traditional `isEqualToDiffableObject:` (where you check all props/values) you'll end up reloading the section, avoiding all the nice cell animations.

I started making a stock object that does this, something like `IGListAutoObject`, but its so simple that it's kind of annoying. It's essentially a model box that will probably end up more confusing than having to learn how to diff w/ the auto SC.

Would love thoughts here.

### Selection

`didSelectItemAtIndex:` is totally broken. Options are:

- Add a new method on the data source protocol
  - Gross b/c that's another empty method for a lot of folks
- Add another delegate protocol you can optionally wire up

We've discussed splitting off `didSelectItemAtIndex:` into a `selectionDelegate` (like the display/scroll/etc delegate) in #184 and elsewhere. That might help, though the view models array is private, so we'd have to expose that or create another delegate API.

## How can I help?

Since this is super early, please hold on nit picks, formatting, etc. I'll work on that as we go along. But there are some big things I'd love your thoughts on:

- The name `IGListAutoSectionController`. Anything better out there? I thought `IGListDiffableSectionController` but idk
- I'm using a data source to split the object into view models, return cells, and handle sizes. I started off passing the SC blocks on init that did the transform (sort of like `IGListSingleSectionController`). I thought the API was kind of gross though, and you have to subclass the SC anyways.
- `IGListBindable` is great, but you can't just use the view model to config the cell. For instance, how do you make the SC the cell delegate? That's why I'm leaving the method to still return a cell, but we will call bind before returning it. Is this ok?
- Sample project now or later?
- Let me know if you see some glaring Swift interop issues. I haven't tried it yet.
- Throw all the crazy test-case scenarios you got at me!


zhubofei:
@rnystrom How about `IGListReactSectionController` 😂? The one-way data flow behavior is similar to react.js

Sherlouk:
Sounds amazing!

I'd say we should definitely try and create some examples with this, an easy way to show usage and will be easier to spot most/some Swift interop issues!

rnystrom:
@zhubofei 

> How about IGListReactSectionController 🤔  The one-way data flow behavior is similar to react.js

That's so true! I worry about stepping on their toes though w/ naming. I should probably leave all the "react" stuff in their court. But I like where you're going w/ this.

@Sherlouk agreed! I'll probably include **all** of the changes in this PR so we can get a feel for things, but then split it up into 3 PRs (updater changes, new SC, example).

Btw forgot to mention that the updater changes should also fix #288.

zhubofei:
@rnystrom I prefer `IGListDiffableSectionController` than `IGListAutoSectionController`.  IMHO, `Auto` may be a little bit vague compared to what we already have i.e. `Stack`, `Single`. Or, can we use sth like `IGListTemplateSectionController`? b\c we have viewModel built into the SC.

Source/IGListAutoSectionController.m
```diff
@@ -0,0 +1,119 @@
...
+typedef NS_ENUM(NSInteger, IGListAutoSectionState) {
+    IGListAutoSectionStateIdle = 0,
+    IGListAutoSectionStateUpdateQueued,
+    IGListAutoSectionStateUpdateApplied
+};
+
+@interface IGListAutoSectionController()
+
+@property (nonatomic, strong) id object;
+@property (nonatomic, strong) NSArray<id<IGListDiffable>> *viewModels;
```
Adlai-Holler: Should we make this publicly readable? Or add a `viewModelAtIndex:` method to access this internal store? As a client it would be good to know what the state-of-the-world for this section controller is.

Source/IGListAutoSectionController.h
```diff
@@ -0,0 +1,55 @@
...
+ - IGListAutoSectionController
+ */
+
+NS_ASSUME_NONNULL_BEGIN
+
+@protocol IGListAutoSectionControllerDataSource <NSObject>
+
+- (NSArray<id<IGListDiffable>> *)viewModelsForObject:(id)object;
+- (UICollectionViewCell<IGListBindable> *)cellForViewModel:(id)viewModel atIndex:(NSInteger)index;
+- (CGSize)sizeForViewModel:(id)viewModel;
```
Adlai-Holler: Should these pass sender i.e. `- (CGSize)autoSectionController:(IGListAutoSectionController *)sectionController sizeForViewModel:(id)viewModel`? It's more formally correct but uglier.

Source/IGListAutoSectionController.m
```diff
@@ -0,0 +1,119 @@
...
+        if (self.state != IGListAutoSectionStateUpdateQueued) {
+            return;
+        }
+
+        oldViewModels = self.viewModels;
+        self.viewModels = [self.dataSource viewModelsForObject:self.object];
+        result = IGListDiff(oldViewModels, self.viewModels, IGListDiffEquality);
+
+        [collectionContext deleteInSectionController:self atIndexes:result.deletes];
+        [collectionContext insertInSectionController:self atIndexes:result.inserts];
```
Adlai-Holler: Should we support formally `reloading` the item for an updated view-model, rather than only binding a new view model to the existing cell?

One idea is, we add a data source method like:
`- (BOOL)shouldReloadCellForUpdatedViewModel:(id)viewModel atIndex:(NSInteger)index`

And call it for each updated view model to decide whether to reload or just `-bindToViewModel:`.

Source/IGListAutoSectionController.m
rnystrom: Yes! Absolutely we should. I can imagine plenty of cases where someone would need to grab the view model.

Source/IGListAutoSectionController.m
rnystrom: What would the purpose be for actual reloading? Bear in mind that we don't actually call `reloadItemsAtIndexPaths:` b/c of `UICollectionView` issues, we convert it to a delete+insert under the hood (that's the whole thing about moving alongside reload/bind here).

Source/IGListAutoSectionController.h
rnystrom: Hrm, good point. I guess to be proper they probably should.

Source/IGListAutoSectionController.m
Adlai-Holler: If the model changes in such a way that we need a new cell class to represent it, a reload (or delete/insert) is the only way to achieve the update. It's not a huge requirement, but I think it's worth considering at this early stage since it would affect the API bigly.

rnystrom: @Adlai-Holler interesting point. I could see that.

I'm leaning towards not supporting this _yet_ and filling it in later if someone really-really needs it (who knows, we might run into it). I'd love to see some concrete use-cases before we design the API too.

Plus it shouldn't be breaking when we add it, only an enhancement.

> bigly

🙊 😅

rnystrom:
Some other naming ideas:

- IGListBindingSectionController (bindable, binder)
- IGListViewModelSectionController

amonshiz:
`IGListDiffingSectionController`
- This new object is taking some actions on behalf of the developer and so the name should be active.
- The action that the object is taking is that of diffing up some view models.
- It is a section controller.

rnystrom:
History got messed up, recreated and moving to #494

PR #423, opened 2017-01-17
WIP: Updating IGListGridCollectionViewLayout to be more flexible

zhubofei:
## Changes in this pull request
Updating the grid layout to fit more general cases. 

Closes #263

## TODO 
- [x] Add support for multi-item sections.
- [x] Add unit test for multi-item sections.
- [x] Add `IGListGridCollectionViewLayoutAlignment` option. Allow user to set alignment for the layout.
- [x] Add unit test for alignments.
- [ ] Add support for horizontal scrolling.
- [ ] Add unit test for horizontal mode.
- [ ] Add support for section inset.
- [ ] Add unit test for section inset.
- [ ] Add attributes cache.
- [ ] Update Changelog

PR #450, opened 2017-01-27
Vertical UICollectionViewLayout supporting inline sections

rnystrom:
Working on porting our collection view layout to IGListKit. I'm doing this because its a solid layout, and we just finished preparing it to work with inline sections. It is designed to work in tandem with IGListKit, so we're adding it.

This is still a WIP as I add more tests, but I'd love as much feedback as possible.

Aside from the glob of header documentation, this has the following features:

- Infinite sections that each have infinite items. Sections and items can fall inline. When they break the width of their container they will fall on the next row.
- Sections can have their own insets, line spacing, and interitem spacing.
- Sticky header support! When you use headers, it will always newline the section.
- Maximum width with a border decoration view
  - Use this to pinch in your content on larger devices

Followup to #423 

## TODO

- [ ] ~~Move decoration view support to delegate~~ removed
- [x] Unit test changing [top y sticky inset](https://coveralls.io/builds/9977284/source?filename=Source%2FIGListCollectionViewLayout.mm#L362)
- [ ] ~~Update `initWithCoder:` to use sensible defaults~~
- [x] Test `initWithCoder:` throws
- [x] Test when remaining cells [outside of visible bounds](https://coveralls.io/builds/9977284/source?filename=Source%2FIGListCollectionViewLayout.mm#L180)
- [ ] ~~Move `decorationBorderViewOfKind:` internal? C function? (or add unit tests)~~ removed
- [x] Unit test changing [bounds size](https://coveralls.io/builds/9977284/source?filename=Source%2FIGListCollectionViewLayout.mm#L337)

### DIFF
--- a/Source/IGListCollectionViewLayout.h
+++ b/Source/IGListCollectionViewLayout.h
@@ -1,0 +1,107 @@
+/**
+ * Copyright (c) 2016-present, Facebook, Inc.
+ * All rights reserved.
+ *
+ * This source code is licensed under the BSD-style license found in the
+ * LICENSE file in the root directory of this source tree. An additional grant
+ * of patent rights can be found in the PATENTS file in the same directory.
+ */
+
+#import <UIKit/UIKit.h>
+
+#import <IGListKit/IGListMacros.h>
+
+NS_ASSUME_NONNULL_BEGIN
+
+/**
+ This UICollectionViewLayout subclass is for vertically-scrolling lists of data with variable widths and heights. It
+ supports an infinite number of sections and items. All work is done on the main thread, and while extremely efficient,
+ care must be taken not to stall the main thread in sizing delegate methods.
+
+ This layout piggybacks on the mechanics of UICollectionViewFlowLayout in that:
+
+ - Your UICollectionView data source must also conform to UICollectionViewDelegateFlowLayout
+ - Header support given via UICollectionElementKindSectionHeader
+
+ All UICollectionViewDelegateFlowLayout methods are required and used by this layout:
+
+ - (CGSize)collectionView:(UICollectionView *)collectionView layout:(UICollectionViewLayout *)collectionViewLayout sizeForItemAtIndexPath:(NSIndexPath *)indexPath;
+ - (UIEdgeInsets)collectionView:(UICollectionView *)collectionView layout:(UICollectionViewLayout *)collectionViewLayout insetForSectionAtIndex:(NSInteger)section;
+ - (CGFloat)collectionView:(UICollectionView *)collectionView layout:(UICollectionViewLayout *)collectionViewLayout minimumLineSpacingForSectionAtIndex:(NSInteger)section;
+ - (CGFloat)collectionView:(UICollectionView *)collectionView layout:(UICollectionViewLayout *)collectionViewLayout minimumInteritemSpacingForSectionAtIndex:(NSInteger)section;
+ - (CGSize)collectionView:(UICollectionView *)collectionView layout:(UICollectionViewLayout *)collectionViewLayout referenceSizeForHeaderInSection:(NSInteger)section;
+
+ Sections and items are put into the same horizontal row until the max-x position of an item extends beyond the width
+ of the collection view. When that happens, the item is "newlined" to the next row. The y position of that row is
+ deteremined by the maximum height (including section insets) of the section/item of the previous row.
+
+ Ex. of a section (2,0) with a large width causing a newline.
+ |[ 0,0 ][ 1,0 ]         |
+ |[         2,0         ]|
+
+ A section with a non-zero height header will always cause that section to newline. Headers are always stretched to the
+ width of the collection view, pinched with the section insets.
+
+ Ex. of a section (2,0) with a header inset on the left/right.
+ |[ 0,0 ][ 1,0 ]         |
+ | >======header=======< |
+ | [ 2,0 ]               |
+
+ Section insets apply to items in the section no matter if they begin on a new row or are on the same row as a previous
+ section.
+
+ Ex. of a section (2) with multiple items and a left inset.
+ |[ 0,0 ][ 1,0 ] >[ 2,0 ]|
+ | >[ 2,1 ][ 2,2 ][ 2,3 ]|
+
+ Interitem spacing applies to items and sections within the same row. Line spacing only applies to items within the same
+ section.
+
+ Please see the unit tests for more configuration examples and expected output.
+ */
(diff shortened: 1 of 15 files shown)


jessesquires:
naming nit: how about just `IGListCollectionViewLayout` ?

rnystrom:
@jessesquires I dig it. I left "vertical" b/c atm it only supports v-scrolling, but we could maybe add horizontal down the road. Or not. I like that better.

rnystrom:
@jeremycohen how do you feel about me landing this and setting up our internal tests? This is copy-pasted and cleaned up from D4455662.

Source/IGListCollectionViewLayout.h
```diff
@@ -0,0 +1,94 @@
...
+ Ex. of a section (2) with multiple items and a left inset.
+ |[ 0,0 ][ 1,0 ] >[ 2,0 ]|
+ | >[ 2,1 ][ 2,2 ][ 2,3 ]|
+
+ Interitem spacing applies to items and sections within the same row. Line spacing only applies to items within the same
+ section.
+
+ Please see the unit tests for more configuration examples and expected output.
+ */
+IGLK_SUBCLASSING_RESTRICTED
```
zhubofei: Can we allow subclass here? Subclassing `UICollectionViewFlowLayout` proved to be useful.

rnystrom: Hmm, maybe we could. It's not really designed to be subclassed, but I guess we could let people do that if they wanted. @jessesquires any thoughts?

Source/IGListCollectionViewLayout.h
```diff
@@ -0,0 +1,94 @@
...
+
+ All UICollectionViewDelegateFlowLayout methods are required and used by this layout:
+
+ - (CGSize)collectionView:(UICollectionView *)collectionView layout:(UICollectionViewLayout *)collectionViewLayout sizeForItemAtIndexPath:(NSIndexPath *)indexPath;
+ - (UIEdgeInsets)collectionView:(UICollectionView *)collectionView layout:(UICollectionViewLayout *)collectionViewLayout insetForSectionAtIndex:(NSInteger)section;
+ - (CGFloat)collectionView:(UICollectionView *)collectionView layout:(UICollectionViewLayout *)collectionViewLayout minimumLineSpacingForSectionAtIndex:(NSInteger)section;
+ - (CGFloat)collectionView:(UICollectionView *)collectionView layout:(UICollectionViewLayout *)collectionViewLayout minimumInteritemSpacingForSectionAtIndex:(NSInteger)section;
+ - (CGSize)collectionView:(UICollectionView *)collectionView layout:(UICollectionViewLayout *)collectionViewLayout referenceSizeForHeaderInSection:(NSInteger)section;
+
+ Sections and items are put into the same horizontal row until the max-y position of an item extends beyond the width
```
zhubofei: > until the max-y position

Should be max-x here?

Source/IGListCollectionViewLayout.mm
```diff
@@ -0,0 +1,512 @@
...
+
+    const NSInteger sectionCount = _sectionData.size();
+
+    if (sectionCount == 0) {
+        return CGSizeZero;
+    }
+
+    const IGSectionEntry section = _sectionData[sectionCount - 1];
+    const CGFloat height = CGRectGetMaxY(section.bounds) + section.insets.bottom;
+    return CGSizeMake(MIN(CGRectGetWidth(self.collectionView.bounds), self.maximumContentWidth), height);
```
zhubofei: Not sure if we should take `collectionView.contentInset` into account here.

zhubofei: It is my understanding that `maximumContentWidth` acts sort of like `collectionView.contentInset`here so that we can have paddings on left and right. So, we do not take `contentInset` into account in this layout at all?

Source/IGListCollectionViewLayout.h
rnystrom: Ooops ya

Source/IGListCollectionViewLayout.mm
rnystrom: I think that `collectionView.contentInset` is used by the actual scroll view and isn't used in the layout calculations. For instance in Instagram we use a variation of this layout and set the content inset to sit underneath the nav bar and it works. But that's a good point I haven't really considered that before. I'm not sure if the layout is supposed to consider that... 🤔

zhubofei: @rnystrom I found this [tutorial](https://www.raywenderlich.com/107439/uicollectionview-custom-layout-tutorial-pinterest) on Raywenderlich.
screen shot 2017-02-04 at 11 10 35 pm
They took `contentInset` into account in this sample code. Idk if it is the common practice or not.

weyert:
Any chance for an example which this new layout will bring us?

stevejcox:
This is exactly what I need, arriving at just the right time!

rnystrom:
@weyert yup! We'll add one shortly. Still ironing out some internal kinks as we land this.

rnystrom:
Had a hiccup where we had to unland this, adding back ASAP

weyert:
Thank you @rnystrom Does it mean you will support horizontal scrolling in inline controllers?

rnystrom:
@weyert you can use embedded lists or scroll views to achieve an h-scroll, inline effect. If you want to use h-scrolling section controllers, check out [our example](https://github.com/Instagram/IGListKit/blob/master/Examples/Examples-iOS/IGListKitExamples/SectionControllers/HorizontalSectionController.swift).

weyert:
@rnystrom Aha, so you need to make a collection view cell which in itself contains a UICollectionView instance again? I have to admit that I am really stupid and lost the clue by now. What's the added advantage of inline sections? Somehow I thought it was to add things like horizontal scrolling items.

rnystrom:
@weyert oh sorry, I should have explained better. Inline sections are nice b/c `UICollectionViewFlowLayout` will "newline" each section. So if you have sections, each with a single cell that is 1/3 the width of the screen (think [Instagram grids](https://cdn.macrumors.com/article-new/2016/05/Instagram-New-Design-duo.jpg)), flow layout will put each section-cell at x position `0` with a bunch of white space filling the remaining 2/3 of the collection view. It'll look like a single column of cells on the left. 

Hope that explains it!

otymartin:
@rnystrom Just curious if the sticky headers is now supported so I can drop my collectionView in a tableviewcell structure?

Issue #315, opened 2016-12-10
Container size and section insets


rnystrom:
It can be a little tricky dealing with `insets` and `collectionContext.containerSize`. The container width doesn't account for insets, so if you have the following setup:

- `insets` of 10, 10, 10, 10
- The width of the collection view is `320`

If your size method looks like

```objc
- (CGSize)sizeForItemAtIndex:(NSInteger)index {
  return CGSizeMake(self.collectionContext.containerSize.width, 50);
}
```

`UICollectionViewFlowLayout` will error b/c the item width is `320` but layout adds left/right insets, making the total item width `340`.

Maybe we should have a `containerInsetSizeForSectionController:` method that returns the container size minus `insets`?

jessesquires:
what about `collectionContext.insetContainerSize` ?

rnystrom:
@jessesquires only thing we have to add is the section controller b/c each section controller is responsible for their own `inset` property

racer1988:
Can we verify this is correctly calculated on iOS 9?
Inset were applied incorrectly for me.
I noticed a issue some time ago but I don't have a phone with iOS 9 to reproduce at the moment.

jessesquires:
@racer1988 -- this work hasn't been completed. do you want to submit a PR?

racer1988:
@jessesquires If it is not done after the Christmas Holiday, I may try. 
I am not a good OBJ-C programmer, I am Swift "onlyish"
but I will have a look if not done in some weeks.

Thanks! :D

zhubofei:
@racer1988 Are you still interested in working on this one?

racer1988:
@zhubofei Hi! I did not start, and I might not have the time for a while. 
So if you want to take it, go on

Issue #183, opened 2016-11-10
Refactor unit test helper macros and functions into single file


rnystrom:
Lots of stuff that we use in multiple places that I think should be cleaned up throughout the test suite

- https://github.com/Instagram/IGListKit/blob/master/Tests/IGListAdapterE2ETests.m#L22-L25
- https://github.com/Instagram/IGListKit/blob/master/Tests/IGListBatchUpdateDataTests.m#L25-L39
- https://github.com/Instagram/IGListKit/blob/master/Tests/IGListDiffTests.m#L25-L28

We also have a similar pattern throughout our tests:

- Create window, adapter, collection view
- Setup data source before testing
  - https://github.com/Instagram/IGListKit/blob/master/Tests/IGListAdapterE2ETests.m#L64-L69

Do we create our own testing suite class? Or maybe just a bunch of setup helper functions? Or not worry about it?

jessesquires:
> Do we create our own testing suite class?

:100: Yes, we should. Lowers barrier for writing tests. Less boilerplate == more likely to write tests. :smile:

PR #456, opened 2017-02-01
Container Inset Size For Section Controller

zhubofei:
Add API for #315. Not sure if this is what you want @rnystrom though 🤔. Will add tests after you confirm.

## Pull request checklist


rnystrom:
This is perfect! Some ideas about naming:

- `maximumContainerSize...`
- `insetContainerSize...`

Or maybe its just `containerSizeForSectionController:`? Just cleaning up the documentation on both this and `containerSize` methods to be clear that this one accounts for insets on the section.

Thoughts?

zhubofei:
@rnystrom `maximumContainerSizeForSC` is definitely better. Although, I was thinking maybe it would make more sense if we separate `containerInsetSize` into `maxWidth` & `maxHeight` for SC (need 2 separate API though, kind of gross..), b/c the "insetSize" itself doesn't carry any meaning as a size (it is rather a maxWidth or maxHeight for each SC depending on scroll direction?).

rnystrom:
@zhubofei 🤔 hmm really great point. I might stay in favor of returning a single `CGSize` since *technically* you could have a 2d scrolling collection view (w/ a custom `UICollectionViewLayout`).

zhubofei:
@rnystrom We only need this for layouts like `UICollectionViewFlowLayout` with a fixed length in one direction, right? But return single `CGSize` is much cleaner anyway. We just need to clarify what the size actually is in documentation. However, if we use "maximumContainerSize...", it is kind of misleading, b/c SC can be larger in at least one direction. So maybe just `containerSizeForSectionController:` 😂.

rnystrom:
`containerSizeForSectionController:` ++

zhubofei:
@rnystrom Since `(inset.left + inset.right)` could be large than `containerSize.width`, I wonder if we need a Assert here. Or just return some  negtive sizes..

rnystrom:
@zhubofei Hmm, I wonder if we should just handle it gracefully and return 0 if that's the case? I wonder if there are some weird scenarios where the collection view has width 0 before its been laid out.

rnystrom:
@zhubofei just checking, still up for this?

zhubofei:
@rnystrom 😂 I totally forget about this one! Will finish it tmr.

jessesquires:
Is this PR good to go?

Need to add a changelog entry

zhubofei:
@jessesquires Not really, still need some unit tests. I've been pretty busy this week 😭. Will finish this one tonight or tmr morning. Sorry about the delay.

jessesquires:
@zhubofei - no rush at all! no worries! 👍

jessesquires:
✨  friendly ping ✨ 

@rnystrom -- we can carry this forward and add tests internally if needed. you have spare cycles?

Tests/IGListAdapterTests.m
```diff
@@ -27,6 +27,12 @@
 XCTAssertEqual(CGPointEqualToPoint(point, p), YES); \
 } while(0)
 
+#define IGAssertEqualSize(size, w, h, ...) \
+do { \
+CGSize s = CGSizeMake(w, h); \
+XCTAssertEqual(CGSizeEqualToSize(size, s), YES); \
+} while(0)
+
```
jessesquires: 🤔  hm... i think i'd rather remove this (especially since it's copied multiple places)

instead, `XCTAssertTrue(CGSizeEqualToSize(....));`

thoughts? @rnystrom ?

zhubofei: @jessesquires Are we still gonna do  #183?

jessesquires: ah 👍 fair enough. let's import this now then.

PR #474, opened 2017-02-07
Re-gen docs for 2.2 release

jessesquires:
Ok so:

This is branched off of **stable** to re-gen 2.x docs (otherwise, we'd be generating unfinished 3.0 docs).

However, this brings with it the cherry-picked commits. 😢 

These should squash into a "no-op" -- I think? 

Maybe we should push doc updates to `stable` directly then PR `stable` into `master`? 

Thoughts?


Sherlouk:
Agree w/ `update-docs` -> `stable`, then `stable` -> `master`

jessesquires:
ok. going to bail on this until we get branches fixed up

PR #477, opened 2017-02-08
Add -[IGListAdapter sectionControllerForSection:]

Adlai-Holler:
Before the diff, you can lookup the object for a given section, and then lookup the section controller for that object, but this seems like a pretty valuable/common operation.


Adlai-Holler:
I put the changelog in the 2.2.0 release, I hope you guys will drop it in.

PR #488, opened 2017-02-13
Handle nil section controllers from data source

rnystrom:

Removing the assert and handling `nil` section controllers. This wont effect Swift code which demands a non-optional section controller, but Objective-C is more nuanced. There is evidence that some callers do not have a default state and are returning `nil`.

Unfortunately this will [OOB here](https://github.com/Instagram/IGListKit/blob/master/Source/Internal/IGListSectionMap.m#L63) if using the original `objects` array provided by the data source. Instead we prune the array, only allowing in objects that have a matching section controller.

Fixes t15773862 internally.

## Pull request checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,6 +28,10 @@ This release closes the [3.0.0 milestone](https://github.com/Instagram/IGListKit
 
 - You can now manually move items (cells) within a section controller, ex: `[self.collectionContext moveInSectionController:self fromIndex:0 toIndex:1]`. [Ryan Nystrom](https://github.com/rnystrom) [(#418)](https://github.com/Instagram/IGListKit/pull/418)
 
+### Fixes
+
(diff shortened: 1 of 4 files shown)


jessesquires:
@rnystrom -- But allowing `nil` here contradicts the API contract, which specifies that implementers must return non-nil section controllers.

Shouldn't we just handle the OOB error instead?
https://github.com/Instagram/IGListKit/blob/master/Source/Internal/IGListSectionMap.m#L63

rnystrom:
@jessesquires and I just chatted offline:

- This diff fixes the OOB by just skipping the object+section controller if the SC is nil
- Keeping the assert makes testing difficult. We could add a check `NSStringFromClass(@"XCTestCase") == nil` before asserting, but that's kind of gross

I'm going to add a warning log if the SC is nil so we can cleanly test and at least raise some complaints.

Issue #38, opened 2016-10-02
RFC: View-model, auto diffing section controller


rnystrom:
We generate view models to drive all of the cells in a particular section controller on Instagram. These models are derived from a top-level model that is created from server data. The view models are then used to configure things like image and text cells within a section controller. 

We can represent this transformation:

``` swift
protocol ViewModel: IGListDiffable {...}

func map(object: Any) -> [ViewModel] {...}
```

We could create a section controller that diffs its own array of `ViewModel` objects, then apply those diffs at the _item_ (e.g. "cell") level.

There are a few gotchas to doing this, that's why I think a built-in and tested version of this would be valuable (similar to the embedded adapter idea in #31).

You end up with diffing of sections/section-controllers _and_ the items/cells within them. It totally converts your setup to being data-driven all the way down to the cell level.

rnystrom:
This would really be _one-way data flow_ once built:

1. Top-level data in `IGListAdapterDataSource`
2. Section controller transforms data into an array of view models
3. View-model to cell class
4. "Bind" view model to cell (e.g. config)

Don't want to lock down the API too much though, for instance cell selection should still be handled manually I think.

jessesquires:
> Section controller transforms data into an array of view models

This is "too late" in the flow.

Where we use view models, we're returning them in `-itemsFromListAdapter:`. The view models need to be **the items** that are diff'd.

rnystrom:
@jessesquires I mean use a top-level array of objects to drive sections, then a view-model array to drive `numberOfItems`. Then the section controller does diffing for its view models when the top-level model changes.

Not sure if that makes sense.

(basically what we do for feed items tho)

rnystrom:
Had some ideas this morning. Need to write them down so they aren't lost.

- Instead of calling reload in the infra, if we have a cell bind/configure architecture, we can actually reconfigure the cell instead of calling reload (which does delete+insert internally). This unlocks moving and reloading the cell at the same time!!
- Have to take special care not to call reloadSection when diffing the top-level object. If that happens the item-level updates are lost.
- Need to avoid over-diffing the view models. Just like the top-level arrays, this should only happen once when the update is performed.
- Should the transform be a SC subclass override? a block passed on init? custom object w/ protocol passed on init? Still 🤔 on this
- Does #392 need to be done first? I don't think so.
- I like the sound of `IGListBindable` for the cells. Thoughts?

jessesquires:
💯

[...]

PR #494, opened 2017-02-15
Auto diffing section controller

rnystrom:
Started work on the plane to get this moving since #418  is up and ready to land. We'll likely need to spend some time fleshing out the API of this, and I think I'll split it up into a couple different PRs once ready for review. Putting this up now to get early feedback.

## What is this?

This adds an auto-diffing section controller as outlined in #38. There are several key parts:

- Subclass a new section controller `IGListAutoSectionController` (naming wip)
- Connect a data source
- Implement the data source methods that do 3 things:
  - Given a top-level object, transform it into an array of **diffable** view models
  - Given a view model, return a cell
  - Given a view model, return a size for a cell
- A new protocol for the cell `IGListBindable` so that we can control when the cell is updated w/ the view model.
  - The most important part of this is that it unlocks moving and reloading a cell, which you can't do w/ `UICollectionView`

## TODO

- [ ] Unit test `reloadObjects:`
- [x] Add `didSelectItemAtIndex:` support
- [x] Verify all headers use FB copyright
- [x] Create PR for `IGListAdapterUpdater` changes
  - [x] Include fix + tests for #288 
- [ ] ~~Create PR for re-entrant updates~~
  - ~~e.g. calling `performBatchAnimated:updates:completion:` within an in-progress update block should execute immediately~~ doing this here
- [x] Create PR for new section controller w/ unit tests
- [x] Create example
- [x] Add header docs
- [ ] Optimize test coverage

### IGListAdapter

I had to make some changes to `IGListAdapterUpdater` in order for this to work. The biggest challenge is when the section controller receives a new object, it has to split it up into view models and then apply changes.

However, `didUpdateToObject:` will be called _inside_ `performBatchUpdates:`, so we need to diff and fire off the changes on the collection view immediately. A few issues:

- You couldn't use `-[IGListCollectionContext performBatchAnimated:updates:completion:]` inside `didUpdateToObject:` and have it update immediately (async updates)
- There are [other ways](https://github.com/Instagram/IGListKit/blob/master/Source/IGListAdapter.m#L325) that `didUpdateToObject:` gets called, so we can't just insert/delete/move
- We're planning to remove non-batched updates entirely in #392

### Diffing top-level models

One super important piece to getting this right is having top-level (the ones given to `IGListAdapter` and handed to the SC in `didUpdateToObject:` always equal itself, even if its values change. I solved this in the unit tests w/ an object that implements diffing like:

```objc
@interface IGTestObject: NSObject <IGListDiffable>
@property NSString *key;
@end

@implementation IGTestObject
- (id)diffIdentifier {
  return self.key;
}

- (BOOL)isEqualToDiffableObject:(id)object {
  return [self.key isEqual:[object key]];
}
@end
```

If you use a traditional `isEqualToDiffableObject:` (where you check all props/values) you'll end up reloading the section, avoiding all the nice cell animations.

I started making a stock object that does this, something like `IGListAutoObject`, but its so simple that it's kind of annoying. It's essentially a model box that will probably end up more confusing than having to learn how to diff w/ the auto SC.

Would love thoughts here.

### Selection

`didSelectItemAtIndex:` is totally broken. Options are:

- Add a new method on the data source protocol
  - Gross b/c that's another empty method for a lot of folks
- Add another delegate protocol you can optionally wire up

We've discussed splitting off `didSelectItemAtIndex:` into a `selectionDelegate` (like the display/scroll/etc delegate) in #184 and elsewhere. That might help, though the view models array is private, so we'd have to expose that or create another delegate API.

## How can I help?

Since this is super early, please hold on nit picks, formatting, etc. I'll work on that as we go along. But there are some big things I'd love your thoughts on:

- The name `IGListAutoSectionController`. Anything better out there? I thought `IGListDiffableSectionController` but idk
- I'm using a data source to split the object into view models, return cells, and handle sizes. I started off passing the SC blocks on init that did the transform (sort of like `IGListSingleSectionController`). I thought the API was kind of gross though, and you have to subclass the SC anyways.
- `IGListBindable` is great, but you can't just use the view model to config the cell. For instance, how do you make the SC the cell delegate? That's why I'm leaving the method to still return a cell, but we will call bind before returning it. Is this ok?
- Sample project now or later?
- Let me know if you see some glaring Swift interop issues. I haven't tried it yet.
- Throw all the crazy test-case scenarios you got at me!


heumn:
> The name IGListAutoSectionController. Anything better out there? I thought IGListDiffableSectionController but idk

I think it is a good name. It describes what it does pretty well without being too verbose 👌

jessesquires:
How about `IGListAutoDiffSectionController` ?

heumn:
Yeah, `IGListAutoDiffSectionController` is actually a better name

rnystrom:
@jessesquires @heumn we had some convo in #449 with other ideas:

- IGListAutoSectionController
- IGListReactSectionController
- IGListDiffableSectionController @zhubofei 
- IGListDiffingSectionController @amonshiz
- IGList(Cell)BindingSectionController
- IGListViewModelSectionController

I like IGListAutoDiffSectionController too!

rnystrom:
Will rebase off of #525 to lighten the load here.

rnystrom:
Anyone got any ideas on what we could demo as an example? Something that isn't crazy complicated, but would show off cells updating automatically.

heumn:
Maybe something like you do in the nested controller example, but use colours and circles to easier track the diffing inside the section? 

Can not wait to play with this :)

skjermbilde 2017-03-05 kl 13 03 17

jessesquires:
@rnystrom - can we defer this to `3.1.0` or does this contain breaking changes?

rnystrom:
@jessesquires no breaking changes, we could kick this from 3.0 but I think I'll have this working + example this weekend.

rnystrom:
Ok, I'm getting super excited about this. Check out the demo (gif below).

I'm about ready to move this from WIP to review. Should be ready tomorrow.

diffing

heumn:
Initial reaction to this is that you've just saved me hundreds of lines of code and a hack 😂  Looks great

I also postponed "Call mom" until the 25th, and rather "Started cycling" the 20th. No flashing reloading of the sections and things updated just as I wanted 👌 💯 

Maybe it could be an idea to add something like this to extend on how powerful this is?

call-mom


I also keep thinking how you could make it extra clear that the section models _has_ to follow this principle? No ideas apart from the obvious documentation that is needed :)


>One super important piece to getting this right is having top-level (the ones given to IGListAdapter and handed to the SC in didUpdateToObject: always equal itself, even if its values change.

heumn:
Btw, I forgot to that big fucking thank you for the awesome work you are putting in @rnystrom 🙌

rnystrom:
@heumn great idea! I'm going to add an assert inside `-[IGListDiffingSectionController didUpdateToObject:]` that checks `[object isEqual:self.object]` (e.g. two objects should always be equal otherwise a section-reload happens). Also going to load the headers w/ docs and examples.

And nice extension! Maybe we can add that as a follow up?

heumn:
> @heumn great idea! I'm going to add an assert inside -[IGListDiffingSectionController didUpdateToObject:] that checks [object isEqual:self.object] (e.g. two objects should always be equal otherwise a section-reload happens). Also going to load the headers w/ docs and examples.

Perfect 🎉 

> And nice extension! Maybe we can add that as a follow up?

👌

rnystrom:
Importing for internal review!

jessesquires:
that calendar view though 😚 👌

amonshiz:
Early morning review before a run is probably the best kind of review right?

rnystrom:
@amonshiz forget to hit "Submit Review"? 🤔

Source/IGListDiffingSectionController.m
```diff
@@ -0,0 +1,117 @@
...
+@property (nonatomic, strong) id object;
+@property (nonatomic, assign) IGListDiffingSectionState state;
+
+@end
+
+@implementation IGListDiffingSectionController
+
+#pragma mark - Public API
+
+- (void)updateAnimated:(BOOL)animated completion:(void (^)())completion {
```
rnystrom: Good eye! Means I need to add a test for this too.

rnystrom:
Waddup @iglistkit-bot catching my own mistakes. I'm in love w/ this tool.

rnystrom:
I'm still going back and forth on naming. Now I'm starting to lean towards `IGListBindingSectionController` b/c:

- The section controller does the "binding"
- "Diffing" is an implementation detail

The emphasis of this API should be on the model -> view model -> binding parts, not that it diffs.

What do you all think? @jessesquires @heumn @PhilCai1993 @amonshiz

amonshiz:
sure? i'm bad at naming.

jessesquires:
i'll defer to @amonshiz for naming strategies.

andrewnamesthings

heumn:
> I'm still going back and forth on naming. Now I'm starting to lean towards IGListBindingSectionController b/c:

> The section controller does the "binding"
> "Diffing" is an implementation detail
> The emphasis of this API should be on the model -> view model -> binding parts, not that it diffs.

> What do you all think? @jessesquires @heumn @PhilCai1993 @amonshiz

Good reasoning 👌  👍

PhilCai1993:
It's excited! I wish I could download this commits but how? 😅

heumn:
It was merged into master, so you can just point you podfile to the master branch :)

jessesquires:
@PhilCai1993 check out installation guide

https://instagram.github.io/IGListKit/installation.html

PhilCai1993:
May I have a question here? If in a collectionview with various "Month", since `isEqualToDiffableObject:` simply returns true, how to reload the entire collectionview?

rnystrom:
@PhilCai1993 covering that in #621 I think

Issue #383, opened 2017-01-04
RFC: Preprocessing API to do expensive work on background before applying batch updates


rnystrom:
This is a proposal to add an API that makes it easy to do expensive things (e.g. size text, file i/o) on a background queue from within a section controller before an update is applied.

## Current situation

If you have expensive work to do before using your section controller, currently your only options are:

- Create a **very specific** method/helper to warm up caches or w/e before calling `[adapter performUpdatesAnimated:YES completion:nil]` 😢 
- Do it on main 😭 

We use both of these solutions and still hit 60fps, but it is _not easy_.

## API Design

- Create a new protocol `IGListPreprocessingDelegate`
- Add a new weak property on `IGListSectionController` similar to the [scroll delegate](https://github.com/Instagram/IGListKit/blob/master/Source/IGListSectionController.h#L105) called `preprocessingDelegate`
  - Let's section controllers assign `self`
  - `nil` by default

### IGListPreprocessingDelegate

The `IGListPreprocessingDelegate` will have 2 **required** methods:

```objc
// Called on a background queue. This is where you would do work, sync or async.
- (void)preprocessWithContext:(IGListPreprocessingContext *)context;

// Method called on main once all work is done.
- (void)preprocessingDidFinishWithContext:(IGListPreprocessingContext *)context;
```

### IGListPreprocessingContext

This is object is similar to the [UIViewControllerContextTransitioning](https://developer.apple.com/reference/uikit/uiviewcontrollercontexttransitioning) (should we name it _IGListContextPreprocessing_?). It's main duties:

- Give the `preprocessingDelegate` info
  - The `object` of the section controller
  - Container size
  - The section index?
- Have an `id value` object to store arbitrary data
  - e.g. calculate the `CGSize` of some text, wrap it in `NSValue`, and do `context.value = wrappedSize;`
- Require calling `[context completePreprocessing]` to mark finished
  - Allows the method to call async, block-based APIs

### Main thread affinity

Every public `IGListKit` API has `IGAssertMainQueue()` in it and I **don't** think we should remove any of that so we don't get into threading chaos. That means that when in `-preprocessWithContext:` you can't do something like `[self.collectionContext containerSize]`. We should strive to put all necessary background data on each `IGListPreprocessingContext` object.

### Concurrent work

Ideally, preprocessing should all be done concurrently, though we have to wait (non-blocking) until all of the work is finished before continuing. I think we could:

[...]

PR #501, opened 2017-02-21
[WIP] Preprocessing Phase for Updates

Adlai-Holler:
Issue fixed: #383 

- [ ] Add Unit Tests
- [ ] Make IGListUpdatingDelegate's new API support blocking on the preprocess phase
- [ ] Fix indentation!
- [ ] Figure out how to pipe the set of section controllers into the preprocessing system

### DIFF
--- a/Source/IGListAdapterUpdater.m
+++ b/Source/IGListAdapterUpdater.m
@@ -16,6 +16,7 @@
 
 #import "UICollectionView+IGListBatchUpdateData.h"
 #import "IGListMoveIndexPathInternal.h"
+#import "IGListPreprocessor.h"
 
 @implementation IGListAdapterUpdater
 
@@ -124,6 +125,7 @@ static NSArray *objectsWithDuplicateIdentifiersRemoved(NSArray<id<IGListDiffable
     NSArray *fromObjects = [self.fromObjects copy];
     NSArray *toObjects = objectsWithDuplicateIdentifiersRemoved(self.toObjects);
     void (^objectTransitionBlock)(NSArray *) = [self.objectTransitionBlock copy];
+    IGListUpdatePreprocessingBlock preUpdateBlock = self.preUpdateBlock;
     NSArray *itemUpdateBlocks = [self.itemUpdateBlocks copy];
     NSArray *completionBlocks = [self.completionBlocks copy];
     const BOOL animated = self.queuedUpdateIsAnimated;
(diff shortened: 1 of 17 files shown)


Source/IGListAdapterUpdater.m
```diff
@@ -124,107 +125,111 @@ - (void)performBatchUpdatesWithCollectionView:(UICollectionView *)collectionView
...
-                completionBlock(finished);
-                [CATransaction commit];
-            }];
-        }
-    } @catch (NSException *exception) {
-        [delegate listAdapterUpdater:self willCrashWithException:exception fromObjects:fromObjects toObjects:toObjects updates:updateData];
-        @throw exception;
-    }
+	
+	// We are beginning to move into callback hell.
```
rnystrom: 😅 

Going to brainstorm ways to avoid this

Source/Internal/IGListPreprocessor.m
```diff
@@ -0,0 +1,98 @@
...
+		_group = dispatch_group_create();
+	}
+	return self;
+}
+
+- (void)schedulePreprocessingWithContainerSize:(CGSize)containerSize completion:(nonnull dispatch_block_t)completionBlock {
+	IGAssertMainThread();
+	_completion = completionBlock;
+	
+	dispatch_queue_t queue = dispatch_queue_create("IGListPreprocessor Queue", DISPATCH_QUEUE_CONCURRENT);
```
rnystrom: nit: queue name should probably follow Apple patterns and be something like

`com.iglistkit.iglistpreprocessor`

👌  concurrent

Source/Internal/IGListPreprocessor.m
```diff
@@ -0,0 +1,98 @@
...
+	}
+	
+	// Call and release our completion block.
+	if (_completion) {
+		_completion();
+		_completion = nil;
+	}
+}
+
+@end
```
rnystrom: This is how you write beautiful async code. Note that there are no gross locks, atomic stuff, or custom containers. Just perfectly designed GCD. Kudos @Adlai-Holler ❤️

Source/Internal/IGListPreprocessor.m
```diff
@@ -0,0 +1,98 @@
...
+		_contextToDelegateMap = [NSMapTable strongToStrongObjectsMapTable];
+		_sectionMap = [sectionMap copy];
+		_group = dispatch_group_create();
+	}
+	return self;
+}
+
+- (void)schedulePreprocessingWithContainerSize:(CGSize)containerSize completion:(nonnull dispatch_block_t)completionBlock {
+	IGAssertMainThread();
+	_completion = completionBlock;
```
ocrickard: I haven't fully understood the scope of this preprocessor, but _completion is being overwritten here right? What happens if a second schedulePreprocessingWithContainerSize is called a second time? If so, then _completion from the first execution is overwritten, and is never executed, which could leave callers in an incomplete state permanently, right? Is calling multiple times allowed? If not, then perhaps an assertion?

Source/Internal/IGListPreprocessor.m
```diff
@@ -0,0 +1,98 @@
...
+		});
+	}];
+	
+	// When the work is all done, call -runCompletionIfNeeded on main.
+	dispatch_group_notify(_group, dispatch_get_main_queue(), ^{
+		[self runCompletionIfNeeded];
+	});
+}
+
+- (void)waitForPreprocessingToFinish {
```
ocrickard: This seems a little dangerous, right? Allowing someone on the main thread to block until a bg queue completes can result in deadlocks/priority inversion.

rnystrom: Good call, we definitely **do not** want to block main at all while preprocessing happens!

Source/Internal/IGListPreprocessor.m
rnystrom: ++ assertion. Shouldn't ever be running 2 preprocessing jobs at a time:

- Preprocess, dispatching work to concurrent queues, wait til done
- Then `performBatchUpdates:`
- Rely on same IGLK infra to prevent multiple batch updates happening at the same time

Source/Internal/IGListPreprocessor.h
```diff
@@ -0,0 +1,34 @@
...
+
+IGLK_SUBCLASSING_RESTRICTED
+@interface IGListPreprocessor : NSObject
+
+- (instancetype)initWithSectionMap:(IGListSectionMap *)sectionMap;
+
+/**
+ * Must be called from the main thread.
+ */
+- (void)schedulePreprocessingWithContainerSize:(CGSize)containerSize completion:(dispatch_block_t)completionBlock;
```
rnystrom: @amonshiz this is the only way to do it w/out making that a locking property. IMO this is much cleaner, but probably something we need to consider supporting down the line. This is following my outline in #383 btw.

amonshiz: oh yes, saw the original issue and read through it. just asking the question if there was another way to get this info considering the sizes may change.

if sizes do change, is a developer expected to kick off the calculations again? how is that handled in this system? does there need to be a way to cancel what may be long running background processes that are no longer valid?

Source/IGListPreprocessingContext.h
```diff
@@ -0,0 +1,41 @@
+//
+//  IGListPreprocessingContext.h
+//  IGListKit
+//
+//  Created by Adlai Holler on 2/20/17.
+//  Copyright © 2017 Instagram. All rights reserved.
+//
```
jessesquires: nit: copy project comment headers 😄

Source/Internal/IGListPreprocessor.h
rnystrom: If size changes and your preprocessing uses the size changes, you'll have to kick off updates again. You could call `adapter.performUpdates(...)` with each bounds change (it'll coalesce) to do that. Atm IGListKit doesn't do much w/ layout (aside from the provided collection view layout).

> does there need to be a way to cancel what may be long running background processes that are no longer valid

Maybe. Ideally nothing should be that long running: imagine just calculating text sizes for caches.

The only other to do this would be to lock on the [containerSize](https://github.com/Instagram/IGListKit/blob/master/Source/IGListAdapter.m#L810) API, but that wont even solve racing since the value could change between begin/end of processing.

Source/IGListPreprocessingContext.h
rnystrom: Some day I'll get danger working...

Adlai-Holler:
OK gentlemen could I get another look? @rnystrom @ocrickard @jessesquires 

Notes:
- **Blocking main** Of course its unpleasant to block the main thread, but it's important in some cases, such as when we want to perform synchronous updates.
  - Safe as long as their `IGListPreprocessingDelegate` doesn't wait on the main queue, synchronously or asynchronously.
- Should we do preprocessing before item-level updates? Seems like it would be valuable, but maybe surprising?
- See my `TODO` about section map. Do y'all agree with my interpretation – that what we have to do is generate the future section map in advance, without actually applying the state to the section controllers, and then simply apply it during `objectTransitionBlock`?

rnystrom:
Sorry for the delay! I'm way overdue on reviewing this. On my docket for the week.

jessesquires:
Hey @Adlai-Holler -- we've finally tagged and released `3.0`. Do you still want to follow-up on and support this PR? If so, now should be a pretty good time -- probably won't be much activity for a bit 😄

Adlai-Holler:
@jessesquires I've been waiting for review for a good long time on this one! 

Unfortunately though, I just don't have the resources to work more on this in the near future. I'll close this diff but of course I'd encourage anyone who picks this up to salvage anything they can from my work here. Cheers gents!

jessesquires:
Thanks @Adlai-Holler ! 👍  💯

rnystrom:
@Adlai-Holler apologies for dropping the ball here, I absolutely still want this and am going to use your work to push this through. So sorry for leaving this in limbo, internal priorities/etc distracted me 😭

PR #510, opened 2017-02-23
fix #460 by making a copy of the section map

### LINKED ISSUE #460: [Core Data] adapter.performUpdates does not trigger any diffing on NSManagedObjects
Hello,

I have a question about the framework, because from the doc I would expect a behaviour, but it does not happen.
I cannot provide a example project, because my app is huge and uses CoreData etc.. and it would take me a lot of time to set it up.

The idea is the following:
- I display CoreData object with IGListKit based on a timestamp
- CoreData objects are updated with background network requests
- The AdapterDataSource is using a NSFetchedResultController to monitor for object changes
- When the NSFetchedResultControllerDelegate is informed of CoreData changes it calls: adapter.performUpdates

What happens is: 
- Datasource method objects(for listAdapter) is called again, the objects obtain the new correct order
- UI is animated and sections shuffle around in the right order

however: 
- IGListDiffable Objects isEqual() method is never called 
- Section Controller are not reloaded

**I would expect that calling performUpdates:**
- 🆗  the datasource refreshes list of object 
- 🆗  the UI is animated
- ⚠️  the objects are checked to see if they have changed
- ⚠️  the content of the section controller is recalculated if needed
- ⚠️  UI content is updated correcly

I could call reloadData but I would lose the animation, and the purpose of using the framework.

There is something I understood wrong, or that I am not doing?

You can reproduce this in @rnystrom sample for RayWendlich:
https://koenig-media.raywenderlich.com/uploads/2016/12/MarsLink_Final.zip

I put the code here: 
https://github.com/racer1988/IGListKitPerformUpdates

-------
EDIT: 

Using a `IGListReloadDataUpdater` instead of a `IGListAdapterUpdater` provides the expected result, without animation and losing the scroll position (due to reload data)

I hence believe there is a bug (?) in the IGListAdapterUpdater that do not take in account model changes but just inserts and delete. Is this the expected behaviour?

Do I need to write my own adapter to manage the changes to the datamodel?

racer1988:
performUpdates was not making any copy of the old map
or the new objects

Due to this, no objects inside the array were getting their
copy() (NSCopying) method called.

This prevented updates to objects in the collectionView.
By making a shallow copy of the map, we trigger the copy()
method of every objects, permitting the comparison of
copied object provided by the user implementation of NSCopying
and the new objects provided by IGListKit Adapter Data Source

Issue fixed: #460 

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -39,6 +39,8 @@ This release closes the [3.0.0 milestone](https://github.com/Instagram/IGListKit
 
 - Gracefully handle a `nil` section controller returned by an `IGListAdapterDataSource`. [Ryan Nystrom](https://github.com/rnystrom) [(tbd)](https://github.com/Instagram/IGListKit/pull/tbd)
 
+- Fixes `-[IGListAdapter performUpdatesAnimated: completion:]`, not triggering the NSCopying copy() method for all objects inside the adapter array, this enables peformUpdates to correctly show in the collectionView edits to model objects that are implementing NSCopying. [Marco Pappalardo](https://github.com/racer1988) [(#510)](https://github.com/Instagram/IGListKit/pull/510)
+
 2.2.0
 -----
 
--- a/Source/IGListAdapter.m
+++ b/Source/IGListAdapter.m
@@ -259,7 +259,7 @@
         return;
     }
 
-    NSArray *fromObjects = self.sectionMap.objects;
+    NSArray *fromObjects = [[NSArray alloc] initWithArray:self.sectionMap.objects copyItems:YES];
     NSArray *newObjects = [dataSource objectsForListAdapter:self];
 
     __weak __typeof__(self) weakSelf = self;


racer1988:
I did not add any tests. I checked the IGListAdapterUpdaterTests and I think they cover what the updater is doing in perfomUpdates.

It would be interesting to add a test to show that a Mock object implementing NSCopying is getting copy() called. Please let me know if you think this is valuable.

rnystrom:
😬 this has really big implications:

- We would have to change `IGListDiffable` to extend `NSCopying`
- Performance could be a serious issue

Feeds with thousands of objects would be making copies (**lots** of `malloc`) for each item instead of the container.

I don't think we're going to be able to merge this 😞 but I definitely want to help come up with a better strategy to dealing with `NSManagedObject`s.

jessesquires:
@rnystrom -- We could support this with a new data source method.

```objc
- (BOOL)copyObjectsWhenDiffing;
```

This could easily be a 3.0 breaking change. Client can migrate by returning `NO`. In docs we can call out the `NSManagedObject` use case.

Further, if we resolve #168 with a superclass-only, then we can default to `NO`.

rnystrom:
I wonder if the data source could just return copies of the objects? Left the same comment [on the issue](https://github.com/Instagram/IGListKit/issues/460#issuecomment-282130944).

racer1988:
I left also a [comment](https://github.com/Instagram/IGListKit/issues/460#issuecomment-277477161) on the PR.

Implications of having DataSource returning always copies (or doing it in perform updates like I did in the PR) is that the collectionView will flicker for every call to the method, because the animation makes the collectionView remove the object and reinsert them (being them new)

racer1988:
Yep, I'll go on and close this PR

Issue #503, opened 2017-02-21
Out of bounds collectionview update (crash)


heumn:
- [X] I have reviewed the [`README`](https://github.com/Instagram/IGListKit/blob/master/README.md) and [documentation](http://instagram.github.io/IGListKit)
- [X] I have searched [existing issues](https://github.com/Instagram/IGListKit/issues) and this is not a duplicate

### General information

- `IGListKit` version: 2.1.0
- iOS version(s): 10.0
- CocoaPods/Carthage version: -
- Xcode version: 8.2.1
- Devices/Simulators affected: All
- Reproducible in the demo project? (Yes/No): No

I have attached an example project where this crash occurs (based on the demo project with minor adjustment): https://www.dropbox.com/s/bnis60igq8hpa6c/crash%20in%20nested%20example.zip?dl=0

How to reproduce: 
1 - Launch app
2 - Select the NestedAdapterViewController
3 - Select refresh
4 - Scroll down
5 - Select refresh (re-do steps 4-5 if it does not crash immediatly)
Crash with the classical "out of bounds"

What have I done? I have adjusted the example project to use a class (EmbeddedSection) to represent the models for the nested HorizontalSectionControllers so they in turn can embody viewmodels for their respective IGListSectionControllers (EmbeddedSectionController).

I have likely misunderstood some fundamentals about this so I am terribly sorry if this is doing something IGListKit was not intended for.

The crash appears to me to be because the  IGListAdapter is getting a re-used IGListCollectionView from the EmbeddedCollectionViewCell. 

My only theory to go from here is that something is retaining (retain cycle) the nested HorizontalSectionController and the adapter is trying to update a reused CollectionView that should have been deallocated when you scroll down/refresh.

Before I burn a lot more energy on this I would love be told how wrong I am for even trying to do this in the first place 😂

rnystrom:
@heumn thanks for reporting and including an example project! I'm going to spin through this and see what's going on.

heumn:
Thanks. Really interested in what you will find. The crash broke my head 😅

PhilCai1993:
This will happen.
According to the code below:
```
    static func from(int: Int) -> EmbeddedSection? {
        let data: [Any] = (0 ... Int(arc4random_uniform(36))).map { _ in Int(arc4random_uniform(100)) }
        return EmbeddedSection(identifier: String(int) as NSString, items: data as! [IGListDiffable])
    }
```

[...]

PR #517, opened 2017-02-27
Prevent stale adapter:collectionView corruptions

rnystrom:

This becomes an issue pretty easily in embedded `IGListAdapter` environments (lists in lists). IGListKit creates a single `IGListSectionController` instance per object, but cells are reused within `IGListCollectionView`. If you assign `adapter.collectionView = cell.collectionView` (like we recommend in our [example](https://github.com/Instagram/IGListKit/blob/master/Examples/Examples-iOS/IGListKitExamples/SectionControllers/HorizontalSectionController.swift#L40)), then you can have many adapters with a weak ref to the **same collection view**.

Obviously when updates happen across all embedded lists, adapters could potentially update the same collection view and apply a corrupted batch update.

I proposed using `<objc/runtime.h>` and using associated objects, but since `OBJC_ASSOCIATION_ASSIGN` is not zerod-on-release reference (its just `unsafe_unretained`) I'd have to make a weak wrapped object, which is overkill.

Instead I think a global weak:weak map is fine. We don't have to worry about locking b/c of the main-thread affinity of the framework. Please push back on me if you have a better idea.

Issue fixed: #503

### Checklist

### DIFF
--- a/Source/IGListAdapter.m
+++ b/Source/IGListAdapter.m
@@ -62,13 +62,23 @@
     return (IGListCollectionView *)_collectionView;
 }
 
+static void *kIGListCollectionViewAdapterKey = &kIGListCollectionViewAdapterKey;
+
 - (void)setCollectionView:(IGListCollectionView *)collectionView {
     IGAssertMainThread();
-    IGParameterAssert([collectionView isKindOfClass:[IGListCollectionView class]]);
 
     // if collection view has been used by a different list adapter, treat it as if we were using a new collection view
     // this happens when embedding a IGListCollectionView inside a UICollectionViewCell that is reused
     if (_collectionView != collectionView || _collectionView.dataSource != self) {
+        // if the collection view was being used with another IGListAdapter (e.g. cell reuse)
+        // destroy the previous association so the old adapter doesn't update the wrong collection view
+        static NSMapTable<IGListCollectionView *, IGListAdapter *> *globalCollectionViewAdapterMap = nil;
+        if (globalCollectionViewAdapterMap == nil) {
+            globalCollectionViewAdapterMap = [NSMapTable weakToWeakObjectsMapTable];
+        }
+        [[globalCollectionViewAdapterMap objectForKey:collectionView] setCollectionView:nil];
+        [globalCollectionViewAdapterMap setObject:self forKey:collectionView];
+
         // dump old registered section controllers in the case that we are changing collection views or setting for
         // the first time
         _registeredCellClasses = [NSMutableSet new];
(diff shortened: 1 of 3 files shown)


rnystrom:
Importing b/c our CI is faster than Travis.

jessesquires:
> our CI is faster than Travis.

lol

Source/IGListAdapter.m
```diff
@@ -62,13 +62,23 @@ - (IGListCollectionView *)collectionView {
     return (IGListCollectionView *)_collectionView;
 }
 
+static void *kIGListCollectionViewAdapterKey = &kIGListCollectionViewAdapterKey;
```
jessesquires: where is this used? 🤔

Source/IGListAdapter.m
```diff
@@ -62,13 +62,23 @@ - (IGListCollectionView *)collectionView {
...
+
 - (void)setCollectionView:(IGListCollectionView *)collectionView {
     IGAssertMainThread();
-    IGParameterAssert([collectionView isKindOfClass:[IGListCollectionView class]]);
```
jessesquires: unrelated to this PR? 🤔

Source/IGListAdapter.m
```diff
@@ -62,13 +62,23 @@ - (IGListCollectionView *)collectionView {
...
 
     // if collection view has been used by a different list adapter, treat it as if we were using a new collection view
     // this happens when embedding a IGListCollectionView inside a UICollectionViewCell that is reused
     if (_collectionView != collectionView || _collectionView.dataSource != self) {
+        // if the collection view was being used with another IGListAdapter (e.g. cell reuse)
+        // destroy the previous association so the old adapter doesn't update the wrong collection view
+        static NSMapTable<IGListCollectionView *, IGListAdapter *> *globalCollectionViewAdapterMap = nil;
```
jessesquires: we're main-thread-only, but should we still consider using `dispatch_once` ?

jessesquires:
💯

Source/IGListAdapter.m
rnystrom: It's not 😅

Source/IGListAdapter.m
rnystrom: Have to remove so we can set to `nil`

Source/IGListAdapter.m
rnystrom: Totally fine if main-thread affined, `dispatch_once` will add unused overhead

heumn:
thisisgiphy-reaction-audience-l4HodBpDmoMA5p9bG

ninokierulf:
after applying patch f639cdd,
the embedded collectionview's cell's focused/selected state is being reset.

PR #525, opened 2017-03-04
Track batch updates in object and explicit state

rnystrom:

Batch updates are complicated b/c its unknown when the update block will actually execute. When executing the block, we want to collect inserts/deletes/reloads/moves (item and section). Allow mutations to happen synchronously outside of the update block.

Tracking state will also help with the auto-diff where we need to allow re-entrant updates.

Peeling off a chunk from #494

Issue fixed: #288

### Checklist

### DIFF
--- a/Source/Internal/IGListBatchUpdatesCollector.h
+++ b/Source/Internal/IGListBatchUpdatesCollector.h
@@ -1,0 +1,26 @@
+/**
+ * Copyright (c) 2016-present, Facebook, Inc.
+ * All rights reserved.
+ *
+ * This source code is licensed under the BSD-style license found in the
+ * LICENSE file in the root directory of this source tree. An additional grant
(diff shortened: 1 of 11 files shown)


rnystrom:
Importing for faster CI + internal review

Source/Internal/IGListBatchUpdatesCollector.h
```diff
@@ -0,0 +1,26 @@
...
+ */
+
+#import <Foundation/Foundation.h>
+
+#import <IGListKit/IGListMacros.h>
+
+@class IGListMoveIndexPath;
+
+IGLK_SUBCLASSING_RESTRICTED
+@interface IGListBatchUpdatesCollector : NSObject
```
jessesquires: nit: just call this `IGListBatchUpdates` ?

PR #539, opened 2017-03-12
consistent use of finall class

heumn:
You are already using `final class` as the default swift declaration as I guess most people agree this is best practice for classes you are not planning to inherit from :) So I just made the example code more consistent. 

I will not add much more reasoning to why, other than open source projects like this is a _big_ inspirations and starting ground to play with swift for new develoeprs joining the iOS community. Nuding them to ask the right questions (composition vs inheritance) early could only be a good thing (?)

### Checklist

- [X] No new tests. No new classes.
- [X] No breaking changes.


rnystrom:
Thanks @heumn!

PR #570, opened 2017-03-21
Added two useful UIScrollViewDelegate methods to IGListAdapterProxy

antoinelamy:

Issue fixed: Added two useful UIScrollViewDelegate methods to IGListAdapterProxy. `IGListAdapterProxyTests` already cover the changes.

### Checklist

- [ X ] All tests pass. Demo project builds and runs.
- [ X ] I added tests, an experiment, or detailed why my change isn't tested.
- [ X ] I added an entry to the `CHANGELOG.md` for any breaking changes, enhancements, or bug fixes.
- [ X ] I have reviewed the [contributing guide](https://github.com/Instagram/IGListKit/blob/master/.github/CONTRIBUTING.md)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,6 +28,8 @@ This release closes the [3.0.0 milestone](https://github.com/Instagram/IGListKit
 
 - Renamed `IGListAdapterUpdaterDelegate` method to `listAdapterUpdater:didPerformBatchUpdates:collectionView:`. [Vincent Peng](https://github.com/vincent-peng) [(#491)](https://github.com/Instagram/IGListKit/pull/491)
 
+- Added `scrollViewDidEndScrollingAnimation:` and `scrollViewDidEndDecelerating:` UIScrollViewDelegate methods to IGListAdapterProxy. [Antoine Lamy](https://github.com/antoinelamy) [(#570)](https://github.com/Instagram/IGListKit/pull/570)
+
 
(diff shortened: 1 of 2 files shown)


antoinelamy:
Not sure the failure is related to this PR... On a side note, what do you guys think of using the runtime function `protocol_get​Method​Description(::::)` to dynamically determine if a given selector should be forwarded or not?

rnystrom:
🤔 very interesting! Would like to know the performance characteristics of doing that. We use the static function b/c its super cheap vs inspecting an instance at runtime.

antoinelamy:
@rnystrom my bad! I'm using `IGListKit` in conjunction with `AsyncDisplayKit` and I thought that some methods of my scrollViewDelegate was not called because of the `IGListAdpaterProxy` but it's rather `ASIGListAdapterBasedDataSource` that is responsible of catching the delegate's calls and not forwarding them. Closing this PR , sorry!

rnystrom:
No worries! Thanks for taking the time to send this up and investigate!

antoinelamy:
@rnystrom it should be fairly cheap but more costly than direct selector comparison and a call to `respondToSelector`. Anyway, there is no real advantage of doing that since every selector sent to the delegate instance are forwarded, I initially taught that only the selectors in the `isInterceptedSelector()` were forwarded. Thank you for your input and great library BTW!

PR #580, opened 2017-03-23
Fix unused variable warning when asserts compiled out

rnystrom:

Found this examining Travis output. Warning removed when asserts aren't compiled.

#trivial


jessesquires:
nice 👍

Issue #400, opened 2017-01-09
SizeForItem with custom layout


risp:
Hi, I was wondering how IGListKit works with a custom UICollectionViewLayout. I've a `CollectionViewHorizontalLayout` that asks the CollectionView for the size using this protocol:

```swift
protocol CollectionViewHorizontalLayoutDelegate { 
    func collectionView(_ collectionView: UICollectionView, layout collectionViewLayout: UICollectionViewLayout, sizeForItemAt indexPath: IndexPath) -> CGSize
    func collectionView(_ collectionView: UICollectionView, layout collectionViewLayout: UICollectionViewLayout, referenceSizeForHeaderInSection section: Int) -> CGSize
}
```
Before using IGListKit I returned the GCSize in the CollectionView. But with IGListKit the `IGListSectionController` controller should return the size. I'm not really sure how to do this. Especially since I'm using multiple IGListSectionControllers? 

And how does the size of an item and the UICollectionViewLayout relate to 
``` swift
public func sizeForItem(at index: Int) -> CGSize
```

I also searched in the examples but couldn't find anything related to customlayouts and dynamic sizes.

rnystrom:
@risp we've added an API on `IGListAdapter` to do just that! [Here's the header docs](https://github.com/Instagram/IGListKit/blob/master/Source/IGListAdapter.h#L212-L230).

So your setup should probably be something like:

```swift
class ViewController: UIViewController, MyLayoutDelegate {

  let adapter = IGListAdapter(...)	

  // MARK: MyLayoutDelegate

  func collectionView(_ collectionView: UICollectionView, layout collectionViewLayout: UICollectionViewLayout, sizeForItemAt indexPath: IndexPath) -> CGSize {
    return adapter.sizeForItem(at: indexPath)
  }

  func collectionView(_ collectionView: UICollectionView, layout collectionViewLayout: UICollectionViewLayout, referenceSizeForHeaderInSection section: Int) -> CGSize {
    return adapter.sizeForSupplementaryView(ofKind: "header_kind_string", at: IndexPath(item: 0, section: section))
  }

}
```

risp:
Thanks a lot, it's all clear now and works great!

HumidBrains:
@rnystrom It would be so helpful with an example project using custom UICollectionViewLayout. Is it possible to dig one out? Thanks :)

rnystrom:
[...]

PR #592, opened 2017-03-28
Fix bug where IGListAdapterProxy cast to UICollectionViewDelegateFlowLayout in Swift always fails.

Istered:

Issue fixed: Added conformsToProtocol implementation to `IGListAdapterProxy` that fixing Swift cast to `UICollectionViewDelegateFlowLayout`.

I noticed that cast to UICollectionViewDelegateFlowLayout always fails when I was trying to implement custom collection layout, so this PR should fix this problem.
But I'm wondering what kind of tests should I add? All the tests are written in Objective C while this problem persists only in Swift code. 

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -90,6 +90,8 @@ This release closes the [3.0.0 milestone](https://github.com/Instagram/IGListKit
 
 - Fix a potential crash when a section is moved and deleted at the same time. [Ryan Nystrom](https://github.com/rnystrom) [(#577)](https://github.com/Instagram/IGListKit/pull/577)
 
+- Fix bug where `IGListAdapterProxy` cast to `UICollectionViewDelegateFlowLayout` in Swift always fails. [Anton Romanov](https://github.com/Istered) [(#592)](https://github.com/Instagram/IGListKit/pull/592)
+
 2.1.0
 -----
 
@@ -242,5 +244,3 @@ You can find a [migration guide here](https://instagram.github.io/IGListKit/migr
 -----
 
 Initial release. :tada:
-
-
--- a/Source/Internal/IGListAdapterProxy.m
+++ b/Source/Internal/IGListAdapterProxy.m
@@ -58,6 +58,10 @@ static BOOL isInterceptedSelector(SEL sel) {
     return self;
 }
 
+- (BOOL)conformsToProtocol:(Protocol *)aProtocol {
+    return aProtocol == @protocol(UICollectionViewDelegateFlowLayout);
+}
+
 - (BOOL)respondsToSelector:(SEL)aSelector {
     return isInterceptedSelector(aSelector)
     || [_collectionViewTarget respondsToSelector:aSelector]


jessesquires:
Thanks @Istered ! 😄 

There are some issues with this. 😅 

1. this is not a correct implementation of `conformsToProtocol:` 
    - it should be a class method 
    - there's no call to super 
2. this is an internal file to the lib and an internal implementation detail that clients shouldn't be relying on

Can you explain what you're trying to do here? Sounds like there's some underlying issue and this is a hack to workaround it.

Istered:
Hello @jessesquires 
I guess I should have created an issue instead of PR 😅  
Nevertheless, the problem is `IGListAdapterProxy` is not conforming to `UICollectionViewDelegateFlowLayout`, because of that the next statement will always fail in Swift:

```Swift
if let flowDelegate = collectionView?.delegate as? UICollectionViewDelegateFlowLayout {
...
}
```

Usually, you don't want to interact with delegate directly until it comes to custom collection layouts, so even though `IGListAdapterProxy` is an internal class it doesn't work as client expects it to.

jessesquires:
Hmm... 🤔 

So why do you need to do this? What are you trying to do? 😄

cc @rnystrom

Istered:
@jessesquires I want to ask collectionView's delegate for items sizes.  The exact same idea as for `IGListCollectionViewLayout`, but in Swift. 
https://github.com/Instagram/IGListKit/blob/master/Source/IGListCollectionViewLayout.mm#L316
https://github.com/Instagram/IGListKit/blob/master/Source/IGListCollectionViewLayout.mm#L358

jessesquires:
Hm, yeah that's weird. 😊 

The delegate is what asks **you** for item sizes. So you should be able to derive this from your own code... 🤔  (Let me know if I'm misunderstanding)

Istered:
Well, it's true. But still that's the only way to get item sizes without messing with hardcoded values or anything like that since I'm not bulding layout for one specific case. I'm planning to use it with `UICollectionView` and `IGListAdapter` as well, so in any way I want to work with collection's delegate as with `UICollectionViewDelegateFlowLayout`. 🙂

The workaround was either making `IGListAdapterProxy` conform to `UICollectionViewDelegateFlowLayout` or making an objective-c helper that will just call this method for me. Right now I'm using the second approach, but I wouldn't say I'm happy about it. 😖
I hope there's a better way to achieve my goal.🙂

rnystrom:
@Istered if all you're concerned about is the item sizes, why not use [this API](https://github.com/Instagram/IGListKit/blob/master/Source/IGListAdapter.h#L237) on `IGListAdapter`?

For custom collection view layouts, we recommend following a pattern [like this](https://github.com/Instagram/IGListKit/issues/400#issuecomment-271412514).

Also, even if you're using Swift, `UICollectionViewLayout`s ref a `UICollectionView`, so you should be able to `as?` cast the delegate to w/e protocol you want. I'm not sure the compiler or anything will ever  get in the way of that.

Are you saying that

```swift
if let flowDelegate = collectionView?.delegate as? UICollectionViewDelegateFlowLayout {
  // ...
}
```

doesn't work using Swift and inside a `UICollectionViewLayout` subclass? Would love an example if that's the case.

Istered:
Thanks! This workaround is clean and nice.

But just to be clear, described problem is not related only to a code that is placed inside custom layout.
Just place code snippet from above into any of example projects and you'll see that cast is always fails.

rnystrom:
That's intended though, we don't want to expose details of the implementation.

PR #593, opened 2017-03-29
Swift name annotations

robertjpayne:

This adds `NS_SWIFT_NAME` annotations to all public API's to provide cleaner integration into Swift:

- Removes the need to prefix classes in Swift code, instead rely on Swift module name spacing
- Adds more argument labels to C function API's like `IGListDiff([], [], .equality)` => `ListDiff(oldArray: [], newArray: [], option: .equality)`

While this is a large API change it should be as easy as:

- Find and replace `(IGList)([^K])` to `List$2` in Xcode with a scope set to Swift
- Build and follow compiler's auto fix corrections for C API's or any missed renames

I have not updated the documentation to reflect this yet, I am totally willing to do so but before I sink that amount of time into it I wanted to see if the Instagram team is even open to this change!

### Checklist

- [ ] I have updated the documentation

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,25 @@ This release closes the [3.0.0 milestone](https://github.com/Instagram/IGListKit
 
 ### Breaking Changes
 
+- Added Swift annotation names which remove `IG` prefixes from class names, C functions, and other APIs. Note, this only affects Swift clients. [Robert Payne](https://github.com/robertjpayne) [(#593)](https://github.com/Instagram/IGListKit/pull/593)
+
+Example:
+
+```swift
+// OLD
+class MySectionController : IGListSectionController { ... }
+
+// NEW
+class MySectionController : ListSectionController { ... }
+
+// OLD
+IGListDiff([], [], .equality)
+
+// NEW
+ListDiff(oldArray: [], newArray: [], .equality)
+
+```
+
 - Updated `didSelect` delegate call in `IGListSingleSectionControllerDelegate` to include object. [Sherlouk](https://github.com/Sherlouk) [(#397)](https://github.com/Instagram/IGListKit/pull/397)
 
 ```objc
--- a/Examples/Examples-iOS/IGListKitExamples/Models/FeedItem.swift
+++ b/Examples/Examples-iOS/IGListKitExamples/Models/FeedItem.swift
@@ -14,7 +14,7 @@
 
 import IGListKit
 
-final class FeedItem: IGListDiffable {
+final class FeedItem: ListDiffable {
 
     let pk: Int
     let user: User
@@ -26,13 +26,13 @@ final class FeedItem: IGListDiffable {
         self.comments = comments
     }
 
-    //MARK: IGListDiffable
+    //MARK: ListDiffable
 
     func diffIdentifier() -> NSObjectProtocol {
         return pk as NSObjectProtocol
     }
 
-    func isEqual(toDiffableObject object: IGListDiffable?) -> Bool {
+    func isEqual(toDiffableObject object: ListDiffable?) -> Bool {
         guard self !== object else { return true }
         guard let object = object as? FeedItem else { return false }
         return user.isEqual(toDiffableObject: object.user) && comments == object.comments
--- a/Examples/Examples-iOS/IGListKitExamples/Models/Month.swift
+++ b/Examples/Examples-iOS/IGListKitExamples/Models/Month.swift
@@ -31,13 +31,13 @@ final class Month {
     
 }
 
-extension Month: IGListDiffable {
+extension Month: ListDiffable {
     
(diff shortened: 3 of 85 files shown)


jessesquires:
@robertjpayne  Nice! 👍 🥇 

Glanced through this and looks pretty good.

cc @rnystrom I'm down for this, but it will be a huge breaking change for 3.0 (but that's probably OK)

robertjpayne:
@jessesquires the *only* think that is actually kinda frustrating is I sort of feel like `IGListSectionController` should be renamed to `IGListBaseSectionController` because it always get's subclassed and having it as `ListSectionController` will probably cause a bit of chaos for anyone that has subclassed it for their own base implementation.

Thoughts on that?

heumn:
> leave IGListKit alone.

image


So sorry. Couldn't resist. I am weak weak man.

CHANGELOG.md
```diff
@@ -9,6 +9,23 @@ This release closes the [3.0.0 milestone](https://github.com/Instagram/IGListKit
 
 ### Breaking Changes
 
+- Added Swift annotation names removing IGL prefixes from API's and improving those of C functions. [Robert Payne](https://github.com/robertjpayne)
```
rnystrom: nit: can you link to the PR? See other changelog entries as an example

robertjpayne:
I think it's probably wise to update the guides at the same time -- I may need a few days to get there though busy end of the week.

jessesquires:
@robertjpayne - No worries!

robertjpayne:
@jessesquires @rnystrom So the documentation is auto generated, I'm not super familiar with Jazzy so I'm not going to touch that, I would hope it can auto-generate the Swift API's vs the ObjC apis?

As far as the guides go, do we assume they are all written for Swift users? I'm curious about what should happen to documentation lines like:

> You subclass `IGListSectionController` and conform to the `IGListSectionType` protocol.

jessesquires:
@robertjpayne Hm, that's a good point.

---------

We should leave the prefixes for everything, *except* example docs, etc. that use Swift. So that means we should leave header doc comments alone.

Actually... jazzy *might* handle this correctly since it will show ObjC *and* Swift APIs. I won't know for sure until this is merged and I can run the docs scripts to see.

Maybe let's hold off on this for now -- I and @rnystrom can take care of refining docs for the release.

robertjpayne:
@jessesquires sounds good -- lemme know if I can help with anything else!

rnystrom:
@robertjpayne ack, mind rebasing? Should just be replacing `ListCollectionView` with `UICollectionView` in all the demos.

jessesquires:
Sorry for those merge conflicts 😊

robertjpayne:
I'll do it this weekend sorry! I think a merge is going to be easier than a rebase given how the changes were made…

robertjpayne:
Alright @jessesquires @rnystrom should be up to date with master now ( ran all the tests and built all the examples ) -- apologies the merge was rather manual instead of a clean rebase because I didn't really feel like using diff merge on every example file!

Also updated the changeling with a link to this PR!

CHANGELOG.md
robertjpayne: Should be done now!

robertjpayne:
Can we get Travis to re-test? I think the timeouts are invalid -- locally all the tests are passing…

jessesquires:
👍

rnystrom:
@jessesquires should we still try to do this for 3.0? Huge rebase necessary. Prepping for documentation updates if we're going to do it.

Tbh I could go either way. I like the idea, but it does make communication a little trickier (ObjC vs Swift language). I'm still on board tho.

jessesquires:
@rnystrom - yeah I still think this is a good change for Swift clients. The foundation overlays from Apple do this for certain classes. (NSData --> Data, NSUserDefaults --> UserDefaults, etc.)

As long as all that's happening is removing the `IG` prefix, I think we're ok regarding communication.

robertjpayne:
There was a few API changes on the C methods (just added argument labels) but Xcode adds "FIXME"'s for any of those it breaks.

I could have left the argument labels off but I felt it actually helped explain the parameters `IGListDiff(...)`

jessesquires:
Thanks @robertjpayne ! 💯 

do you mind rebasing and we'll get this imported and merged?

robertjpayne:
@jessesquires when is it best to land? I can rebase it again (it's actually easier to just use the regex I have to redo it than rebase it). Are you planning on merging some day specific this week so I can line it up to avoid another rebase?

rnystrom:
@robertjpayne we can land at any time (we manually click a button), so its on me not merging this fast enough. If you want to, rebase & push your changes then [tweet/DM me](https://twitter.com/_ryannystrom) and I'll init a land. Sorry for the constant rebasing 😞

Issue #542, opened 2017-03-13
Vision doc


rnystrom:
Wonder if it'd be worth documenting what the long-term vision is of IGListKit so we can reference it when closing PRs or Issues that are out of scope? Similar to [danger](https://github.com/danger/danger/blob/master/VISION.md)'s.

Thoughts?

jessesquires:
💯 ✖️ 💯

PR #607, opened 2017-04-01
Add vision doc

rnystrom:
Adding a `VISION.md` doc at the root dir to give collaborators and users guidance about what we will work on and how we will prioritize features/fixes for IGListKit.

Issue fixed: #542

#trivial


jessesquires:
If you move this to `/Guides/` we can render it with `jazzy`.

Might be a weird place to put this, but I think it would be nice to render with the rest of the docs.

jessesquires:
Also, we should link to this from the README

jessesquires:
updated PR. going to import.

PR #612, opened 2017-04-03
Demo for dependency injection and announcing changes

rnystrom:
Created a new demo that shows off how to:

- Inject dependencies to section controllers
  - Objects shared between section controllers
  - Things that aren't the core "object"
- Announce changes from the VC that make their way to the section controllers
  - Without calling updates

### DIFF
--- a/Examples/Examples-iOS/IGListKitExamples/SectionControllers/ListeningSectionController.swift
+++ b/Examples/Examples-iOS/IGListKitExamples/SectionControllers/ListeningSectionController.swift
@@ -1,0 +1,58 @@
+/**
+ Copyright (c) 2016-present, Facebook, Inc. All rights reserved.
+
+ The examples provided by Facebook are for non-commercial testing and evaluation
+ purposes only. Facebook reserves all rights not expressly granted.
+
+ THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL
+ FACEBOOK BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN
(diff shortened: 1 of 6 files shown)


Examples/Examples-iOS/IGListKitExamples/SectionControllers/ListeningSectionController.swift
```diff
@@ -0,0 +1,58 @@
...
+ IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL
+ FACEBOOK BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN
+ ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION
+ WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+ */
+
+import IGListKit
+
+class ListeningSectionController: IGListSectionController, IGListSectionType, IncrementListener {
```
heumn: nit: `final class` this one as well? 🤔

rnystrom: Oops good eye! I need to figure out how to make Danger linters

heumn: Thoughts about closing down the internal variables of demo classes? Nothings stops idiots like me access and updating `var value: Int = 0` directly and assuming `things will just work`. Is that too nit picky? Why I ask is again that you guys are such a huge inspiration for us "lesser mortals" and pushing us in the right direction can only be a good thing :)

rnystrom: I wouldn't mind a follow-up that does that. Tbh I really only use Swift on these example projects and am just used to hiding private stuff in implementation files. A pass of `fileprivate` or w/e on the demos would be good I think, we want to promote best-practices while not sacrificing the demo.

heumn: 👌  lets see if I can catch up on this

Issue #483, opened 2017-02-10
NSInternalInconsistencyException:  'collectionContext!.insert' method has no influence over UICollectionView immediately after 'adapter.performUpdates'


vldalx:
## New issue checklist

- [x] I have reviewed the [`README`](https://github.com/Instagram/IGListKit/blob/master/README.md) and [documentation](http://instagram.github.io/IGListKit)
- [x] I have searched [existing issues](https://github.com/Instagram/IGListKit/issues) and this is not a duplicate

### General information

- `IGListKit` version: 2.1.0 
- iOS version(s): 9.3.3 or later
- CocoaPods/Carthage version: 1.1.1
- Xcode version: 8.2.1
- Devices/Simulators affected: Both affected
- Reproducible in the demo project? (Yes/No): There is no a necessary environment.
- Related issues:

I have a grouped datasource in my app. A group is "equal" to `IGListSectionController`. There are two behaviors in my app. I add in runtime either new groups or new items in an existing group.

If I invoke `collectionContext!.insert` method immediately after `adapter.performUpdates` the first one will be ignored (no impact on `UICollectionView`). The next invoke of `collectionContext!.insert` will crash my app.
(See my comment for [#228](https://github.com/Instagram/IGListKit/issues/288#issuecomment-276175846))

**I've created an easy example how to reproduce the issue.**
There is a button in ViewController.
**Just click the button as fast as you can**. On each click, a new group will be added to a datasource as the first element, or a new item will be added to the first existing group.

Datasource structure:

```
class RefArray<T: Any> {
    var array: [T]
    
    init(array: [T]) {
        self.array = array
    }
    
    convenience init() {
        self.init(array: [T]())
    }
    
    subscript(index: Array<T>.Index) -> T {
        return array[index]
    }
}

final class IntsGroup: RefArray<Int>, IGListDiffable {
    let uid = (UUID().uuidString + Date().description) as NSString
    
    func diffIdentifier() -> NSObjectProtocol {
        return uid
    }
    
    func isEqual(toDiffableObject object: IGListDiffable?) -> Bool {
        guard let other = object as? IntsGroup else { return false }
        return self.array.count == other.array.count
    }
}
```

SectionController:

```
class MultiSectionController: IGListSectionController, IGListSectionType, IGListSupplementaryViewSource
{
    private var data: IntsGroup!

    override init() {
        super.init()
        supplementaryViewSource = self
    }
    
    func numberOfItems() -> Int {
        return data.array.count
    }
[...]

PR #616, opened 2017-04-03
Prevent a crash when inserting the same index twice

rnystrom:

If you insert or delete into the same index twice **and** update your data source to reflect those changes, we will squash the insert/delete into a single update because we're using `NSSet` internally.

This becomes very apparent when multiple updates are coalesced.

Issue fixed: #483

### Checklist

### DIFF
--- a/Source/Common/IGListBatchUpdateData.h
+++ b/Source/Common/IGListBatchUpdateData.h
@@ -41,17 +41,17 @@ IGLK_SUBCLASSING_RESTRICTED
 /**
  Item insert index paths.
  */
-@property (nonatomic, strong, readonly) NSSet<NSIndexPath *> *insertIndexPaths;
+@property (nonatomic, strong, readonly) NSArray<NSIndexPath *> *insertIndexPaths;
 
 /**
  Item delete index paths.
  */
-@property (nonatomic, strong, readonly) NSSet<NSIndexPath *> *deleteIndexPaths;
+@property (nonatomic, strong, readonly) NSArray<NSIndexPath *> *deleteIndexPaths;
 
 /**
  Item moves.
  */
-@property (nonatomic, strong, readonly) NSSet<IGListMoveIndexPath *> *moveIndexPaths;
+@property (nonatomic, strong, readonly) NSArray<IGListMoveIndexPath *> *moveIndexPaths;
 
 /**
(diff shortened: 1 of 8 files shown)


Source/Common/IGListBatchUpdateData.h
```diff
@@ -68,9 +68,9 @@ IGLK_SUBCLASSING_RESTRICTED
 - (instancetype)initWithInsertSections:(NSIndexSet *)insertSections
                         deleteSections:(NSIndexSet *)deleteSections
                           moveSections:(NSSet<IGListMoveIndex *> *)moveSections
-                      insertIndexPaths:(NSSet<NSIndexPath *> *)insertIndexPaths
-                      deleteIndexPaths:(NSSet<NSIndexPath *> *)deleteIndexPaths
-                        moveIndexPaths:(NSSet<IGListMoveIndexPath *> *)moveIndexPaths NS_DESIGNATED_INITIALIZER;
+                      insertIndexPaths:(NSArray<NSIndexPath *> *)insertIndexPaths
+                      deleteIndexPaths:(NSArray<NSIndexPath *> *)deleteIndexPaths
+                        moveIndexPaths:(NSArray<IGListMoveIndexPath *> *)moveIndexPaths NS_DESIGNATED_INITIALIZER;
```
jessesquires: this class is public, right?

need a changelog for breaking changes?

rnystrom: Oh weird, it is. But, why? This isn't supposed to be used on macOS.

jessesquires: why's that? there's nothing iOS-specific about this class, right? isn't it also needed for diffing?

rnystrom: Not needed for diffing, its like a "give me all your updates and I'll squish them". You're right it doesn't have anything iOS specific tho.

jessesquires: 🤷‍♂️

jessesquires:
updated change log. going to import

jessesquires:
cancelled import. something is broken

jessesquires:
Going to re-open this

jessesquires:
re-opened at #634

Issue #310, opened 2016-12-09
Add -[IGListAdapter debugInfo] method to print information to help solving GH issues


rnystrom:
This would really help when dealing with bizarre issues that don't/can't have a sample project attached (e.g. #305). People can dump the state of the adapter and upload into an issue/gist/etc so we can understand the state of their setup. Things like:

- Internal `IGListSectionMap` info
  - Objects and section controllers
  - Object-to-section controller mapping
- Object info
  - Diff identifiers
- Collection view state
  - `numberOfSections` and `numberOfItemsInSection:` for each section
- Track updates
  - Track and print all updates and transactions (reloadData and `IGListBatchUpdateData`)

Use Instagram debug tools as a starting point.

Sherlouk:
@rnystrom Is this complete or are there still aspects to be done?

PR #617, opened 2017-04-04
Add debug dump scripts

rnystrom:

Got a little time on the plane to mess around with this idea. Very much want comments on this! Some questions:

- We should probably compile out parts of this using `#if DEBUG`.
  - Should I compile out all of the description methods?
  - Maybe I wrap the entire debug files w/ `#if DEBUG ... #endif` so none of it is loaded?
- @ryanolsonk I used [FLEX](https://github.com/Flipboard/FLEX/blob/master/Classes/Utility/FLEXHeapEnumerator.m) for searching the heap, lmk if you see any problems w/ this
  - Its lifted w/ only a few changes for my specific needs (including comments 😂)
  - Chalk us up for using [another bit](https://github.com/Instagram/IGListKit/blob/master/Source/Internal/IGListAdapterProxy.m#L77-L82) of your code
- Added a basic test so that the basics of this work and coverage doesn't tank
  - Anything else I should test?
- Is there more data we should be dumping?
- I have to track batch update state on the updater in order to dump it. If this seems dangerous lmk and I can revert that part
  - Though the batch update info is super useful
- Hit me w/ your nits on style and stuff

Here's an [example dump](https://gist.github.com/rnystrom/7a0f4d39bf9e1320953be1e99d0e0bf8).

Issue fixed: #310

### Checklist

- [ ] ~~I added an entry to the `CHANGELOG.md` for any breaking changes, enhancements, or bug fixes.~~

### DIFF
--- a/Source/Internal/IGListDebugger.m
+++ b/Source/Internal/IGListDebugger.m
@@ -14,9 +14,20 @@
 
 @implementation IGListDebugger
 
+static NSHashTable<IGListAdapter *> *livingAdaptersTable = nil;
+
++ (void)trackAdapter:(IGListAdapter *)adapter {
+#if IGLK_DEBUG_DESCRIPTION_ENABLED
+    if (livingAdaptersTable == nil) {
+        livingAdaptersTable = [NSHashTable weakObjectsHashTable];
+    }
+    [livingAdaptersTable addObject:adapter];
+#endif // #if IGLK_DEBUG_DESCRIPTION_ENABLED
+}
+
 + (NSArray<NSString *> *)adapterDescriptions {
     NSMutableArray *descriptions = [NSMutableArray new];
-    for (IGListAdapter *adapter in IGListAllAdpaterInstances()) {
+    for (IGListAdapter *adapter in livingAdaptersTable) {
         [descriptions addObject:[adapter debugDescription]];
     }
     return descriptions;
--- a/Source/Internal/IGListDebugger.m
+++ b/Source/Internal/IGListDebugger.m
@@ -1,0 +1,29 @@
+/**
+ * Copyright (c) 2016-present, Facebook, Inc.
+ * All rights reserved.
+ *
+ * This source code is licensed under the BSD-style license found in the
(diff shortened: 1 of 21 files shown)


jessesquires:
🎉 

glanced through everything. looks rad.

rnystrom:
Talked w/ @ryanolsonk who recommended I read through @kastiglione's code on https://github.com/facebook/chisel but I'm not able to find the method for searching the heap.

I _think_ this should be ok as-is, but still needs compiled out if not debugging (pls never ship this lol).

rnystrom:
Alright updated w/ compile-out on the debug scripts. I think this is ready.

kastiglione:
I haven't gotten around to pushing the heap search to the public chisel repo. That code is currently internal. I'm not sure what your future plans are, but as it is, you could also do this without heap scanning by having `IGListAdapter.m` track its live instances when built for debug.

rnystrom:
@kastiglione that.... would probably be a lot simpler. Good call!

rnystrom:
Importing for more eyeballs

Source/Internal/IGListDebugger.m
```diff
@@ -0,0 +1,40 @@
...
+@implementation IGListDebugger
+
+static NSHashTable<IGListAdapter *> *livingAdaptersTable = nil;
+
++ (void)trackAdapter:(IGListAdapter *)adapter {
+#if IGLK_DEBUG_DESCRIPTION_ENABLED
+    if (livingAdaptersTable == nil) {
+        livingAdaptersTable = [NSHashTable weakObjectsHashTable];
+    }
+    [livingAdaptersTable addObject:adapter];
```
kastiglione: Are `IGListAdapter` instances only created on main thread? For future consideration: an alternative to tracking using weak references (which involves a global lock and I believe may cause the objc runtime to use slow path in some cases), you could track `__unsafe_unretained` pointers and have them be removed them from the hash table from `-[IGListAdapter dealloc]`.

rnystrom: Yup main thread affined. This is also a good idea though, I like it.

PR #637, opened 2017-04-10
Chore/variable accessibility, close #618

### LINKED ISSUE #618: Scope example variables (fileprivate, etc)
In an effort to promote architecture and design best-practices, we should recommend people scope their variables appropriately.

e.g. [this variable](https://github.com/Instagram/IGListKit/blob/master/Examples/Examples-iOS/IGListKitExamples/SectionControllers/LabelSectionController.swift#L20)

Should do for all

- Section controllers
- Cells
- View controllers
- Models

Followup from #612

heshamsalman:

Issue fixed: #618 

### Checklist

### DIFF
--- a/Examples/Examples-iOS/IGListKitExamples/Views/DetailLabelCell.swift
+++ b/Examples/Examples-iOS/IGListKitExamples/Views/DetailLabelCell.swift
@@ -18,7 +18,7 @@ final class DetailLabelCell: UICollectionViewCell {
 
     fileprivate let padding: CGFloat = 15.0
 
-    lazy var titleLabel: UILabel = {
+    lazy private var titleLabel: UILabel = {
         let view = UILabel()
         view.backgroundColor = .clear
         view.textAlignment = .left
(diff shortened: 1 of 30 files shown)


Examples/Examples-iOS/IGListKitExamples/Views/DetailLabelCell.swift
```diff
@@ -38,6 +38,19 @@ final class DetailLabelCell: UICollectionViewCell {
         return view
     }()
 
+
+    var title: String? {
+        didSet {
+            titleLabel.text = title
+        }
+    }
```
jessesquires: i'm not a fan of these changes on all of the cells. the properties are redundant, and create unnecessary storage.

if we do something like this, they should be computed, not stored:

```swift
var title: String? {
    get { 
        return titleLabel.text
    }
    set {
        titleLabel.text = newValue
    }
}
```

heshamsalman: Ooh. I like that quite a bit better 👍

PR #649, opened 2017-04-17
Replaced placeholders in changelog

Sherlouk:
Travis is complaining about dead links, so this fixes the one in changelog but there's still a broken link in guides to the core data guide which isn't published yet (via Jazzy).

Should we regen docs but reset everything which isn't the guides? Or are we happy to just leave that one until we release 3.0?


jessesquires:
>  Or are we happy to just leave that one until we release 3.0?

Let's leave until 3.0 because APIs have changed

Issue #651, opened 2017-04-17
"Delete twice" test flaky because of bug in UICollectionView


rnystrom:
Woot, we found another bug in `UICollectionView`! It looks like UICV has no problem handling multiple inserts at the same index path (e.g section = 0, item = 0) within the same batch update. 

However if you try to delete from the same index path twice in a single batch update, things will _eventually_ break.

This bug was exposed when we landed a415ef5552cbfbf2d3884a705241973f4ea72dde #483.

Here's a sample project demonstrating the issue. Radar time!

[UICollectionViewMultipleDeletesSameIndex.zip](https://github.com/Instagram/IGListKit/files/925777/UICollectionViewMultipleDeletesSameIndex.zip)

rnystrom:
cc @amonshiz @jessesquires this explains some of the test flakiness (t17306420)

jessesquires:
@rnystrom Nice find!

Somebody call Steve 😄

PR #657, opened 2017-04-17
Prevent duplicate item deletes and drop reload collisions

rnystrom:

a415ef5552cbfbf2d3884a705241973f4ea72dde exposed a bug in `UICollectionView` where its state gets corrupted when deleting the same index path more than once in a single batch update block. This resulted crashes like

```
* thread #1, queue = 'com.apple.main-thread', stop reason = EXC_BAD_ACCESS (code=1, address=0x4)
#1  0x000000010c63a451 in -[UICollectionViewData layoutAttributesForItemAtIndexPath:] ()
#2  0x000000010c63a648 in -[UICollectionViewData layoutAttributesForGlobalItemIndex:] ()
#3  0x000000010c5f18d8 in -[UICollectionView _viewAnimationsForCurrentUpdate] ()
#4  0x000000010c5f6c7f in __71-[UICollectionView _updateWithItems:tentativelyForReordering:animator:]_block_invoke.2012 ()
#5  0x000000010bd5bd3e in +[UIView(Animation) performWithoutAnimation:] ()
#6  0x000000010c5f58b9 in -[UICollectionView _updateWithItems:tentativelyForReordering:animator:] ()
#7  0x000000010c5efbbf in -[UICollectionView _endItemAnimationsWithInvalidationContext:tentativelyForReordering:animator:] ()
#8  0x000000010c5f8259 in -[UICollectionView _endUpdatesWithInvalidationContext:tentativelyForReordering:animator:] ()
#9  0x000000010c5f85a0 in -[UICollectionView _performBatchUpdates:completion:invalidationContext:tentativelyForReordering:animator:] ()
#10 0x000000010c5f83c8 in -[UICollectionView _performBatchUpdates:completion:invalidationContext:tentativelyForReordering:] ()
#11 0x000000010c5f834a in -[UICollectionView _performBatchUpdates:completion:invalidationContext:] ()
#12 0x000000010c5f829f in -[UICollectionView performBatchUpdates:completion:] ()

```

And bizarre ones like

```
+[NSError _isCell]: unrecognized selector sent to class 0x1aa34
```

This PR prevents that crash and fixes the flakiness of test:

```
test_whenDeletingItemsTwice_withDataUpdatedTwice_thatAllUpdatesAppliedWithoutException
```

I originally fixed the crash by "staggering" duped delete index paths (see function in `IGListBatchUpdateData`), but that revealed another bug when reloads (which are converted to delete+insert) collide w/ actual deletes. In that case we want to drop the reload entirely. Fixing this required introducing another data structure, `IGListReloadIndexPath`.

Issue fixed: #651

Internal tasks t17306420, t17422443

### Checklist


CHANGELOG.md
```diff
@@ -58,6 +58,8 @@ This release closes the [3.0.0 milestone](https://github.com/Instagram/IGListKit
 
 - `IGListBatchUpdateData` replaced its `NSSet` properties with `NSArray` instead. [Ryan Nystrom](https://github.com/rnystrom) [(#616)](https://github.com/Instagram/IGListKit/pull/616)
 
+- `IGListUpdatingDelegate` now has a required method to handle reloading cells between index paths. [Ryan Nystrom](https://github.com/rnystrom) [(#657)](https://github.com/Instagram/IGListKit/pull/657)
```
jessesquires: let's note the method

Source/Common/IGListBatchUpdateData.mm
```diff
@@ -29,16 +29,45 @@ static void convertMoveToDeleteAndInsert(NSMutableSet<IGListMoveIndex *> *moves,
...
+                    indexPaths:(NSMutableArray<NSIndexPath *> *)indexPaths
                        deletes:(NSMutableIndexSet *)deletes
                        inserts:(NSMutableIndexSet *)inserts {
-    for (NSIndexPath *path in [indexPaths copy]) {
+    for (NSInteger i = indexPaths.count - 1; i >= 0; i--) {
+        NSIndexPath *path = indexPaths[i];
         const auto it = map.find(path.section);
         if (it != map.end() && it->second != nil) {
-            [indexPaths removeObject:path];
+            [indexPaths removeObjectAtIndex:i];
```
jessesquires: uh.. 🤔  is this mutating while enumerating?

Source/IGListUpdatingDelegate.h
```diff
@@ -105,6 +105,21 @@ typedef void (^IGListReloadUpdateBlock)();
...
+ @param fromIndexPath The source index path of the item to reload.
+ @param toIndexPath The destination index path of the item to reload.
+
+ @note Since UICollectionView is unable to handle calling -[UICollectionView reloadItemsAtIndexPaths:] safely while also
+ executing insert and delete operations in the same batch updates, the updater must know about the origin and
+ destination of the reload to perform a safe transition.
+ */
+- (void)reloadItemInCollectionView:(UICollectionView *)collectionView
+                     fromIndexPath:(NSIndexPath *)fromIndexPath
+                       toIndexPath:(NSIndexPath *)toIndexPath;
```
jessesquires: 🎉

Source/Internal/IGListReloadIndexPath.h
```diff
@@ -0,0 +1,47 @@
...
+ :nodoc:
+ */
+- (instancetype)init NS_UNAVAILABLE;
+
+/**
+ :nodoc:
+ */
++ (instancetype)new NS_UNAVAILABLE;
+
+@end
```
jessesquires: add nullability macros?

Source/Internal/IGListReloadIndexPath.h
```diff
@@ -0,0 +1,47 @@
...
+ * LICENSE file in the root directory of this source tree. An additional grant
+ * of patent rights can be found in the PATENTS file in the same directory.
+ */
+
+#import <Foundation/Foundation.h>
+
+#import <IGListKit/IGListMacros.h>
+
+IGLK_SUBCLASSING_RESTRICTED
+@interface IGListReloadIndexPath : NSObject
```
jessesquires: need docs for class

rnystrom:
Test failed b/c of timeout

Source/Common/IGListBatchUpdateData.mm
```diff
@@ -29,16 +29,45 @@ static void convertMoveToDeleteAndInsert(NSMutableSet<IGListMoveIndex *> *moves,
...
+    const NSInteger count = [dupes countForObject:path];
+    if (count == 0) {
+        [dupes addObject:path];
+        return path;
+    }
+
+    NSIndexPath *nextPath = [NSIndexPath indexPathForItem:path.item + count inSection:path.section];
+    NSIndexPath *staggeredPath = staggeredIndexPath(dupes, nextPath);
+    [dupes addObject:staggeredPath];
+    return staggeredPath;
```
amonshiz: this method is a bit confusing to read due to the fact that the return value isn't just the new index path (which happens to be the last one generated, i think) but also the implicit change to `dupes`. i'd prefer this to return a struct that returns the `NSCountedSet` and `NSIndexPath` as elements, even if that set is the same reference as what was passed in.

Tests/IGListBatchUpdateDataTests.m
```diff
@@ -123,14 +123,28 @@ - (void)test_whenMovingSections_withMoveFromConflictWithDelete_thatResultDropsTh
...
+- (void)test_whenDuplicateDeleteIndexPaths_thatDeletesStaggered {
+    IGListBatchUpdateData *result = [[IGListBatchUpdateData alloc] initWithInsertSections:indexSet(@[])
+                                                                           deleteSections:indexSet(@[])
+                                                                             moveSections:[NSSet new]
+                                                                         insertIndexPaths:@[]
+                                                                         deleteIndexPaths:@[newPath(2, 0), newPath(2, 0), newPath(2, 1), newPath(2, 4)]
+                                                                           moveIndexPaths:@[]];
+    XCTAssertEqual(result.deleteIndexPaths.count, 4);
+
+    NSArray *expected = @[newPath(2, 0), newPath(2, 1), newPath(2, 3), newPath(2, 4)];
```
amonshiz: while i understand this matches the logic, won't this end up causing the data source to delete the path (2,3) even though that wasn't an index path that should be deleted?

amonshiz:
I don't completely understand how this won't cause other items to be deleted within a section that has duplicate deletes for the same index path.

Source/Common/IGListBatchUpdateData.mm
rnystrom: Ya, but reversed the enum so mutations happen at the tail.

Source/Common/IGListBatchUpdateData.mm
rnystrom: 👍 will add some docs to this.

I think the struct will get really messy w/ the recursion tho

Tests/IGListBatchUpdateDataTests.m
rnystrom: Ya, but I _think_ that'll be ok b/c they were telling the datasource to delete {2,0} and {2,0} and updating their data source to reflect 2 deletions. So we **have** to delete 2 things, otherwise UICV will throw. And UICV will bug if we delete {2,0} 2x, so instead {2,0} {2,1} is the next-best thing.

Tho I think there might be a bug here, I think the expectation should be

```objc
NSArray *expected = @[newPath(2, 0), newPath(2, 1), newPath(2, 2), newPath(2, 4)];
```

rnystrom: Made me double-check how this works. There's a bug that if you delete the _last_ item twice, it will throw b/c its deleting an item that doesn't exist...

rnystrom:
Ok backing out my "stagger" idea b/c if you delete the **last** item twice you will delete something that doesn't exist.

Maybe we could change the API so it validates that deletes are within the range of `[collectionView numberOfItemsInSection:section]`?

Changing deletes back to `NSSet` leaves us open to data inconsistency crashes...

Tests/IGListBatchUpdateDataTests.m
amonshiz: What about (2,2)? Is it going to be deleted? Won't that be an internal inconsistency since it really wasn't deleted and there is no corresponding insert to ensure that the correct number of items is in the end?

rnystrom:
Eating my words, I want to fix this for real. Going to need to get clever. Dang u collection view.

jessesquires:
could we just bounds-check the "double delete" case and ignore the second delete?

rnystrom:
@jessesquires If we drop one of the deletes but a data source is updated assuming 2 deletes are applied, UICV will throw for inconsistency.

rnystrom:
Ok stopgap solution for now: 

- Restore the delete test with a FIXME prefix so it doesn't run
- File an issue that this is a known problem
- Land this PR with `NSSet` to dedupe delete operations

We've survived shipping IGListKit on Instagram for over a year w/out this, so we're in no hurry to get this totally working. However w/out a fix this _is_ crashing now and causing tests to fail. Let's get this cleaned up and working for now.

rnystrom:
Importing to get internal CI running on this, not landing

Issue #656, opened 2017-04-17
Fix Storyboard example app


rnystrom:
Getting a crash in the examples:

```
2017-04-17 12:08:52.987 IGListKitExamples[78921:735444] Failed to set (minimumLineSpacing) user defined inspected property on (IGListCollectionViewLayout): [<IGListCollectionViewLayout 0x7fec4d6338f0> setValue:forUndefinedKey:]: this class is not key value coding-compliant for the key minimumLineSpacing.
```

PR #658, opened 2017-04-17
[#656] Fix storyboard example app, Fix example build issue

Sherlouk:
Collection view layout has "user defined runtime attributes" which were left over from the old IG grid layout. I've removed them as we don't use them any more.

Also I couldn't build the project because of the map function in the dependency injection example, so changed that slightly to make it run (tested that and still works)

Issue fixed: #656


jessesquires:
nice! thanks @Sherlouk !

Issue #609, opened 2017-04-02
RFC: Add section property to IGListSectionController


amonshiz:
`IGListKit` should remove the `-[IGListCollectionContext sectionForSectionController:]` method from `IGListCollectionContext` and replace it with a `section` property on `IGListSectionController`.

### Reasons
- `sectionForSectionController:` has nothing to do with the collection displaying the section/list. This method alone requires that the `IGListAdapter` be the `collectionContext` object.
- A section controller's actual section index cannot change without the `IGListAdapter` knowing, and at that point it would be a new section controller anyhow.
- The property allows the `collectionContext` object to really be anything and not something that owns a mapping from section controller to an index.

The reasons that this method was originally added are all solved by adding a `section` property and as mentioned, that property is known at creation time.

### Options
- Add this as a property to the `init` method of `IGListSectionController`
- Add this as `@property (nonatomic, readwrite, assign) NSInteger index;`

The `init` option is great because it means the property can be `readonly`, but it does mean that all subclasses have to be careful to include the property also.

The `readwrite` property is nice because then the `IGListKit` infra can own setting it. The downside is everyone knows that property can be written to. Maybe this is declared in `IGListSectionController.h` as `readonly` and in `IGListSectionControllerInternal.h` it is redeclared as `readwrite`?

### Context
Speaking with @rnystrom in person about ways to break up implicit dependencies on `UICollectionView` and this method stood out as one that could be culled and would easily break up a hidden dependency on `IGListAdapter`.

jessesquires:
Makes sense to me.

> Maybe this is declared in IGListSectionController.h as readonly and in IGListSectionControllerInternal.h it is redeclared as readwrite?

Yep. This is what we already do for `.isFirstSection` and `.isLastSection`

jessesquires:
Let's do this in 3.0

rnystrom:
On board w/ this, ++ 3.0

amonshiz:
Will do. Finally back from a pretty crazy week. Gonna start this week and should have PR pretty quickly.

Adlai-Holler:
How would this handle updating the index of all section controllers, if I inserted a section at index 0?

jessesquires:
^^ I assumed that's always what this was about, just a convenience wrapper

amonshiz:
[...]

PR #671, opened 2017-04-18
Move section index to property

amonshiz:
This pull request removes the `sectionForSectionController:` method from the `IGListCollectionContext` protocol so that the protocol is exclusively for presentation methods.

This should not add new functionality, but rather makes the index directly accessible on the section controllers themselves. This change makes sense because at no time will there be an update to the list that the list adapter is unaware of and so it will always be able to set and update any indexes for a section controller that has changed.

Issue fixed: #609

### Checklist

- [X] I added an entry to the `CHANGELOG.md` for any breaking changes, enhancements, or bug fixes.


Tests/IGListAdapterE2ETests.m
```diff
@@ -1477,25 +1477,25 @@ - (void)test_whenInsertingItemsTwice_withDataUpdatedTwice_thatAllUpdatesAppliedW
...
+//        object.value = @2;
+//        [batchContext deleteInSectionController:sectionController atIndexes:[NSIndexSet indexSetWithIndex:0]];
+//        [batchContext deleteInSectionController:sectionController atIndexes:[NSIndexSet indexSetWithIndex:0]];
+//    } completion:^(BOOL finished2) {
+//        XCTAssertEqual([self.collectionView numberOfSections], 1);
+//        XCTAssertEqual([self.collectionView numberOfItemsInSection:0], 2);
+//        [expectation fulfill];
+//    }];
+//    [self waitForExpectationsWithTimeout:15 handler:nil];
+//}
```
jessesquires: 🤔

Tests/IGListStackSectionControllerTests.m
```diff
@@ -246,12 +246,12 @@ - (void)test_whenQueryingSectionControllerSection_thatSectionMatchesStackSection
...
-    XCTAssertEqual([section11.collectionContext sectionForSectionController:section11], 0);
-    XCTAssertEqual([section12.collectionContext sectionForSectionController:section12], 0);
-    XCTAssertEqual([section21.collectionContext sectionForSectionController:section21], 1);
-    XCTAssertEqual([section22.collectionContext sectionForSectionController:section22], 1);
+    XCTAssertEqual(stack1.sectionIndex, 0);
+    XCTAssertEqual(stack2.sectionIndex, 1);
+    XCTAssertEqual(section11.sectionIndex, 0);
+    XCTAssertEqual(section12.sectionIndex, 0);
+    XCTAssertEqual(section21.sectionIndex, 1);
+    XCTAssertEqual(section22.sectionIndex, 1);
```
jessesquires: add more complex cases here?

Tests/IGListAdapterTests.m
```diff
@@ -1133,4 +1133,32 @@ - (void)test_whenQueryingInsetContainerSize_thatResultIsBoundsInsetByContent {
...
+
+    IGListSectionController<IGListSectionType> *controller1b = [self.adapter sectionControllerForObject:@1];
+    XCTAssertEqual(controller1b.sectionIndex, 1);
+    XCTAssertFalse(controller1b.isFirstSection);
+
+    XCTAssertEqual([self.adapter sectionControllerForObject:@2].sectionIndex, 2);
+    XCTAssertEqual([self.adapter sectionControllerForObject:@3].sectionIndex, 3);
+    XCTAssertTrue([self.adapter sectionControllerForObject:@3].isLastSection);
+}
+
```
jessesquires: we should probably test: (1) append, (2) insert in middle

jessesquires:
`sectionIndex` will always be correct, right? even after ListAdapter mutations?

jessesquires:
let's make sure we handle the cases that @Adlai-Holler mentioned in the issue

amonshiz:
All the existing tests pass, so at least we have those mutations covered. Would it be worth adding a list adapter mutations specific testing file?

This one test was to cover the case mentioned in the task. However, speaking with Ryan it seems this *shouldn't* be passing currently so I will need to debug a bit.

Tests/IGListStackSectionControllerTests.m
amonshiz: My preference would just be to 86 the stack controller ...

Tests/IGListAdapterE2ETests.m
amonshiz: science

Source/Internal/IGListSectionMap.h
```diff
@@ -38,7 +38,7 @@ IGLK_SUBCLASSING_RESTRICTED
  @param objects The objects in the collection.
  @param sectionControllers The section controllers that map to each object.
  */
-- (void)updateWithObjects:(NSArray <id <NSObject>> *)objects sectionControllers:(NSArray <id <NSObject>> *)sectionControllers;
+- (void)updateWithObjects:(NSArray <id <NSObject>> *)objects sectionControllers:(NSArray <IGListSectionController *> *)sectionControllers;
```
amonshiz: at some point this method was not updated to the `IGListSectionController<IGListSectionType> *` type and was missed in a few refactorings. the next method down, `sectionControllerForSection:` returned an `IGListSectionController *` so the types were a significant mismatch and needed to be updated.

Tests/IGListSectionMapTests.m
```diff
@@ -23,39 +23,39 @@ @implementation IGListSectionMapTests
 
 - (void)test_whenUpdatingItems_thatArraysAreEqual {
     NSArray *objects = @[@0, @1, @2];
-    NSArray *sectionControllers = @[@"a", @"b", @"c"];
+    NSArray *sectionControllers = @[[IGListTestSection new], [IGListTestSection new], [IGListTestSection new]];
```
amonshiz: all of these are just so that the test actually respects the types that `IGListSectionMap` require.

jessesquires:
@amonshiz -- looks like this is good to go, except for resolving conflicts?

amonshiz:
will do!

PR #674, opened 2017-04-19
Fix empty cells on fast scroll

madmag77:
Added protection from this kind of bug in _UICollectionView_ - [Bug description on stackoverflow](http://stackoverflow.com/a/13389461/377384). In certain conditions (when section insets, lineSpacing and interItemSpacing are set) when one scrolls fast then blocks of cells are disappearing from _CollectionView_. In my case almost 40 cells were dissapeared.
It happens because in _layoutAttributesForItemAtIndexPath_ enumeration of cells stops on first cell out that is of _rect_, but on above conditions _CollectionView_ provide us with duplicate cell attribute that is out of _rect_, so at this moment function stops enumerating and returns resulting array with not all cells.
I added check that if exactly two consecutive elements are out of rect - only in this case algorithm stops enumeration. It fixes above bug.

Can't add a test for this changes without huge rewriting of _layoutAttributesForElementsInRect_ because attributes comes from intrinsic function - _layoutAttributesForItemAtIndexPath_ and I can't change them.

Issue fixed: #

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -74,6 +74,8 @@ This release closes the [3.0.0 milestone](https://github.com/Instagram/IGListKit
 
 ### Fixes
 
+- Added protection from this kind of bug in `UICollectionView` - [Bug description](http://stackoverflow.com/a/13389461/377384). In certain conditions: when section insets, lineSpacing and interItemSpacing are set, when scroll fast blocks of cells are disappearing from collectionView. I added check that exactly two consecutive elements are out of rect - only in this case enumeration of attributes in `layoutAttributesForItemAtIndexPath` will be stopped.
+
 - Gracefully handle a `nil` section controller returned by an `IGListAdapterDataSource`. [Ryan Nystrom](https://github.com/rnystrom) [(#488)](https://github.com/Instagram/IGListKit/pull/488)
 
 - Fix bug where emptyView's hidden status is not updated after the number of items is changed with `insertInSectionController:atIndexes:` or related methods. [Peter Edmonston](https://github.com/edmonston) [(#395)](https://github.com/Instagram/IGListKit/pull/395)
--- a/Source/IGListCollectionViewLayout.mm
+++ b/Source/IGListCollectionViewLayout.mm
@@ -126,7 +126,12 @@ static void adjustZIndexForAttributes(UICollectionViewLayoutAttributes *attribut
     // since we iterate through sections ascending, we may hit a section who is only partially visible
     // in that case we short circuit building layout attributes
     BOOL remainingCellsOutOfRect = NO;
-
+    
+    // sometimes, when scrolling fast in UICollectionView appears bug with duplicated element attributes
+    // one of wich is out of rect, so we shouldn't immidiatelly return once we found only one cell
+    // that is out of rect - wait second one to make sure we are finish with cells in rect
+    BOOL secondCellOutOfRectDetected = NO;
+    
     const NSRange range = [self rangeOfSectionsInRect:rect];
     if (range.location == NSNotFound) {
         return nil;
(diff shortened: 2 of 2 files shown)


rnystrom:
@madmag77 I'm a little confused, the SO link says that the bug is inside `UICollectionViewFlowLayout`, right? Our layout just subclasses `UICollectionViewLayout`.

If I'm reading your fix correctly, `UICollectionView` will ask for cells that aren't in ascending order [which we assume](https://github.com/Instagram/IGListKit/blob/master/Source/IGListCollectionViewLayout.mm#L126-L128). Right?

Do you have an example project that reproduces this? I've never seen it before.

alexcurylo:
@rnystrom That's just the closest thing we could find on SO to this problem, so apparently it's more generic than the link suggests. The example project that reproduces this is up to 5.6 RC2 of [the Agoda app](https://itunes.apple.com/us/app/agoda-hotel-accommodation-booking-deals/id440676901?mt=8) since I decided to adopt `IGListKit` to Swiftify the search funnel with and getting it to work with the non-Swiftied cell and header classes was a trifle problematic. Off to review now so it's all good, but this looked like a fix worth sending upstream since our QA was pretty upset with great big stretches of empty screen and all. It didn't show up until after we'd been playing with insets to make it look _just_ as if nothing changed, so it's definitely a hard case to hit. We can try to construct a minimal sample project if you like, or would a screen video of the RL problem be enough?

rnystrom:
@alexcurylo a sample would be best, that way we can inspect the internals of what UICV is doing and

- Confirm this is the right fix
- Possibly file a radar w/ Apple

Sorry for the trouble! We haven't run into a report of this before, very bizarre!

madmag77:
I changed stock Demo project in IGListKit here: https://github.com/madmag77/IGListKit/tree/demo-with-emty-cells - just run on Simulator iPad Pro 12.9, choose Display delegate and fast scroll and you will notice empty spaces as on screenshot in Issue
Added new Issue about this bug: https://github.com/Instagram/IGListKit/issues/680

rnystrom:
Ok I took a spin on the example and figured out what's wrong. This actually doesn't have anything to do w/ scrolling fast, its just a bug in our layout!

The problem shows up when the first item in the section intersects the top of the rect b/c its height is tall enough, but the second element **in the same row** is **too short** to intersect w/ the rect.

Take a look at "item 1" in this drawing:

img_3312

I wrote a unit test for `IGListCollectionViewLayoutTests` that fails b/c of this:

```objc
- (void)test_whenSecondItemDoesntIntersectRect_thatOtherAttributesExist {
    [self setUpWithStickyHeaders:NO topInset:0];
    NSMutableArray *data = [NSMutableArray new];
    for (NSInteger i = 0; i < 6; i++) {
        [data addObject:[[IGLayoutTestSection alloc] initWithInsets:UIEdgeInsetsZero
                                                        lineSpacing:0
                                                   interitemSpacing:0
                                                       headerHeight:0
                                                              items:@[
                                                                      [[IGLayoutTestItem alloc] initWithSize:(CGSize){50, 100}],
                                                                      [[IGLayoutTestItem alloc] initWithSize:(CGSize){50, 10}],
                                                                      ]]];
    }
    [self prepareWithData:data];

    NSArray *attributes = [self.layout layoutAttributesForElementsInRect:CGRectMake(0, 50, 100, 100)];
    NSArray *paths = [[attributes valueForKeyPath:@"indexPath"] sortedArrayUsingSelector:@selector(compare:)];
    NSArray *expectation = @[
                             genIndexPath(0, 0),
                             genIndexPath(1, 0),
                             genIndexPath(1, 1),
                             ];

    // should include 2 of the 100-height items and one of the 10-height
    XCTAssertEqualObjects(paths, expectation);
}
```

I actually have a feeling we can just remove [remainingCellsOutOfRect](https://github.com/Instagram/IGListKit/blob/master/Source/IGListCollectionViewLayout.mm#L128) entirely, and only add cells that intersect.

Want to give this a shot? Would be a great fix!

alexcurylo:
@rnystrom That sounds totally plausible yep, where I said

> getting it to work with the non-Swiftied cell and header classes was a trifle problematic.

that was in fact because our main search results have hotel cells of varying heights that can look just like your drawing there, and the currently shipping pod kinda needs a fixed size grid layout, that's why we picked the _almost_ working version from `master`. We're on Bangkok time here so won't be in the office for a few hours yet but we'll try and make sure we get this checked out today, thanks!

madmag77:
@rnystrom thanks a lot! You are completely right, this bug is not a UICollectionView fault.
I made new PR - [PR-684](https://github.com/Instagram/IGListKit/pull/684) where remove optimization flag, added your test and one more test for situation when 2 consecutive cells don't intersect with the Rect (because my previous fix doesn't brake your test). 
Should I close this PR?

rnystrom:
Closing in favor of #684

Issue #675, opened 2017-04-19
Clean up documentation after IGListSectionType is removed


rnystrom:
Followup once #168 lands internally: do a pass through documentation and make sure nothing references `IGListSectionType`.

PR #676, opened 2017-04-19
Remove refs to IGListSectionType, fix TV app example

rnystrom:
- Clean up docs
  - Remove `IGListSectionType` and `IGListCollectionView`
- Fix TV app example

#trivial

Issue fixed: #675

### Checklist

### DIFF
--- a/Guides/Best Practices and FAQ.md
+++ b/Guides/Best Practices and FAQ.md
@@ -4,7 +4,7 @@ This guide provides notes and details on best practices in using `IGListKit`, ge
 
 ## Best Practices
 
-- We recommend adding an assert to check [`-isKindOfClass:`](https://developer.apple.com/reference/objectivec/1418956-nsobject/1418511-iskindofclass) on the object you receive in [`-didUpdateToObject:`](https://instagram.github.io/IGListKit/Protocols/IGListSectionType.html#/c:objc(pl)IGListSectionType(im)didUpdateToObject:) in your section controllers. 
+- We recommend adding an assert to check [`-isKindOfClass:`](https://developer.apple.com/reference/objectivec/1418956-nsobject/1418511-iskindofclass) on the object you receive in [`-didUpdateToObject:`](https://github.com/Instagram/IGListKit/blob/master/Source/IGListSectionController.h#L63-L72) in your section controllers. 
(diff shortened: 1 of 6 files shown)


Guides/Best Practices and FAQ.md
```diff
@@ -4,7 +4,7 @@ This guide provides notes and details on best practices in using `IGListKit`, ge
 
 ## Best Practices
 
-- We recommend adding an assert to check [`-isKindOfClass:`](https://developer.apple.com/reference/objectivec/1418956-nsobject/1418511-iskindofclass) on the object you receive in [`-didUpdateToObject:`](https://instagram.github.io/IGListKit/Protocols/IGListSectionType.html#/c:objc(pl)IGListSectionType(im)didUpdateToObject:) in your section controllers. 
+- We recommend adding an assert to check [`-isKindOfClass:`](https://developer.apple.com/reference/objectivec/1418956-nsobject/1418511-iskindofclass) on the object you receive in [`-didUpdateToObject:`](https://github.com/Instagram/IGListKit/blob/master/Source/IGListSectionController.h#L63-L72) in your section controllers.
```
jessesquires: once 3.0 is ready, let's re-gen docs and re-link to the docs page

Guides/Getting Started.md
```diff
@@ -8,23 +8,29 @@ After installing `IGListKit`, creating a new list is easy.
 
 ### Creating a section controller
 
-Creating a new section controller is simple. You subclass `IGListSectionController` and conform to the `IGListSectionType` protocol. Once you conform to `IGListSectionType`, the compiler will make sure you implement all of the required methods.
+Creating a new section controller is simple. All you have to do is subclass `IGListSectionController` and override at least `cellForItemAtIndex:` and `sizeForItemAtIndex:`.
```
jessesquires: nit: remove "all you have to do", just "Sublcass..."

also -- we should list `didUpdateToObject:`. and **i really think** this method should assert.

Sherlouk: Not 100% sure I agree with that Jesse, whilst the majority will override `didUpdateToObject:` in the case of decoration views such as adverts or loading views that don't have an object necessarily bound to them I don't see the reason to have the method overridden

PR #677, opened 2017-04-19
Invalidate layout when collection view reused between IGListAdapters

### LINKED ISSUE #659: Invalidate layout when collection view reused between IGListAdapters
Discovered by @yifanpeng in a nested-section controller. Setting a new collection view wont necessarily invalidate the layout, even if the data is reloaded. Looks like this requires an `invalidateLayout` call.

jessesquires:
Closes #659


jessesquires:
@rnystrom  -- hm... probably just want to mock and verify that `invalidateLayout` is called? or... ?

PR #642, opened 2017-04-13
Integrating Danger-SwiftLint #trivial

heshamsalman:
## Changes in this pull request

Integrates SwiftLint into DangerBot. 
You may want to append a `?w=1` to the end of the files-changed url to exclude whitespace-only changes. 

Should be good to go -- failing due to no changes to `CHANGELOG.md`. I don't believe white space is enough reason to add to the change log. 

### Checklist

- [x] All tests pass. Demo project builds and runs.
- [x] I added tests, an experiment, or detailed why my change isn't tested.
- [ ] I added an entry to the `CHANGELOG.md` for any breaking changes, enhancements, or bug fixes.
- [x] I have reviewed the [contributing guide](https://github.com/Instagram/IGListKit/blob/master/.github/CONTRIBUTING.md)

Issue #394, opened 2017-01-07
Automate PR checking with Danger


rnystrom:
http://danger.systems/ cc @orta

### Warnings

- No CHANGELOG.md entry for any change
- Non-final class in examples
- Big PR
- [Swiftlint](https://github.com/ashfurrow/danger-swiftlint) for examples
- Markdown verification using [prose](https://github.com/dbgrandi/danger-prose) (incl. link checking)
- ObjC classes without `IG_SUBCLASSING_RESTRICTED`

### Failures

- New CHANGELOG.md entry when `.h` file is changed
- Facebook copyright in headers
- Only allow editing `/docs`from admins
- WIP pull requests
- Changes to restricted file types (`.lock`, `.xcconfig`, etc)
- **Always fail**: "Merging should be done internally"

Any other ideas?

Sherlouk:
Failure on changes to Pod files (`.lock`, `.xcconfig`)

Source https://github.com/Instagram/IGListKit/pull/368#issuecomment-269092317

jessesquires:
Danger plugins:

- Proselint: https://github.com/dbgrandi/danger-prose
- Swiftlint: https://github.com/ashfurrow/danger-swiftlint

Sherlouk:
Could probably also do with a changelog format validator.
`{summary} {Name} (#{PR Number})`

Should be easy to at least validate the correct PR and PR link

orta:
There's [a semi-standard](http://keepachangelog.com/en/0.3.0/) for CHANGELOGs too - this plugin implements it - http://danger.systems/plugins/changelog.html

looks like https://github.com/ruby-grape/grape/blob/master/CHANGELOG.md

Sherlouk:
As discussed in #399, we should **always** fail with a message similar to "Merging should be done internally" to prevent the merge button from ever being usable.

@rnystrom Should I go ahead and update the original comment with the various rules (as discussed above in thread) so we can tick them off once done?

rnystrom:
@Sherlouk go for it!

Some linter ideas:

- Error on Swift classes in the examples marked `final`
- Warn on ObjC classes w/out using `IG_SUBCLASSING_RESTRICTED`

Sherlouk:
@rnystrom Assume you meant classes not marked `final` for Swift?

Sherlouk:
@rnystrom Worth keeping this issue open as there are still unimplemented rules?

rnystrom:
@Sherlouk good call!

[...]

PR #681, opened 2017-04-20
Add Swiftlint #trivial

heshamsalman:
Replaces #642

Issue ref: #394 

Integrates SwiftLint into DangerBot.
You may want to append a ?w=1 to the end of the files-changed url to exclude whitespace-only changes.

#trivial


jessesquires:
@rnystrom -- still having trouble getting this imported. any updates?

rnystrom:
Ok so this is adding about 35mb to the repo from including the SwiftLint framework 3x from each of the examples. Wow. Our internal tools flagged this and wont let us import it b/c of the size.

I'm not totally sure what the best course of action here should be. Obviously the main issue is having 3x separate example projects, each with their own `Podfile`, and the fact that we check in all Pod data.

Couple options:

- We drop all the SwiftLint stuff and just keep the lint changes to .swift files
- We update all of the examples to be a single project/workspace so we only SwiftLint import once
- We drop this altogether ☹️

heshamsalman:
@rnystrom surprised it's 35mb. 
I think short-term, the right move may be to take the changes to the `.swift` files and drop SwiftLint from the example projects -- it won't flag errors during development of the Example project, but the bot should still flag errors on the PR. 

Restructuring into one `xcworkspace` sounds like the right longer-term move to me.

Sherlouk:
Another option is to get the framework (1 version of it) store that in a directory (e.g. Examples/Swiftlint) and then reference that with a built script from all 3 projects. (So stop using CocoaPods to get Swiftlint).

I'd honestly prefer this!
@rnystrom  @heshamsalman

jessesquires:
I think @Sherlouk 's idea is good.

That's how the local `IGListKit` pod is working -- all examples just reference `../../IGListKit`

heshamsalman:
Occam's Razor 😎
Seems like the right move to me. I'll make the change late Friday or Sunday 👍

jessesquires:
cc @rnystrom ?

rnystrom:
On board w/ @Sherlouk's plan, sounds great!

heshamsalman:
@Sherlouk @jessesquires @rnystrom 

Should be a good deal smaller now. 👍

Sherlouk:
Changes look a lot nicer, will leave it to the insta crew to check on Phabricator! Hopefully we can get it in 🙏

Sherlouk:
@rnystrom @jessesquires Think we could try and see if Phabricator will except this some time this week? Would be nice to get this in (and I get that it's a pain to keep rebasing 😅)

jessesquires:
@Sherlouk - yep, everything looks alright to me!

heshamsalman:
Ayyy 👍

jessesquires:
Ok, bad news:

So imported succeed but land failed internally because the binary files are too large. 😞 

The best path forward is probably to run this on CI, but *not* include `SwiftLint` in the repo? Not 100% on what the best option is. 

@rnystrom is out for a bit, when he returns we can sort this out. 👍 

Sorry for the churn @heshamsalman ! 😅

heshamsalman:
@jessesquires I could alternatively change the SwiftLint script to rely on a user's `HomeBrew` installed copy of SwiftLint (instead of a locally referenced binary), and echo installation instructions if they don't have it installed?

jessesquires:
@heshamsalman -- that sounds awesome! Let's **definitely** do that.

That's how we handle `jazzy`. Use must have it installed.

jessesquires:
@heshamsalman -- because of how the importing system works, we'll need to open a new PR 😄

heshamsalman:
@jessesquires haha classic. I'll re-open soon

jessesquires:
thanks so much @heshamsalman ! 🎉

BasThomas:
> User must have it installed. (Though we don't provide a helpful message about how to install. that would be nice)

Might be a good idea to create an issue for that? :)

jessesquires:
@BasThomas Yep! 😊

BasThomas:
Created #744. :)

Issue #414, opened 2017-01-13
Remove workingRangeSize from initializer


rnystrom:
Talked w/ @ocrickard, probably default disabled and allow setting the size (which would internally create and update the WR state) after initing the `IGListAdapter`.

Incredibly useful feature, but:

- Other apps have their own implementation
- We should investigate how well the `UICollectionView` prefetch API works
- Since this is a more advanced feature, most implementations don't need to think about it at init time

Options:

- 2 initializers (`initWithUpdater:viewController:` and `initWithUpdater:viewController:workingRangeSize:`
  - Makes transition easier
  - Maybe add `__attribute__ ((deprecated))` to the old init?
- `@property NSUInteger workingRangeSize`
  - Have to worry about setting after feed created and used. Means we have to find what the range is and what's in it so once scrolling starts we're g2g.

BasThomas:
Which of the two options would you prefer and why?

rnystrom:
@BasThomas let's do the first! I agree, easier for everyone.

jessesquires:
@BasThomas if you want to / have time to do this within the next week or so, we can include in the 3.0 release 😄

BasThomas:
So I was looking at this - do we want to deprecate and remove the option to provide a working range size completely? Or should we still provide the init for advanced usage?

rnystrom:
@BasThomas let's leave the current init as the designated initializer still, but provide an initializer that internally does `workingRangeSize: 0`.

Also would you mind double-checking what work [IGListWorkingRangeHandler](https://github.com/Instagram/IGListKit/blob/master/Source/IGListAdapter.m#L47) does when the size is 0? I think it basically no-ops, but we _might_ want to wrap [initializing the handler](https://github.com/Instagram/IGListKit/blob/master/Source/IGListAdapter.m#L47) with something like:

```objc
if (workingRangeSize > 0) {
  _workingRangeHandler = // ...
}
```

Might want to sync w/ @ocrickard really quick too to see if that's the best course of action.

BasThomas:
When we do a no-op (not initializing the `workingRangeHandler`), `test_whenDisplayingItemAtPath_withWorkingRangeSizeZero_thatItemEntersWorkingRange` and `test_whenDisplayingItemAtPath_withWorkingRangeSizeZero_thenHidingThatItem_thatItemLeavesWorkingRange` fail, as they seem to rely / test the rangeHandlers without them being present.

rnystrom:
[...]

PR #686, opened 2017-04-21
Init without working range size

BasThomas:

Issue fixed: #414 

### Checklist

What should we do with testing? I changed all the occurrences of the old initializer with the new one, so I think this is good to go?


BasThomas:
@ocrickard, @rnystrom thought you might be interested in looking at this as well.

Source/IGListAdapter.h
```diff
@@ -103,6 +103,17 @@ IGLK_SUBCLASSING_RESTRICTED
                workingRangeSize:(NSInteger)workingRangeSize NS_DESIGNATED_INITIALIZER;
 
 /**
+ Initializes a new `IGListAdapter` object.
```
rnystrom: Let's add that this uses workingRangeSize 0

BasThomas: Also changed the opt-out to prefer the new init over providing `0` as a value.

Source/IGListAdapter.h
```diff
@@ -96,13 +96,26 @@ IGLK_SUBCLASSING_RESTRICTED
  the previous and succeeding 2 objects will be notified that they are within the working range. As you scroll the list
  the range is updated as objects enter and exit the working range.
 
- To opt out of using the working range, you can provide a value of `0`.
+ To opt out of using the working range, use `initWithUpdater:viewController` instead.
```
rnystrom: Suuuuuper nit: can this be

> To opt out of using the working range, use `initWithUpdater:viewController:` or provide a working range of `0`.

BasThomas: Of course!

Issue #714, opened 2017-05-03
sizeForSupplementaryViewOfKind is called always for index=0


gobetti:
## New issue checklist

- [x] I have reviewed the [`README`](https://github.com/Instagram/IGListKit/blob/master/README.md) and [documentation](http://instagram.github.io/IGListKit)
- [x] I have searched [existing issues](https://github.com/Instagram/IGListKit/issues) and this is not a duplicate

### General information

- `IGListKit` version: 2.1.0
- iOS version(s): Any
- CocoaPods version: 1.2.1
- Xcode version: 8.3.2
- Devices/Simulators affected: Any
- Reproducible in the demo project? (Yes/No): Yes
- Related issues: -

No matter which is the actual section index, the delegate method `-[IGListSupplementaryViewSource sizeForSupplementaryViewOfKind:atIndex:]` is always called for an index=0. The bug comes from the fact that an `indexPath` is calculated by `-[UICollectionViewDelegateFlowLayout collectionView:layout:referenceSizeForHeaderInSection:]` and `-[UICollectionViewDelegateFlowLayout collectionView:layout:referenceSizeForFooterInSection:]` like this:
`NSIndexPath *indexPath = [NSIndexPath indexPathForItem:0 inSection:section];`
however in the defective method mentioned above, the `index` that is passed is extracted from the `NSIndexPath` like this:
`const CGSize size = [supplementaryViewSource sizeForSupplementaryViewOfKind:elementKind atIndex:indexPath.item];`
(notice that the `item` is taken instead of the `section`, thus always resulting 0)

jessesquires:
Thanks @gobetti !

rnystrom:
@gobetti thanks for looking into this! I'm not sure I understand what the bug is though. Supplementary views are meant to be used with index paths when used beyond `UICollectionViewFlowLayout`. Notice how the [UICollectionViewLayout supplementary attributes API](https://developer.apple.com/reference/uikit/uicollectionviewlayout/1617792-layoutattributesforsupplementary?language=objc) asks for an index path (meaning you can have more than one, thus needing the `item`).

We designed the `IGListSupplementaryViewSource` API with this in mind.

Do you have an example of how this breaks? With the change in #715, implementations with more than one supplementary view will break (b/c they receive the section and not the index of the view).

gobetti:
[...]

PR #715, opened 2017-05-03
Fixes sizeForSupplementaryViewOfKind always at index=0

gobetti:
An `NSIndexPath` is being built from IGListAdapter+UICollectionView.m
before being passed to
`-[IGListSupplementaryViewSource sizeForSupplementaryViewOfKind:atIndex:]`
however this later is taking the `item` instead of the `section` -
resulting in the index always being equal to 0, even for section
indexes > 0.

Issue fixed: #714

### Checklist

### DIFF
--- a/Tests/IGListAdapterTests.m
+++ b/Tests/IGListAdapterTests.m
@@ -1067,6 +1067,24 @@
     XCTAssertEqual(size.height, 0.0);
 }
 
+- (void)test_whenSectionIndexGreaterThanZero_thatAdapterReturnsSizeForSectionIndexesGreaterThanZero {
+  self.dataSource.objects = @[@1];
+  [self.adapter reloadDataWithCompletion:nil];
+  
+  IGTestSupplementarySource *supplementarySource = [IGTestSupplementarySource new];
+  supplementarySource.collectionContext = self.adapter;
+  supplementarySource.supportedElementKinds = @[UICollectionElementKindSectionFooter];
+  
+  IGListSectionController *controller = [self.adapter sectionControllerForObject:@1];
+  controller.supplementaryViewSource = supplementarySource;
+  supplementarySource.sectionController = controller;
+  
+  const CGSize size = [self.adapter sizeForSupplementaryViewOfKind:UICollectionElementKindSectionHeader
+                                                       atIndexPath:[NSIndexPath indexPathForItem:0 inSection:1]];
+  XCTAssertEqual(size.width, supplementarySource.sizeForSectionIndexesGreaterThanZero.width);
+  XCTAssertEqual(size.height, supplementarySource.sizeForSectionIndexesGreaterThanZero.height);
+}
+
 - (void)test_whenQueryingContainerInset_thatMatchesCollectionView {
     self.dataSource.objects = @[@2];
     [self.adapter reloadDataWithCompletion:nil];
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -114,6 +114,8 @@ This release closes the [3.0.0 milestone](https://github.com/Instagram/IGListKit
 
 - When setting the collection view on `IGListAdapter`, its layout is now properly invalidated. [Jesse Squires](https://github.com/jessesquires) [(#677)](https://github.com/Instagram/IGListKit/pull/677)
 
+- `IGListSupplementaryViewSource` is now able to receive section indexes different than 0.
+
 2.1.0
 -----
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -114,7 +114,7 @@ This release closes the [3.0.0 milestone](https://github.com/Instagram/IGListKit
 
 - When setting the collection view on `IGListAdapter`, its layout is now properly invalidated. [Jesse Squires](https://github.com/jessesquires) [(#677)](https://github.com/Instagram/IGListKit/pull/677)
 
-- `IGListSupplementaryViewSource` is now able to receive section indexes different than 0.
+- Fixed a bug in `IGListSupplementaryViewSource` where it would always have a section index of 0. [Marcelo Gobetti](https://github.com/gobetti) [(#715)](https://github.com/Instagram/IGListKit/pull/715)
 
 2.1.0
 -----
--- a/Source/IGListAdapter.m
+++ b/Source/IGListAdapter.m
@@ -474,7 +474,7 @@
     IGAssertMainThread();
     id <IGListSupplementaryViewSource> supplementaryViewSource = [self supplementaryViewSourceAtIndexPath:indexPath];
     if ([[supplementaryViewSource supportedElementKinds] containsObject:elementKind]) {
-        const CGSize size = [supplementaryViewSource sizeForSupplementaryViewOfKind:elementKind atIndex:indexPath.item];
+        const CGSize size = [supplementaryViewSource sizeForSupplementaryViewOfKind:elementKind atIndex:indexPath.section];
         return CGSizeMake(MAX(size.width, 0.0), MAX(size.height, 0.0));
     }
     return CGSizeZero;
--- a/Tests/Objects/IGTestSupplementarySource.h
+++ b/Tests/Objects/IGTestSupplementarySource.h
@@ -17,6 +17,8 @@
 
 @property (nonatomic, assign) CGSize size;
 
+@property (nonatomic, assign) CGSize sizeForSectionIndexesGreaterThanZero;
+
 @property (nonatomic, strong, readwrite) NSArray<NSString *> *supportedElementKinds;
 
 @property (nonatomic, weak) id<IGListCollectionContext> collectionContext;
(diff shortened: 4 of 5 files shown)


MarvinNazari:
@gobetti, @rnystrom: This actually fixes the crash Im having.

rnystrom:
@MarvinNazari what is the crash?

MarvinNazari:
@rnystrom crash happens when used with ASDK
basically happens when i insert a new section that have `UICollectionElementKindSectionHeader` and update the adapter.
```swift
Terminating app due to uncaught exception 'NSInternalInconsistencyException', reason: 'no UICollectionViewLayoutAttributes instance for -layoutAttributesForSupplementaryElementOfKind: UICollectionElementKindSectionHeader at path <NSIndexPath: 0xc000000000000016> {length = 2, path = 0 - 0}'
```

Tests/IGListAdapterTests.m
```diff
@@ -1067,6 +1067,24 @@ - (void)test_whenSupplementarySourceReturnsNegativeSize_thatAdapterReturnsZero {
...
+  IGTestSupplementarySource *supplementarySource = [IGTestSupplementarySource new];
+  supplementarySource.collectionContext = self.adapter;
+  supplementarySource.supportedElementKinds = @[UICollectionElementKindSectionFooter];
+  
+  IGListSectionController *controller = [self.adapter sectionControllerForObject:@1];
+  controller.supplementaryViewSource = supplementarySource;
+  supplementarySource.sectionController = controller;
+  
+  const CGSize size = [self.adapter sizeForSupplementaryViewOfKind:UICollectionElementKindSectionHeader
+                                                       atIndexPath:[NSIndexPath indexPathForItem:0 inSection:1]];
```
rnystrom: Reading through this test I'm even more convinced that this isn't correct. Why would a non-zero size be returned for an object that doesn't exist? The only object in the array is `@1`, so its at section 0 in the collection view. Requesting the size of a supplementary view at section 1 should return `CGSizeZero` b/c the object is out of bounds.

gobetti: IMO you can consider this a second bug, then. Taking your last statement:
> Requesting the size of a supplementary view at section 1 should return CGSizeZero b/c the object is out of bounds.
If you create a test for that in the current codebase, it will fail, as it won't return CGSizeZero, instead it will return the CGSize that you have defined for the IGListSupplementaryViewDelegate. But that is not a concern to me. The method's responsibility is to return the size for a supplementary view at a given index path, if you want to add the responsibility of checking if the index path actually exists, that's cool. In my test, I'm considering such responsibility doesn't exist, so the method is just returning the size for a section with index = 1. It wouldn't be able to do so in the current code base as the only possible index is 0.

I see you won't be convinced, which should mean I'm wrong, so please instruct me which is the expected way to return different sizes for different section indexes. Thanks @rnystrom !

rnystrom: @gobetti the way we use `IGListSupplementaryViewSource` in Instagram is to either:

- Create an object per `IGListSectionController` that conforms to the supplementary source protocol. Attach it as a delegate, inject as many dependencies as you want. @jessesquires might have more tips on this too.
- Just assign `self.supplementaryViewSource = self` inside the section controller and implement the required methods [like this](https://github.com/Instagram/IGListKit/blob/master/Examples/Examples-iOS/IGListKitExamples/SectionControllers/FeedItemSectionController.swift#L46-L65)

I'll admit you're definitely convincing me that this API could be updated though. I like the idea of creating a single object at the VC level that you can assign to individual section controllers, but there's no way to distinguish between section controllers.

Starting to think updating the API to include the object would be a nice update! Something like:

```
- (__kindof UICollectionReusableView *)viewForSupplementaryElementOfKind:(NSString *)elementKind
  withObject:(id)object
  atIndex:(NSInteger)index;

- (CGSize)sizeForSupplementaryViewOfKind:(NSString *)elementKind
  withObject:(id)object
  atIndex:(NSInteger)index;
```

What do you think?

gobetti: The code you've linked as an example is exactly the one I used as base for mine lol and I don't see any difference other than my section controller not being the implementer of IGListSupplementaryViewSource (we have a separate class for that, which is injected/composited into the section controller and set as the supplementaryViewSource). The implementation inside that class is identical to yours, except that we want to return one size for section index 0 and another size for other indexes. Sorry for being so repetitive: in he current codebase there's no way we'll receive an index different than 0. My PR does exactly what we need, and I'm not trying to convince you in any way, I'm actually trying to understand where I'm wrong to have this bugfix working for me, fixing @MarvinNazari 's crash and not making sense for you.
This extra object parameter doesn't seem to help me, I won't use it because today I don't need it either, I need exactly what's in your code example. Try to return a different size for the second section in your example and update me.

Thanks for still going through this!

gobetti: Ooooooooh I finally understood. You want every section to have an index 0 because each section controller is some sort of auto-sufficient unit that has no knowledge about the other sections, so if I'm a section controller and I'm implementing my own supplementary view source, I should know everything I need for my own view, without depending on other sections. I see what you mean, this is good!

But do you see now that it also makes sense if the section controller knows its section index? I thought this was possible and intended to be possible from the index parameter that is being passed to the supplementary view source. However, now I need to go back to my question, you answered it before but I didn't understand a practical usage. What is this index parameter for in your current architecture? Where do you use it in Instagram for example?

gobetti: Just adding that the same fix would have to be applied to this method if someone wanted to return a different view according to the section index:
`- (UICollectionReusableView *)collectionView:(UICollectionView *)collectionView viewForSupplementaryElementOfKind:(NSString *)kind atIndexPath:(NSIndexPath *)indexPath`

It is also taking the item from the index path. So I can see you're offering people the ability of returning multiple views and multiple sizes for each view, and probably all I need to complete my puzzle is to understand:
1) what should I do in order to be able to receive different indexes in my IGListSupplementaryViewSource implementation?
2) let's say that I succeed to have 2 supplementary views, which are the same class but they have different sizes. So I have indexes 0 and 1 for my supplementary views. How can I code the view index 0 to be used for section index 0, and the view index 1 to be used for the other section indexes?

I'm feeling this will end the discussion. Thanks again!

rnystrom: > You want every section to have an index 0 because each section controller is some sort of auto-sufficient unit that has no knowledge about the other sections

💯 

> I thought this was possible and intended to be possible from the index parameter that is being passed to the supplementary view source.

You're right that it's not possible when a completely separate object implements this protocol. We originally designed this API thinking section controllers would be the ones implementing `IGListSupplementaryViewSource` and not separate objects. Since then, internally we have implementations creating their own supplementary source objects but also providing refs to a section controller as a workaround.

Atm a section controller can get info about its position with [these APIs](https://github.com/Instagram/IGListKit/blob/master/Source/IGListSectionController.h#L100-L113), so if the section controller implemented `IGListSupplementaryViewSource`, it becomes trivial to get the section.

> what should I do in order to be able to receive different indexes in my IGListSupplementaryViewSource implementation?

I'd make your section controllers conform to the protocol and use `self.isFirstSection` to determine if its the first section controller or not.

> let's say that I succeed to have 2 supplementary views, which are the same class but they have different sizes. So I have indexes 0 and 1 for my supplementary views. How can I code the view index 0 to be used for section index 0, and the view index 1 to be used for the other section indexes?

If I'm understanding correctly, I think that might overcomplicate things. Instead do something like this:

```swift
class MySectionController: IGListSupplementaryViewSource {

  // implement stuff you need here

  // MARK: IGListSupplementaryViewSource

  func supportedElementKinds() -> [String] {
    return [UICollectionElementKindSectionHeader]
  }

  func viewForSupplementaryElement(ofKind elementKind: String, at index: Int) -> UICollectionReusableView {
  let view = collectionContext?.dequeueReusableSupplementaryView(
    ofKind: UICollectionElementKindSectionHeader,
    for: self,
    class: MyViewClass.self,
    at: index) as! MyViewClass
  // do some config
  return view
  }

  func sizeForSupplementaryView(ofKind elementKind: String, at index: Int) -> CGSize {
    let width = collectionContext!.containerSize.width
    let height: CGFloat = isFirstSection ? 100 : 50
    return CGSize(width: width, height: height)
  }

}
```

gobetti: Thank you @rnystrom ! I'm not a big fan of how IGListKit forces sometimes that some protocols are meant to be implemented by certain classes, but I also acknowledge that in order to achieve such power and flexibility, some drawbacks will always be in the way. With that said, I didn't know the `isFirstSection` property which is exactly what I need for my specific problem. I'll keep having a separate object implementing the protocol, but I'll set this property in it. I'm closing both my PR and my issue. I'm sorry for wasting your time - at least this ended up being productive, right? 😄

rnystrom: @gobetti no worries at all, and def not a waste of anyone's time! Very productive feedback, and helps me recognize some weaknesses in our language/documentation. Will def be making some additions from this, and really appreciate your patience!

Issue #699, opened 2017-04-25
Crash with embedded CollectionView


shuhrat10:
Hi,

I have CollectionView with embedded horizontal CollectionView. For example:

Category 1:
   Items [Horizontal collectionView]
Category 2:
   Items [Horizontal collectionView]
e.t.c. 

When I start moving between sections and items my application crashed. It's not happening all the time. I'm not reloading any section, I think it's happening automatically when you move from top to the end. Any idea what can cause this issue? See attached screenshot for more details: 

I'm using latest Xcode 8.3.2 and IGListKit from master branch. commit: ee4f3c95c6861f2b5b6b958e549e1842a740ae61

Thank you.

![iglistadapter_m](https://cloud.githubusercontent.com/assets/62485/25389433/9eda8c1c-299e-11e7-8d2d-cc6b76a47dd4.jpg)

rnystrom:
@shuhrat10 do you have an example reproducing this issue? We had this happen internally this week and it ended up being implementation error.

When are you attaching the collection view to the adapter?

shuhrat10:
Hi @rnystrom,

I don't have an example with this issue. It's happen in our new project today, and I wont be able to reproduce this crash. But I will keep testing to try reproduce this issue. 

In viewDidLoad I set:
`
adapter.collectionView = collectionView
adapter.dataSource = self
`

and then

`
func objects(for listAdapter: IGListAdapter) -> [IGListDiffable] {
  return FeaturedLoader.shared.categories
}
`

I don't need to execute adapter.performUpdates(animated: true) because I all data ready for use before this controller will be initialized.

rnystrom:
@shuhrat10 where/how is the embedded adapter's collection view set?

shuhrat10:
On ViewController I have 

```
func listAdapter(_ listAdapter: IGListAdapter, sectionControllerFor object: Any) -> IGListSectionController {
    return CategorySectionController(height: 240)
  }
```

```
class CategorySectionController: IGListSectionController, IGListAdapterDataSource {


  // MARK: - Variables
  
  var model: Featured?
  var sectionHeight: CGFloat

  
  // MARK: - Lifecycle
  
  init(height: CGFloat, sectionInset: UIEdgeInsets? = nil) {
    sectionHeight = height
  }
  
  
  // MARK: - IGListAdapterDataSource

  lazy var adapter: IGListAdapter = {
[...]

PR #721, opened 2017-05-09
Fix reusing collection views among multiple adapters

rnystrom:

When multiple `UICollectionView`s are reused among multiple `IGListAdapter`s (horizontal lists embedded in a bigger, vertical list), the mapping between the old view and adapter is not broken. When a new collection view is attached to the old adapter, an unrelated adapter->collection view reference will be erroneously broken.

The unit test in this diff fails on the second `XCTAssertEqual(adapter2.collectionView, collectionView1);` without this patch.

Issue fixed: #699

t15661237 internally

cc @amonshiz since we were talking about this

### Checklist


rnystrom:
h/t @yifanpeng for finding this!

Issue #707, opened 2017-05-01
Cut 3.0.0 release


rnystrom:
- [x] Final review and update of header docs and guides
- [x] Final review of changelog
- [x] Regenerate docs
- [x] Update pod spec version
- [x] Update xcodeproj version
- [x] Run `pod install` on all examples (**must happen on FB internal** because of sync issues)
- [x] Merge `master` into `stable` via cmd-line and push
- [x] Confirm `stable` is `0|0` [ahead/behind](https://github.com/Instagram/IGListKit/branches)
- [x] Create [GitHub release](https://github.com/Instagram/IGListKit/releases) from `stable`
- [x] Paste changelog into GH release notes
- [x] Publish GitHub release
- [x] Run `pod lib lint`
- [x] Push updated podspec: `pod trunk push IGListKit.podspec`
- [x] Verify new release on [CocoaPods](https://cocoapods.org/pods/IGListKit)
- [x] Tweet all the tweets

rnystrom:
Blocked on #593, #661, and #671

amonshiz:
#671 is now closed

Sherlouk:
Not sure if this is covered by `Final review and update of header docs and guides` but we need to update all the version references to 3.0 (for example the installation guide and on the readme)

jessesquires:
@Sherlouk yep! want to PR that?

jessesquires:
@rnystrom @amonshiz @Sherlouk 

looks like 3.0 is ready! just need to re-gen docs (#728) and double check that everything is good to go.

all extra eyes will help! 👀

jessesquires:
3.0 is released!

PR #661, opened 2017-04-17
3.0 migration guide

rnystrom:
Will add to this as we close out the [milestone](https://github.com/Instagram/IGListKit/milestone/3).

Issue fixed: #475

PR #728, opened 2017-05-10
Regen docs for 3.0

rnystrom:
~~Land after #661, #730 and D5031184~~

Rebuild the docs from script. Couple notes:

- Went from 100% to 99%, does that matter? Script output says
  - > 99% documentation coverage with 2 undocumented symbols
  - Looks like `NS_SWIFT_NAME` throws documentation for a loop on `IGListDiffOption` and `IGListExperiment`
- Swift names are different, will that be a problem?

Part of #707

#trivial


jessesquires:
@rnystrom -- do those 2 symbols still show up in the generated docs and the % is just wrong? or does jazzy actually not generate docs?

rnystrom:
Docs are there, I think jazzy counts them as 2x since we have the Swift rename annotation.

jessesquires:
found a few minor pods issues and re-podded. otherwise all looks good to land!

jessesquires:
going to proof-read later and import

PR #734, opened 2017-05-11
Update version number in docs

Sherlouk:
Part of #707 
#trivial


BasThomas:
🤔, do you want to lock in clients to `3.0.0`? Specifying just `3.0` will make things like `pod update` allow upgrading to patch versions.

jessesquires:
@BasThomas the `~>` should allow minor and patch version updates

BasThomas:
As I [understand](http://stackoverflow.com/a/20213930) it, this will let you go up to `8.0.9`, but not higher.

jessesquires:
@BasThomas 👍 i think that might be right. will update

PR #796, opened 2017-05-26
Amend swiftlint to work with older installations

Sherlouk:
image

Noticed an issue when trying to build the examples project that I couldn't because of the Swiftlint build phase (see the image above)

As it turns out in between the version I have installed (0.16.1) and the latest (0.19.0) they've made it so you can drop `lint` from the command (which is the default task).

I see two options really:
1. We add `lint` to the command which at least lets older versions of Swiftlint compile the project without making changes to run phases. It's worth noting that this means that certain (newer) rules won't be detected.
2. We add a way of detecting if they have an older version (likely by comparing the output of `$(swiftlint version)` to the version we know it works on. And display a more appropriate message and not fail the build.

Happy to discuss this

@heshamsalman @jessesquires @rnystrom

### DIFF
--- a/Examples/Examples-iOS/IGListKitExamples.xcodeproj/project.pbxproj
+++ b/Examples/Examples-iOS/IGListKitExamples.xcodeproj/project.pbxproj
@@ -797,7 +797,7 @@
 			);
 			runOnlyForDeploymentPostprocessing = 0;
 			shellPath = /bin/sh;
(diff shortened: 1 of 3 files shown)


heshamsalman:
I don't take issue with adding `lint` to the command -- I think it's definitely the simplest solution. 

The only issue that may arise is a discrepancy between Danger output re: linting, and local output. Not sure if it's immediately obvious to the end user that the discrepancy is bc of outdated dependencies.

Sherlouk:
@heshamsalman I think we can probably deal with the discrepancy within the PR if Danger raises something - better than someone raising an issue saying they can't build the project!

jessesquires:
Agreed, sounds good to me! Thanks @Sherlouk !

PR #820, opened 2017-06-21
Improved the documentation

weyert:
Improved the documentation so it that the Getting Started guide calls out that you need to set `dataSource`-property of the `IGListAdapter` as I totally misse that.


Source/IGListAdapter.m
```diff
@@ -272,6 +272,7 @@ - (void)performUpdatesAnimated:(BOOL)animated completion:(IGListUpdaterCompletio
     id<IGListAdapterDataSource> dataSource = self.dataSource;
     UICollectionView *collectionView = self.collectionView;
     if (dataSource == nil || collectionView == nil) {
+        IGLKLog(@"Warning: Your call to performUpdatesAnimated-method is ignored as dataSource or collectionView haven't been set.");
```
jessesquires: instead of "performUpdatesAnimated-method" we can use `%s` and pass `__PRETTY_FUNCTION__` 😄

Source/IGListAdapter.m
```diff
@@ -308,6 +309,7 @@ - (void)reloadDataWithCompletion:(nullable IGListUpdaterCompletion)completion {
     id<IGListAdapterDataSource> dataSource = self.dataSource;
     UICollectionView *collectionView = self.collectionView;
     if (dataSource == nil || collectionView == nil) {
+        IGLKLog(@"Warning: Your call to reloadDataWithCompletion-method is ignored as dataSource or collectionView haven't been set.");
```
jessesquires: same as above 😊

jessesquires:
thanks @weyert ! just a few comments 😄

weyert:
Thank you, I will make the changes :)

weyert:
@jessesquires @rnystrom made the change as requested 👍

PR #828, opened 2017-06-23
Compare using FLT_EPSILON

amonshiz:
Floating point math loses precision and so comparing two floats that are likely to only differ in the mantissa, and even then by extremely small amounts, is tricky. Exact equality often fails in these cases and asserting on exact equality is probably not necessary.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,11 @@
 The changelog for `IGListKit`. Also see the [releases](https://github.com/instagram/IGListKit/releases) on GitHub.
 
 3.1.0 (**upcoming release**)
+
+### Fixes
+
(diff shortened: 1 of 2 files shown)


amonshiz:
wait, nope, this isn't 100% correct because it is only checking case of numbers extremely close to each other.

jessesquires:
🎉

amonshiz:
what is travis' problem?

BasThomas:
Danger, it seems?

amonshiz:
@rnystrom @jessesquires any ideas on the danger issue?

BasThomas:
I think its just because there is no Changelog entry :)

amonshiz:
okay.

Issue #831, opened 2017-06-25
Sticky headers example


weyert:
Create an example that uses the IGList's collection view layout's support for sticky headers

weyert:
I will update the Supplementary View example to use sticky headers

bogren:
Not up to date in the sticky header discussion, I don't think it's been mentioned before. If the layout is set to `UICollectionViewFlowLayout `, you can set [sectionHeadersPinToVisibleBounds](https://developer.apple.com/documentation/uikit/uicollectionviewflowlayout/1617699-sectionheaderspintovisiblebounds) to the collection view it's set to support. 

Adding works fine for apps supporting iOS 9.
```
let layout = collectionView.collectionViewLayout as? UICollectionViewFlowLayout
        if #available(iOS 9.0, *) {
            layout?.sectionHeadersPinToVisibleBounds = true
        }
```

aelam:
@bogren how about iOS 8

bogren:
I'm currently reading up on the discussion here, and running a few tests in the examples. There is support and here is an [example](https://github.com/Instagram/IGListKit/pull/834) not using the `UICollectionViewFlowLayout` property.

heumn:
Maybe close this as the PR to add it to the example was closed @rnystrom ? 🤔

PR #834, opened 2017-06-25
Created a more charming example out of the supplementary views exampl…

weyert:
…e so it uses sticky headers

Issue fixed: #831 

### Checklist

### DIFF
--- a/Examples/Examples-iOS/IGListKitExamples/ViewControllers/SupplementaryViewController.swift
+++ b/Examples/Examples-iOS/IGListKitExamples/ViewControllers/SupplementaryViewController.swift
@@ -16,12 +16,13 @@ import UIKit
 import IGListKit
 
 final class SupplementaryViewController: UIViewController, ListAdapterDataSource {
-    
+
     lazy var adapter: ListAdapter = {
         return ListAdapter(updater: ListAdapterUpdater(), viewController: self)
     }()
-    let collectionView = UICollectionView(frame: .zero, collectionViewLayout: ListCollectionViewLayout(stickyHeaders: true, topContentInset: 0.0, stretchToEdge: false))
-    
+    let collectionView = UICollectionView(frame: .zero, collectionViewLayout: ListCollectionViewLayout(
+        stickyHeaders: true, topContentInset: 0.0, stretchToEdge: false))
+
     let feedItems = [
         FeedItem(pk: 1, user: User(pk: 100, name: "Jesse", handle: "jesse_squires"), comments: [
             "You rock!",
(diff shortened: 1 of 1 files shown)


Examples/Examples-iOS/IGListKitExamples/ViewControllers/SupplementaryViewController.swift
```diff
@@ -20,17 +20,50 @@ final class SupplementaryViewController: UIViewController, ListAdapterDataSource
...
+            "What's the eta?"
+            ])
     ]
 
     override func viewDidLoad() {
         super.viewDidLoad()
+        if let flowLayout = collectionView.collectionViewLayout as? ListCollectionViewLayout {
+            let statusBarSize = UIApplication.shared.statusBarFrame.size
+            let navigationBarVisible = self.navigationController?.isNavigationBarHidden == false
+            let navigationBarHeight = ( navigationBarVisible ? self.navigationController!.navigationBar.frame.height : 0 )
```
rnystrom: Any chance we can remove reaching into the nav and status bar frames? We don't want to recommend doing this as its not really future-friendly. What about the top layout guide?

weyert: @rnystrom Sure, I will see how I can change the code to make it work without accessing the nav/status

weyert:
@rnystrom Thank you for suggesting topLayoutGuide. Sadly, it's zero in `viewDidLoad` because the view hasn't been drawn yet. I have moved it to `viewDidAppear`

rnystrom:
@weyert another idea (and I honestly have no idea how good a practice this is...) is to hand the _layout guide itself_ to the section controller. Then, come sizing time, the SC can ask for the `length`.

I came up w/ this idea on my GitHub app. [Here](https://github.com/rnystrom/Freetime/blob/master/Classes/Notifications/NotificationsViewController.swift#L175) is where I pass in the layout guide, and [here](https://github.com/rnystrom/Freetime/blob/master/Classes/Notifications/NoNewNotificationsSectionController.swift#L26) is where I use its value.

The end result is a cell that fits perfectly within the remaining height.

What do you think?

weyert:
@rnystrom I would be happy to implement it. Only wouldn't it make the code of this specific example a bit more complicated then really necessary? It's your call, let me know :)

weyert:
@rnystrom I have had a look at your suggestion again but I don't think it would useful here. For the sticky headers we need to set a offset in `IGListKit`'s layout so that sticks below the navigation bar instead of under it. Looking at your code in `FreeTime` we won't be able to use that approach while it's great idea! The problem is that it's a collection view layout property and you can not access that through `collectionContext` but I also think it's not a great idea to keep changing the layout properties between section controllers.

weyert:
How shall we progress this?

rnystrom:
@weyert I'm going to close this out, I want to make sure our examples are really lightweight, and having to explain what is happening beyond code might be a little confusing. I think what we have now are pretty good examples using our custom layout. Lmk if you disagree and think this is going to be really useful.

weyert:
Sounds good to me! Just wanted let one example use the sticky headers :-)

PR #835, opened 2017-06-26
Fixed the issue with the Search example not working

weyert:
I have updated the Search Autocomplete example which stopped working as the words don't have unique `diffIdentifier`'s. This caused by the check that good introduced to ensure all objects assigned to a section controller are unique. Fixed this issue by removing all the double words.

### Checklist


amonshiz:
thanks for doing this! i noticed this on Friday and discussed with @rnystrom. we agreed that going with a `Set` instead of `Array` for the `var words` would solve this and ensure no individual words are lost.

weyert:
Why would that fix the issue? Would that cause different diffIdentifier values or is the check skipped when using a Set? But I am happy to change it!

amonshiz:
it would fix the issue because a `Set` would only have one instance of each word, and since the word itself is used as the diff identifier there would be no duplicate identifiers. :)

weyert:
Oh right, somehow I thought you wanted to keep the double words 🙈

weyert:
@amonshiz Fixed it 👍

PR #836, opened 2017-06-26
fixed a bug:contentSize was inaccurate in some case

HuylensHu:

Issue fixed: contentSize was inaccurate in some case such as item `Storyboard` of demo,if the last box is smaller than it is in front of it.the contentSize is always erroneously calculated as the height of last box element.

### Checklist


Examples/Examples-iOS/IGListKitExamples-UITests/StoryboardViewControllerUITests.swift
```diff
@@ -0,0 +1,68 @@
...
+            }
+            
+            if cellFrame.origin.x == 0{
+                //the last whole line has been traversed
+                break;
+            }
+        }
+        return true
+    }
+}
```
rnystrom: We already have a test suite for the layout, mind moving this to [here](https://github.com/Instagram/IGListKit/blob/master/Tests/IGListCollectionViewLayoutTests.m)?

Examples/Examples-iOS/IGListKitExamples-UITests/StoryboardViewControllerUITests.swift
```diff
@@ -0,0 +1,68 @@
+//
+//  StoryboardViewControllerUITests.swift
+//  IGListKitExamples
+//
+//  Created by Huylens on 2017/7/1.
+//  Copyright © 2017年 Instagram. All rights reserved.
+//
```
rnystrom: We should move stuff to the layout tests, but for future's sake we require the FB copyright header in all source, test, and example files.

rnystrom:
Did some more thinking about this. Definitely a real bug, but this changes the `collectionViewContentSize` method from an `O(1)` op to `O(n)`. Here's an idea:

- Alongside the [sectionData ivar](https://github.com/Instagram/IGListKit/blob/master/Source/IGListCollectionViewLayout.mm#L84), add a `CGFloat _height;` ivar
- Assign `_height = nextRowY` at the end of [this method](https://github.com/Instagram/IGListKit/blob/master/Source/IGListCollectionViewLayout.mm#L429-L432)

That'll find the max Y bounds of the entire layout _and_ keep the algo constant! @HuylensHu what do you think?

edmonston:
100% agree with @rnystrom's suggestion -- you've got the height handy at the end of `cacheLayout` so you could just save it rather than having to recompute in linear time. When/if this is merged I can adjust my PR #857 (which adds horizontal scrolling support) to save the width in the horizontal case.

rnystrom:
@HuylensHu are you still interested in fixing this?

HuylensHu:
I'm sorry, I don't have more energy to focus on this

PR #837, opened 2017-06-27
Fixed the Swift lint issues

weyert:
Fixed a bunch of swift lint issues the easy way :)
Guess we can change it to """ once Swift 4 is introduced

### DIFF
--- a/Examples/Examples-iOS/IGListKitExamples/ViewControllers/MixedDataViewController.swift
+++ b/Examples/Examples-iOS/IGListKitExamples/ViewControllers/MixedDataViewController.swift
@@ -23,6 +23,7 @@ final class MixedDataViewController: UIViewController, ListAdapterDataSource {
     let collectionView = UICollectionView(frame: .zero, collectionViewLayout: UICollectionViewFlowLayout())
 
     let data: [Any] = [
+        // swiftlint:disable:next line_length
         "Maecenas faucibus mollis interdum. Duis mollis, est non commodo luctus, nisi erat porttitor ligula, eget lacinia odio sem nec elit.",
         GridItem(color: UIColor(red: 237/255.0, green: 73/255.0, blue: 86/255.0, alpha: 1), itemCount: 6),
         User(pk: 2, name: "Ryan Olson", handle: "ryanolsonk"),
@@ -73,10 +74,15 @@ final class MixedDataViewController: UIViewController, ListAdapterDataSource {
 
     func objects(for listAdapter: ListAdapter) -> [ListDiffable] {
         guard selectedClass != nil else {
+            // swiftlint:disable:next force_cast
             return data.map { $0 as! ListDiffable }
         }
+        // swiftlint:disable:next force_cast
         return data.filter { type(of: $0) == selectedClass! }
-            .map { $0 as! ListDiffable }
+            .map {
+                // swiftlint:disable:next force_cast
+                $0 as! ListDiffable
+            }
     }
 
     func listAdapter(_ listAdapter: ListAdapter, sectionControllerFor object: Any) -> ListSectionController {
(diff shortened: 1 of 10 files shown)


weyert:
@rnystrom I have made the requested changes. Let me know if you are agree with my change regarding force casts. It's the only way I know 😬

Examples/Examples-iOS/IGListKitExamples/ViewControllers/MixedDataViewController.swift
```diff
@@ -23,7 +23,8 @@ final class MixedDataViewController: UIViewController, ListAdapterDataSource {
     let collectionView = UICollectionView(frame: .zero, collectionViewLayout: UICollectionViewFlowLayout())
 
     let data: [Any] = [
```
rnystrom: @weyert what about changing this to

```swift
let data: [ListDiffable]
```

You might have to soft-cast the stuff in the array literal, like

```swift
"String" as ListDiffable,
// etc
```

But its not a force-cast. Wanna give that a shot? Will get rid of the fatals too I think.

weyert: @rnystrom I have given it a try I had to make a code change to make the filtering work again. As now it didn't wanted to filter the strings straight away. After some struggling to push my commit to my branch I think it made it up here.

jessesquires:
@rnystrom @heshamsalman -- we're using the user's `SwiftLint` install, correct?

How can we ensure all users are using the same version? For example, if I lint with `0.19` but you lint with `0.21` -- we'd probably get different results...

heshamsalman:
@jessesquires yup -- for in-editor issues we rely on the user's installations -- there's not much we can do there except modify our script to expect a specific version (maybe poll for the current version, and expect _that_ version?)

weyert:
Does that mean you are going to accept this PR? :)

Examples/Examples-iOS/IGListKitExamples/ViewControllers/MixedDataViewController.swift
```diff
@@ -72,18 +75,27 @@ final class MixedDataViewController: UIViewController, ListAdapterDataSource {
...
+                return $0
+            }
         }
-        return data.filter { type(of: $0) == selectedClass! }
-            .map { $0 as! ListDiffable }
+        return data.filter {
+                if selectedClass == String.self, $0 is String {
+                    return true
+                }
+                return type(of: $0) == selectedClass
```
heshamsalman: Isn't this condition a superset of the one above?  Also -- not a big fan of the double indent above.

jessesquires:
@heshamsalman 👍 

+1 #847

Examples/Examples-iOS/IGListKitExamples/ViewControllers/MixedDataViewController.swift
weyert: I didn't get it working to filter out Strings without this if-statement

jessesquires:
what's the status here? @weyert @rnystrom

weyert:
I will merge the latest branch into my `swiftlint`-branch and see if I can fix any outstanding lint issues. I think after that we could then merge it?

rnystrom:
Been thinking about this a bit. I'm almost inclined to remove SwiftLint from the examples. I feel it just adds noise and confusion, and all the stuff we have to do to satisfy it honestly leaves stuff _more_ confusing. @jessesquires what do you think?

heumn:
@rnystrom for most of my projects I just use Nick Lockwoods `SwiftFormat`
https://github.com/nicklockwood/SwiftFormat 
Highly reccomended and way simpler to use than SwiftLint. I kinda see that as more appropriate when maintaining a style and linting across a larger team on a bigger project 🤔

rnystrom:
Ya, having it just for an examples lib seems like overkill. I'm gonna gut it.

Issue #524, opened 2017-03-04
Cell preselection


pharosproduction:
I'm trying to preselect a cell in `cellForItem(at index: Int) -> UICollectionViewCell` by `cell.isSelected = true` in section controller.
It works fine and cell becomes selected. When I try to deselect a cell by tapping it cell doesn't react at all. Tap triggers `isHighlighted` method but doesn't want to trigger neither `isSelected` in cell nor `didDeselectItemAt` delegate method.
This issue is related only to preselected cells. All other cells work fine in the same collection.

rnystrom:
@pharosproduction can you describe a bit what you're trying to do? `UICollectionViewCell` selection APIs aren't really supposed to be changed manually, you're supposed to let the `UICollectionView` internals handle all of that for you.

The [docs](https://developer.apple.com/reference/uikit/uicollectionviewcell/1620130-isselected) say:

> You typically do not set the value of this property directly. Changing the value of this property programmatically does not change the appearance of the cell. The preferred way to select the cell and highlight it is to use the selection methods of the collection view object.

If you're trying to do custom selection state, we recommend making your own properties and settings.

pharosproduction:
There is a `Set` of groups displayed as a two column list, one section with one row for each group. Some of them should be selected when list appears. Each group contains a selection state(bool). Groups amount is a variable(pagination load). What is the best way to select marked for selection cells and where?
I have tried already different approaches including `selectItem(at: IndexPath?, animated: Bool, scrollPosition: UICollectionViewScrollPosition)` but no luck to find the best way to do this. The goal is just to select a cell without triggering any delegate methods so user can deselect it or proceed with a default selection set(fetched by `indexPathForSelectedItems`)

rnystrom:
@pharosproduction at Instagram we use custom properties to represent selected state in our UI instead of relying on the `UICollectionView` built-in stuff. We would do something like:

- Add a custom `@property BOOL checkmarkDisplayed;` to our cell subclass
- In the `setCheckmarkDisplayed:(BOOL)checkmarkDisplayed` setter, set the state and call `[self setNeedsLayout]` and set `hidden` on our checkmark image view
- Position the view in `layoutSubviews`

Then in `cellForItemAtIndex:` we can manually set the state w/out worrying about `UICollectionView` internals.

Then in `didSelectItemAtIndex:` you can flip the selection state, both in the cell and wherever you're tracking it.

pharosproduction:
Got it. Thanks.
[...]

PR #853, opened 2017-07-12
Add cell deselection API

rnystrom:

Adding support for a cell deselection API. Trying to make some headway to move and drag+drop support, but also want better stock `UICollectionView` API support. Will also assist eventual `UITableView` support.

- Added overridable API to `IGListSectionController`
- Support for stacked SC
- Breaking, required protocol for binding SC

Assists #524 and #184

### Checklist

### DIFF
--- a/Source/IGListBindingSectionControllerSelectionDelegate.h
+++ b/Source/IGListBindingSectionControllerSelectionDelegate.h
@@ -36,7 +36,10 @@ NS_SWIFT_NAME(ListBindingSectionControllerSelectionDelegate)
  @param sectionController The section controller the deselection occurred in.
  @param index The index of the deselected cell.
  @param viewModel The view model that was bound to the cell.
+
+ @note Method is `@optional` until the 4.0.0 release where it will become required.
  */
+@optional
 - (void)sectionController:(IGListBindingSectionController *)sectionController
    didDeselectItemAtIndex:(NSInteger)index
                 viewModel:(id)viewModel;
--- a/Source/IGListBindingSectionControllerSelectionDelegate.h
+++ b/Source/IGListBindingSectionControllerSelectionDelegate.h
@@ -30,6 +30,17 @@ NS_SWIFT_NAME(ListBindingSectionControllerSelectionDelegate)
      didSelectItemAtIndex:(NSInteger)index
                 viewModel:(id)viewModel;
 
+/**
+ Tells the delegate that a cell at a given index was deselected.
+
+ @param sectionController The section controller the deselection occurred in.
+ @param index The index of the deselected cell.
+ @param viewModel The view model that was bound to the cell.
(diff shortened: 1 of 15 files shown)


weyert:
Looks alright to me. I see you made a default implementation for this method

jessesquires:
`3.1.0` cannot contain breaking changes according to SemVer, we'll need to make this `4.0.0`. 😊 

I think this is fine though. People tend to freak out about version numbers, I don't think it's a big deal. What's important is communicating changes + proper SemVer.

Looks like there's a handful of things in the [3.1.0 milestone](https://github.com/Instagram/IGListKit/milestone/4) that we could wrap up -- we could merge those and this, rename to 4.0 and push a release. The current 4.0 would then get bumped to 5.0.

Thoughts?

cc @amonshiz

rnystrom:
@jessesquires TIL! Another option: I could make this with `@optional` then remove for 4.0? Not opposed to a bigger 4.0 release either.

Or maybe I hold off on this, land some internal changes + fixes (I have some diffs up inside), release 3.1, then land + prep 4.0?

amonshiz:
i'm fine bumping table view support to 5.0

jessesquires:
@rnystrom -- either of those sound good to me.

RE: 4.0 release -- We *could* just merge this and immediately push 4.0, then anything else goes into 4.1. I was just thinking, since we have some low-hanging fruit, we could include more in a 4.0 release to make it "more worth it"

rnystrom:
Ya I'd rather package a bit more into the 4.0 since semantically its kind of a big deal. Atm there's barely anything compared to 3.0.

Source/IGListBindingSectionControllerSelectionDelegate.h
```diff
@@ -30,6 +30,20 @@ NS_SWIFT_NAME(ListBindingSectionControllerSelectionDelegate)
...
+/**
+ Tells the delegate that a cell at a given index was deselected.
+
+ @param sectionController The section controller the deselection occurred in.
+ @param index The index of the deselected cell.
+ @param viewModel The view model that was bound to the cell.
+
+ @note Method is `@optional` until the 4.0.0 release where it will become required.
+ */
+@optional
```
amonshiz: i thought swift didn't support `@optional` which is why there has been a push internally to remove all `@optional` protocol methods and properties.

rnystrom: Still supports it, discouraged b/c of how it will silently break things

https://useyourloaf.com/blog/swift-optional-protocol-methods/

rnystrom:
Importing for eyeballs

jessesquires:
@rnystrom sounds good to me 👍 

let's make sure we create a 4.0 task to remove `@optional`

Issue #698, opened 2017-04-24
Add debug dump support to IGListBindingSectionController


rnystrom:
Inspired by #696, printing the view models arrays would be really helpful when dumping debug logs.

PR #856, opened 2017-07-16
Added debug dump support to IGListBindingSectionController

candance:

Issue fixed: #698 

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,10 @@ The changelog for `IGListKit`. Also see the [releases](https://github.com/instag
 3.1.0 (**upcoming release**)
 -----
 
+### Enhancements
+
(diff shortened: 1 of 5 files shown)


candance:
Your suggestions make sense, but I think my implementation covers the issue. I wouldn't be too concerned about the global state or the extra hash map, since we're already using this approach for 'IGListAdapter'. Also, considering this code is only complied in debug versions of IGListKit, I think the fix should suffice as is.

However,  I'm happy to implement your suggestions, if you feel it is important to reuse the print logic in 'IGListSectionMap' for 'IGListBindingSectionController'. I do have a few questions if we go that route.

Let me know!

rnystrom:
@candance this def works as is, but I’m mostly concerned with the long term maintenance cost of the project, and global state usually adds expense 😅. 

I’d prefer making the changes before we can merge. Thanks!

Issue #752, opened 2017-05-22
Horizontal list support in IGListCollectionViewLayout


rnystrom:
Inspired by #750. If you wanted to have a horizontally scrolling list that put sections on the "same line" (which would actually be a column), this wont work with IGListKit since `UICollectionViewFlowLayout` doesn't support this.

jessesquires:
@rnystrom Isn't this "collection view inside a collection cell" ?

rnystrom:
@jessesquires it's actually making `IGListCollectionViewLayout` work w/ horizontal scrolling (its fixed to vertical using height/y atm).

edmonston:
So I was going to take a crack at this one, if no one is working on it... If we generalize `IGListCollectionViewLayout`, what's the desired behavior for section headers when scrolling horizontally -- always fill vertically (minus insets) and, if sticky, pin to the left (offset by `stickyHeaderOriginAdjustment`)?

zhubofei:
@rnystrom @jessesquires @edmonston With UICollectionViewFlowLayout, supplementary views are placed on the left/right side of each section. But that's kind of lame... If you take a look at built-in emoji keyboard, each section has a sticky header on top. It looks nice but might not be easy to copy? (For example, headers have to have same height?)

edmonston:
Yeah I agree @zhubofei that `UICollectionViewFlowLayout`'s way of handling horizontal scrolling with supplementary views is not super useful from UI perspecive (see video). But as you point out, keeping the headers horizontal would need them to be the same height (and could be strange with short/narrow sections)... I have a local branch that generalizes `IGListCollectionViewLayout` to behave more like `UICollectionViewFlowLayout` in the horizontal axis. @rnystrom was that what you had in mind with this enhancement? If so I can work it into a PR...
![horizontal](https://user-images.githubusercontent.com/885428/27566879-038b930c-5ab6-11e7-83e5-00660a2cf27f.gif)

PR #182, opened 2016-11-10
Change all @discussion to @note

zhubofei:
Fix #171

PR #857, opened 2017-07-16
Add horizontal scrolling support to IGListCollectionViewLayout

edmonston:

Issue fixed: #752

### Checklist

This PR generalizes the layout logic in `IGListCollectionViewLayout.mm` to handle horizontally scrolling layouts, mainly by generalizing references to `width`, `height`, `x` and `y` to take scrolling direction into account. This changes the signature of `IGListCollectionViewLayout.init` as well as the names of a few properties, so it would be a breaking change. 

I added a couple of unit tests specifically for horizontal layouts -- but held off from adding a horizontal version of *every* unit test for this class, as it would basically double the number of tests. But if you want that, just let me know and I'm happy to do it.

Also let me know if you want me to add a demo VC to the Examples project that uses this new horizontal flow layout -- I have some demo code handy (I used it for testing), but didn't want to clutter up the PR if you didn't want/need it.

### DIFF
--- a/Source/IGListCollectionViewLayout.mm
+++ b/Source/IGListCollectionViewLayout.mm
@@ -14,6 +14,41 @@
 
 #import <IGListKit/IGListAssert.h>
 
+static CGFloat UIEdgeInsetsLeadingInsetInDirection(UIEdgeInsets insets, UICollectionViewScrollDirection direction) {
+    return direction == UICollectionViewScrollDirectionVertical ? insets.top : insets.left;
+}
+
+static CGFloat UIEdgeInsetsTrailingInsetInDirection(UIEdgeInsets insets, UICollectionViewScrollDirection direction) {
+    return direction == UICollectionViewScrollDirectionVertical ? insets.bottom : insets.right;
+}
+
+static CGFloat CGPointGetCoordinateInDirection(CGPoint point, UICollectionViewScrollDirection direction) {
+    return direction == UICollectionViewScrollDirectionVertical ? point.y : point.x;
+}
+
+static CGRect CGRectInset(CGRect rect, UIEdgeInsets insets) {
+    return CGRectMake(rect.origin.x + insets.left,
+                      rect.origin.y + insets.top,
+                      rect.size.width - insets.left - insets.right,
+                      rect.size.height - insets.top - insets.bottom);
+}
+
+static CGFloat CGRectGetLengthInDirection(CGRect rect, UICollectionViewScrollDirection direction) {
+    return direction == UICollectionViewScrollDirectionVertical ? rect.size.height : rect.size.width;
+}
+
+static CGFloat CGRectGetMaxInDirection(CGRect rect, UICollectionViewScrollDirection direction) {
+    return direction == UICollectionViewScrollDirectionVertical ? CGRectGetMaxY(rect) : CGRectGetMaxX(rect);
+}
+
+static CGFloat CGRectGetMinInDirection(CGRect rect, UICollectionViewScrollDirection direction) {
+    return direction == UICollectionViewScrollDirectionVertical ? CGRectGetMinY(rect) : CGRectGetMinX(rect);
+}
+
+static CGFloat CGSizeGetLengthInDirection(CGSize size, UICollectionViewScrollDirection direction) {
+    return direction == UICollectionViewScrollDirectionVertical ? size.height : size.width;
+}
+
 static NSIndexPath *headerIndexPathForSection(NSInteger section) {
     return [NSIndexPath indexPathForItem:0 inSection:section];
 }
@@ -75,7 +110,7 @@ static void adjustZIndexForAttributes(UICollectionViewLayoutAttributes *attribut
 @interface IGListCollectionViewLayout ()
 
 @property (nonatomic, assign, readonly) BOOL stickyHeaders;
-@property (nonatomic, assign, readonly) CGFloat topContentInset;
+@property (nonatomic, assign, readonly) CGFloat initialContentInset;
 @property (nonatomic, assign, readonly) BOOL stretchToEdge;
 
 @end
@@ -99,11 +134,13 @@ static void adjustZIndexForAttributes(UICollectionViewLayoutAttributes *attribut
 }
 
 - (instancetype)initWithStickyHeaders:(BOOL)stickyHeaders
-                      topContentInset:(CGFloat)topContentInset
+                      scrollDirection:(UICollectionViewScrollDirection)scrollDirection
+                  initialContentInset:(CGFloat)initialContentInset
                         stretchToEdge:(BOOL)stretchToEdge {
     if (self = [super init]) {
+        _scrollDirection = scrollDirection;
         _stickyHeaders = stickyHeaders;
-        _topContentInset = topContentInset;
+        _initialContentInset = initialContentInset;
         _stretchToEdge = stretchToEdge;
         _attributesCache = [NSMutableDictionary new];
         _headerAttributesCache = [NSMutableDictionary new];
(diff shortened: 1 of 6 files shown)


rnystrom:
@edmonston in order to land this earlier than the 4.0 release, what about keeping the existing init but just call the new initializer with `UICollectionViewScrollDirectionVertical` (and `topContentInset` can stay). We can file a follow-up for 4.0 to remove it in favor of a single initializer (which I prefer).

Just want to get this in for 3.1 or 3.2 and not have to hold it. We're trying to avoid breaking changes in minor releases.

Thoughts?

Source/IGListCollectionViewLayout.mm
```diff
@@ -14,6 +14,41 @@
...
+static CGFloat CGPointGetCoordinateInDirection(CGPoint point, UICollectionViewScrollDirection direction) {
+    return direction == UICollectionViewScrollDirectionVertical ? point.y : point.x;
+}
+
+static CGRect CGRectInset(CGRect rect, UIEdgeInsets insets) {
+    return CGRectMake(rect.origin.x + insets.left,
+                      rect.origin.y + insets.top,
+                      rect.size.width - insets.left - insets.right,
+                      rect.size.height - insets.top - insets.bottom);
+}
```
rnystrom: Isn't this exactly what `UIEdgeInsetsInsetRect` does already?

Source/IGListCollectionViewLayout.mm
```diff
@@ -14,6 +14,41 @@
...
+static CGFloat UIEdgeInsetsLeadingInsetInDirection(UIEdgeInsets insets, UICollectionViewScrollDirection direction) {
+    return direction == UICollectionViewScrollDirectionVertical ? insets.top : insets.left;
+}
+
+static CGFloat UIEdgeInsetsTrailingInsetInDirection(UIEdgeInsets insets, UICollectionViewScrollDirection direction) {
+    return direction == UICollectionViewScrollDirectionVertical ? insets.bottom : insets.right;
+}
+
+static CGFloat CGPointGetCoordinateInDirection(CGPoint point, UICollectionViewScrollDirection direction) {
+    return direction == UICollectionViewScrollDirectionVertical ? point.y : point.x;
```
rnystrom: Thoughts on using a `switch(...)` instead so the compiler will error if something ever changes? It likely wont, just good for longterm maintenance.

Source/IGListCollectionViewLayout.mm
```diff
@@ -14,6 +14,41 @@
...
+    return direction == UICollectionViewScrollDirectionVertical ? CGRectGetMaxY(rect) : CGRectGetMaxX(rect);
+}
+
+static CGFloat CGRectGetMinInDirection(CGRect rect, UICollectionViewScrollDirection direction) {
+    return direction == UICollectionViewScrollDirectionVertical ? CGRectGetMinY(rect) : CGRectGetMinX(rect);
+}
+
+static CGFloat CGSizeGetLengthInDirection(CGSize size, UICollectionViewScrollDirection direction) {
+    return direction == UICollectionViewScrollDirectionVertical ? size.height : size.width;
+}
```
rnystrom: 👌 love these methods 👌

Source/IGListCollectionViewLayout.mm
```diff
@@ -341,63 +396,78 @@ - (void)cacheLayout {
...
-            IGAssert(size.width <= paddedWidth || fabs(size.width - paddedWidth) < FLT_EPSILON,
-                     @"Width of item %zi in section %zi must be less than container %.0f accounting for section insets %@",
-                     item, section, width, NSStringFromUIEdgeInsets(insets));
-            CGFloat itemWidth = MIN(size.width, paddedWidth);
+            IGAssert(CGSizeGetLengthInDirection(size, fixedDirection) <= paddedLengthInFixedDirection
+                     || fabs(CGSizeGetLengthInDirection(size, fixedDirection) - paddedLengthInFixedDirection) < FLT_EPSILON,
+                     @"%@ of item %zi in section %zi must be less than container %.0f accounting for section insets %@",
+                     self.scrollDirection == UICollectionViewScrollDirectionVertical ? @"Width" : @"Height",
+                     item, section, CGRectGetLengthInDirection(contentInsetAdjustedCollectionViewBounds, fixedDirection),
+                     NSStringFromUIEdgeInsets(insets));
```
rnystrom: I love it!

Source/IGListCollectionViewLayout.mm
```diff
@@ -422,11 +498,11 @@ - (void)cacheLayout {
...
         sectionData[section].insets = insets;
 
-        // bump the x for the next section with the right insets
-        itemX += insets.right;
+        // bump the coord for the next section with the right insets
+        itemCoordInFixedDirection += UIEdgeInsetsTrailingInsetInDirection(insets, fixedDirection);
 
-        // find the lowest point in the section and add the bottom inset to find the next row's Y
-        nextRowY = MAX(nextRowY, CGRectGetMaxY(rollingSectionBounds) + insets.bottom);
+        // find the farthest point in the section and add the trailing inset to find the next row's coord
```
rnystrom: Even these comment updates are great!

Tests/IGListCollectionViewLayoutTests.m
```diff
@@ -106,6 +110,39 @@ - (void)test_whenLayingOutCells_withHeaderHeight_withLineSpacing_withInsets_that
...
+                                                                  items:@[
+                                                                          [[IGLayoutTestItem alloc] initWithSize:(CGSize){45,30}],
+                                                                          ]],
+                            ]];
+    XCTAssertEqual(self.collectionView.contentSize.width, 140);
+    IGAssertEqualFrame([self headerForSection:0].frame, 10, 10, 10, 85);
+    IGAssertEqualFrame([self cellForSection:0 item:0].frame, 20, 10, 45, 10);
+    IGAssertEqualFrame([self cellForSection:0 item:1].frame, 20, 20, 45, 20);
+    IGAssertEqualFrame([self headerForSection:1].frame, 80, 10, 10, 85);
+    IGAssertEqualFrame([self cellForSection:1 item:0].frame, 90, 10, 45, 30);
```
rnystrom: Great test, can see the items stacking vertically

Source/IGListCollectionViewLayout.mm
edmonston: 👍  will do

Source/IGListCollectionViewLayout.mm
edmonston: Doh, I did not even know this function already existed...

edmonston:
Definitely @rnystrom! Makes total sense to preserve the old initializer and just add a new one for now, so it isn't breaking and doesn't have to wait to 4.0.

When I update the `CHANGELOG.md` which release should I target this PR for?

I'll make this change, add the unit tests you suggested and make the other tweaks from the comments. Stay tuned.

rnystrom:
@edmonston we can make this a 3.1 enhancement 😁

edmonston:
So, I did those changes to make this PR non-breaking (in the semver sense), and everything seemed fine -- *except* I just realized that I had altered one other element of the public-facing `IGListCollectionViewLayout` API as well: I had changed the property name `stickyHeaderOriginYAdjustment` to the more general `stickyHeaderOriginOffset`. So technically it's still a breaking update. 

Should I just revert to the original name (referencing the 'Y')  for compatibility's sake, even though it's not quite accurate for a horizontal layout, and then we fix it later in 4.0?

Even nit-pickier: As it stands now, the `NS_DESIGNATED_INITIALIZER` will become the new init method, while the old one transitions to a convenience initializer. I'm hoping this is not considered a breaking change...

weyert:
@edmonston Maybe consider adding both properties and just return the value of the new property for `stickyHeaderOriginYAdjustment` and then remove `stickyHeaderOriginYAdjustment`-property in 4.0?

edmonston:
Hey @weyert that's a good suggestion. I've updated the PR so `stickyHeadersYOffset` is just a passthrough to the new property. So this should be thoroughly non-breaking, API-wise.

(I also changed two occurrences of `@discussion` to `@note` in the header comments because it seems like we made that change globally in #182 to make Jazzy docs happy.)

@rnystrom lemme know if those new unit tests (I actually added 4) seem ok.

rnystrom:
@edmonston awesome! Don't worry about changing the DI, that shouldn't break anything. Even though the naming is inconsistent, let's keep the `Y` for now so that people can upgrade w/out worry while 4.0 is WIP.

I'm out of the country this week and will land this when I'm back, just in case this breaks something I don't want to make people scramble while I'm gone!

Issue #757, opened 2017-05-22
Cache gems in Travis CI builds


heshamsalman:
We should cache gems in Travis CI to speed up builds & improve build reliability. May involve moving to a Gemfile and caching bundler.

rnystrom:
I think we used to have this but removed b/c I was fighting caches while adding more stuff. Maybe, can't remember.

Either way 👍 from me if it helps Travis, lil fella needs all it can get!

jessesquires:
for anyone looking to do this:

https://github.com/Instagram/IGListKit/blob/master/.travis.yml#L4

PR #865, opened 2017-07-24
[Bundler] Add Gemfile and setup travis to cache the dependencies

bogren:

Issue fixed: #757

Added gems to a `Gemfile`, CocoaPods is part of the default [gem set](https://docs.travis-ci.com/user/reference/osx/#Gems-in-the-global-gem-set). 

### Checklist

- [ ] ~All tests pass. Demo project builds and runs.~
- [ ] ~I added tests, an experiment, or detailed why my change isn't tested.~


rnystrom:
We added bundler in https://github.com/Instagram/IGListKit/commit/fd22609e7e014552084ba691ca3b911dd41c2cd3#diff-354f30a63fb0907d4ad57269548329e3 but then removed it in https://github.com/Instagram/IGListKit/commit/d6cae084bb1565dcdf9afb7e96d4f0e80ba6e4f0. @jessesquires you removed this and know more about this stuff than me. Any tips? Also cc @benasher44 who added it initially.

bogren:
Yes, so I read up on the git history before my PR. Moving fast traded for cached gems(?). Would be interesting to know more about the pros and cons running with Bundler and the issues you had before.

rnystrom:
ping @jessesquires @benasher44 for some input

benasher44:
I'm not sure about caching and Travis. Bundler will ensure your sanity in the long term because it won't matter what version of gems Travis has installed. It'll always use the gem you installed using `bundle install` so long as you use `bundle exec <gem cmd>` to use your installed gems. This makes it easier to reproduce issues locally.

jessesquires:
> I'm not sure about caching and Travis.

Yeah, I think this was the issue. Travis wasn't being cooperative and so at the time I was like "fuck it, let's just make sure we're using all travis defaults". I intended to restore the `Gemfile` but... 😅 you know. 😆

CHANGELOG.md
```diff
@@ -12,6 +12,7 @@ The changelog for `IGListKit`. Also see the [releases](https://github.com/instag
 ### Enhancements
 
 - Added `-[IGListSectionController didDeselectItemAtIndex:]` API to support default `UICollectionView` cell deselection. [Ryan Nystrom](https://github.com/rnystrom) (tbd)
+- Added Gemfile and setup Travis to cache the dependencies to speed up the build times [Emil Bogren](https://github.com/bogren) (tbd)
```
jessesquires: we don't need a CHANGELOG for this

Gemfile
```diff
@@ -0,0 +1,6 @@
+source 'https://rubygems.org'
+
+gem 'danger', '~> 4.3.0'
```
jessesquires: we should add cocoapods to this

jessesquires:
actually, we should probably just cherry-pick fd22609#diff-354f30a63fb0907d4ad57269548329e3 and then make any changes we need on top of that.

bogren:
> I intended to restore the Gemfile in a follow-up but... 😅 you know. 😆 
Hehe yes, that's why you open source ;) perfect for us who'd like to get started.

PR #869, opened 2017-07-27
Directly executes the batch update or reload data if the collection view is idle

ethanjin:
So the view controller can prepare a default page between viewDidLoad
and viewWillAppear by providing a default data to the collection view
and executing reloadData.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,8 @@ The changelog for `IGListKit`. Also see the [releases](https://github.com/instag
 
 - Added `-[IGListSectionController didDeselectItemAtIndex:]` API to support default `UICollectionView` cell deselection. [Ryan Nystrom](https://github.com/rnystrom) (tbd)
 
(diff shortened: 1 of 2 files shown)


rnystrom:
@ethanjin curious, what was this trying to solve? Also why close it?

ethanjin:
@rnystrom Sorry I‘m not very familiar with the PR process. I would like to refine the tests for there're still a few errors, so I closed this before everything is fine.

The problem I am trying to solve is, when I want to prepare a default page during 'viedDidLoad' (e.g., using cached data), and start requesting new data with an animation at 'viewWillAppear', I need to execute 'reloadData' immediately without waiting for the GCD queue. And I think this change won't cause problems.

rnystrom:
@ethanjin 

1. Don't worry about leaving PRs open as "work in progress". Just put WIP in the title and ppl will leave it alone. Can be a good way to track your work in the open!

2. `UICollectionView` will internally reload data when asked to layout for the first time, so what you want already happens. Do you have an example of the issue, e.g. example project or failing unit test?

PR #891, opened 2017-08-11
Remove SwiftLint

rnystrom:

Follow up from #837. Too much noise to keep working, and cost to maintain in an examples app is too high.


heumn:
are you open to me adding SwiftFormat in a PR? :)

rnystrom:
@heumn lets hold on that for now. Idk if there's a lot of value in adding/maintaining all these tools just for a sample app. Once we start using Swift in the main lib, then it'll be much much much more important (IMO).

heshamsalman:
@rnystrom LGTM as a solution. Could reducing the # of active rules also achieve the same effect?

heumn:
Got it and agreed :)

rnystrom:
@heshamsalman that's a good point. Maybe that'd be simpler than gutting it?

jessesquires:
I agree with @heshamsalman  -- I don't think we should just gut this. Let's relax the rules instead 😄

Issue #755, opened 2017-05-22
Add guide to using IGListBindingSectionController


rnystrom:
- Theory (one-way data flow)
- Visuals
- Model setup
- Common usages

shuhrat10:
Will nice to have example with lazy loading for top section. For example: At the top we will horizontal CollectionView with categories, at the bottom we will have content for selected categories with lazy loading.

Issue #799, opened 2017-05-28
Invalid number of items in section


cikpis:
## New issue checklist

- [x] I have reviewed the [`README`](https://github.com/Instagram/IGListKit/blob/master/README.md) and [documentation](http://instagram.github.io/IGListKit)
- [x] I have searched [existing issues](https://github.com/Instagram/IGListKit/issues) and this is not a duplicate

### General information

- `IGListKit` version: 3.0.0
- iOS version(s): 10.3.2
- CocoaPods/Carthage version: CocoaPods 1.2.1
- Xcode version: AppCode 2017.1.2
- Devices/Simulators affected: Device iPhone 6S
- Reproducible in the demo project? (Yes/No): Yes
- Related issues: None

### Debug information
Hi guys! Basically I'm stuck with one crash and I'm not getting why.

**Model**
```swift
import Foundation
import IGListKit

class CategoryDetailsSection: NSObject {
    var images: [Image]
    var offSet: OffSet

    init(withImages images: [Image], offSet: OffSet) {
        self.images = images
        self.offSet = offSet
        super.init()
    }
}

// MARK: - ListDiffable
extension CategoryDetailsSection: ListDiffable {

    func diffIdentifier() -> NSObjectProtocol {
        return self
    }

    func isEqual(toDiffableObject object: ListDiffable?) -> Bool {
        return isEqual(object)
    }
}
```

**In ViewController I have the an endpoint response that I need to display**
```swift
func displayGetCategoryDetailsSuccess(viewModel: CategoryDetailsModel.GetCategoryDetails.ViewModel) {
    activityView.stopAnimating()
    guard case .success(let categoryDetailsSection) = viewModel.resultData else { return }
    self.categoryDetailsSection.offSet = categoryDetailsSection.offSet
    self.categoryDetailsSection.images.append(contentsOf: categoryDetailsSection.images)
    adapter.performUpdates(animated: true, completion: nil)
}
```

**ListAdapterDataSource is:**
```swift
extension CategoryDetailsViewController: ListAdapterDataSource {

    func objects(for listAdapter: ListAdapter) -> [ListDiffable] {
        return [categoryDetailsSection]
    }

    func listAdapter(_ listAdapter: ListAdapter, sectionControllerFor object: Any) -> ListSectionController {
        return CategoryDetailsSectionController()
    }

    func emptyView(for listAdapter: ListAdapter) -> UIView? { return nil }
}
```
[...]

Issue #697, opened 2017-04-24
Create an IGListBindingSectionController guide


rnystrom:
Inspired from #696, there are some "tricks" to using it correctly. Should write up some documentation and diagrams to compliment the calendar example.

rnystrom:
Gonna bite the bullet and get this done!

Issue #644, opened 2017-04-15
The position of the scroll indicator when prepending data


yusuftor:
I have a view that is like the comments system in instagram, where the newest data appears at the bottom of the collection view. As I scroll up, I get the next page of data and call performUpdates() to prepend the new data to my existing data. However, when the collection view adds on this extra data, the scroll indicator is still right at the top of the collection view instead of at the bottom of the new data. This can be visualised like this:

```
UICollectionView cells with data: 

- data <- Scroll indicator scrolled to the top of the collectionview and new data starts to load
- data
- data
- data <- Scroll indicator starts here at the bottom of the collectionview

Scrolling up does this:

- new data <- Scroll indicator stays at the top and the new data is inserted
- new data
- new data
- new data
- data
- data
- data
- data 

However it should look like this:

- new data 
- new data
- new data
- new data 
- data <- Scroll indicator
- data
- data
- data 

```

How can I change that?

rnystrom:
Tracking this in #242 right now. We do this in Instagram in a few places and are planning on open sourcing how we do it, just lower priority atm. If you want to get it working ASAP:

- Get the topmost visible cell's frame using the `UICollectionViewLayout`
- `performUpdates(animated: false)`
- In the completion block, find the same cell's frame via `UICollectionViewLayout`
- Calculate the delta between the old and new frame, apply to `contentOffset`

yusuftor:
Oh yes!! Thank you :)

darrarski:
@yusuftor not sure if you manage to solve your problem, but you can check my approach, which doesn't require to save visible cell [here](https://github.com/Instagram/IGListKit/issues/242#issuecomment-296863521).

yusuftor:
Thanks @darrarski! I couldn't get it to work the way Ryan said, so instead I calculated the difference in contentSize before and after the update and applied that to the contentOffset. I will also try your way to see if it's any better :)

pkuklis:
Any updates on this issue?

I'm able to scroll to the correct offset but it always has a little stutter/visual glitch.

hahtml:
@pkuklis Do u solve the stutter/visual glitch issue. I implement in the same way above and also have glitch issue. Cannot figure out how to solve it.

pkuklis:
> @pkuklis Do u solve the stutter/visual glitch issue. I implement in the same way above and also have glitch issue. Cannot figure out how to solve it.

Unfortunately not

hahtml:
[...]

PR #903, opened 2017-08-18
Create binding and modeling guide/tutorial

rnystrom:

Adds a tutorial-style guide on how to approach modeling given a design, how to effectively use `IGListBindingSectionController`, and how to do local mutations/updates.

You can read the [formatted guide here](https://github.com/Instagram/IGListKit/blob/binding-guide/Guides/Modeling%20and%20Binding.md).

Issue fixed: #755, #799, #697


Resources/ModelingAndBinding-Starter/Podfile
```diff
@@ -0,0 +1,11 @@
+# Uncomment the next line to define a global platform for your project
+platform :ios, '10.0'
+
+use_frameworks!
+
+pod 'IGListKit', :git => 'https://github.com/Instagram/IGListKit.git', :branch => 'master'
+pod 'SDWebImage'
```
jessesquires: can we avoid this dep?

if not, let's pin it to a specific version to avoid future breakage

Resources/ModelingAndBinding-Starter/ModelingAndBinding/ActionCell.swift
```diff
@@ -0,0 +1,16 @@
+//
+//  ActionCell.swift
+//  ModelingAndBinding
+//
+//  Created by Ryan Nystrom on 8/18/17.
+//  Copyright © 2017 Ryan Nystrom. All rights reserved.
+//
```
jessesquires: comment headers?

Resources/ModelingAndBinding-Starter/ModelingAndBinding/AppDelegate.swift
```diff
@@ -0,0 +1,46 @@
...
+        // Called as part of the transition from the background to the active state; here you can undo many of the changes made on entering the background.
+    }
+
+    func applicationDidBecomeActive(_ application: UIApplication) {
+        // Restart any tasks that were paused (or not yet started) while the application was inactive. If the application was previously in the background, optionally refresh the user interface.
+    }
+
+    func applicationWillTerminate(_ application: UIApplication) {
+        // Called when the application is about to terminate. Save data if appropriate. See also applicationDidEnterBackground:.
+    }
```
jessesquires: remove all boilerplate?

Resources/ModelingAndBinding-Starter/ModelingAndBinding/ActionCell.swift
rnystrom: Duh ya

Resources/ModelingAndBinding-Starter/Podfile
rnystrom: Ya, we can. I don't want to promote bad practices, but its probably fine to just hack it and download here

Resources/ModelingAndBinding-Starter/Podfile
```diff
@@ -0,0 +1,11 @@
+# Uncomment the next line to define a global platform for your project
+platform :ios, '10.0'
+
+use_frameworks!
+
+pod 'IGListKit', :git => 'https://github.com/Instagram/IGListKit.git', :branch => 'master'
```
jessesquires: this needs to reference the local spec. `../../IGListKit.podspec`

otherwise, we've duped all source into this directory 😅 

why not put this stuff in the existing `Examples/` dir?

rnystrom: Lol ya duh.

There's 2 versions of them: the starter and the finished. Not really an example, more like a working project for someone to go along w/ the guide.

rnystrom:
[Read draft of guide here](https://github.com/Instagram/IGListKit/blob/binding-guide/Guides/Modeling%20and%20Binding.md)

Resources/ModelingAndBinding-Starter/Podfile
jessesquires: Ahh. I see. Maybe we keep the markdown guide here and all the code, etc. somewhere else?

Starting to feel a bit bloated in this repo. Could we put the projects under `rnystrom/IGListKit-binding-guide` or something?

rnystrom:
Added link to view guide in branch

Resources/ModelingAndBinding-Starter/Podfile
rnystrom: I dig it! I'll move there

rnystrom:
@cikpis would love your feedback on this!

jessesquires:
nice! 👍 💯 

did you pngcrush these assets? 😄

jessesquires:
#m lgtm-nystrom 😏

rnystrom:
Oops nope, gotta save them bytes. Will crush.

cikpis:
@rnystrom Hey man! Thanks for it. Right now I cannot look at it because I'm on holidays. Yes, sometimes devs get holidays xD

I'll take a look next week! Thanks for the effort :)

Guides/Modeling and Binding.md
```diff
@@ -0,0 +1,584 @@
...
+  }
+
+  // MARK: ListDiffable
+
+  func diffIdentifier() -> NSObjectProtocol {
+    return "image" as NSObjectProtocol
+  }
+
+  func isEqual(toDiffableObject object: ListDiffable?) -> Bool {
+    guard let object = object as? ImageViewModel else { return false }
```
heumn: Could I propose that you suggest that they `fatalError("expected ImageViewModel but got something else")` (wording?) here? Something is _seriously_ wrong in their setup of they send the wrong ViewModel to a cell.

Guides/Modeling and Binding.md
```diff
@@ -0,0 +1,584 @@
...
+```swift
+final class ActionCell: UICollectionViewCell, ListBindable {
+
+  @IBOutlet weak var likesLabel: UILabel!
+  @IBOutlet weak var likeButton: UIButton!
+
+  // MARK: ListBindable
+
+  func bindViewModel(_ viewModel: Any) {
+    guard let viewModel = viewModel as? ActionViewModel else { return }
```
heumn: If you agree, same goes for these ones (wrong viewModel received) :)

Guides/Modeling and Binding.md
```diff
@@ -0,0 +1,584 @@
...
+```swift
+protocol ActionCellDelegate: class {
+  func didTapHeart(cell: ActionCell)
+}
+```
+
+Add a new delegate variable to the `ActionCell`, beneath the outlets:
+
+```swift
+weak var delegate: ActionCellDelegate? = nil
```
heumn: Nit: no need for the `= nil` here :)

heumn:
This is _such_ a great tutorial. I would have loved this one back when I was familiarizing myself with IGListKit 💯  and `IGListBindingSectionController` is my favorite part of IGListKit now 👌 

jzaZ23z45UxK8

Guides/Modeling and Binding.md
```diff
@@ -0,0 +1,584 @@
...
+```
+
+1. Make sure to import `IGListKit`!
+2. Have the cell conform to `ListBindable`
+3. Guard against the view model type. This will always be what `PostSectionController` decides, but guarding is a safer practive than force casting.
+4. Use the [SDWebImage](https://github.com/rs/SDWebImage) library to set the image URL.
+
+Now do exactly the same thing for each of the other cells:
+
+- `ActionCell` binds `ActionViewModel`
```
weyert: You are repeating yourself here ;) I think you meant t fallout the other models?

weyert:
Interesting reading, thanks @rnystrom!
I was wondering would be useful to add an example of self sizing cells in this guide? E.g for the Comment cell maybe?

rnystrom:
@weyert self sizing cells are loaded with complications/bugs and atm we're not able to provide first class support 😔

weyert:
I am noticing that at the moment experiencing some weirdness with it. I assume the bugs/complications are related to UICollectionView itself? Anyways, it was just a suggestion :-)

rnystrom:
@jessesquires gonna spell check this then land if you're good?

Guides/Modeling and Binding.md
```diff
@@ -0,0 +1,584 @@
...
+You can already start mentally modelling your data:
+
+- The top cell has a **username** and **timestamp** label
+- The image cell will need some sort of image `URL`
+- An action cell with **like count**. There will also need to be some sort of action handling when someone taps the heart
+- Then there are a _dynamic_ number of comment cells that contain a **username** and **comment**
+
+Remember that `IGListKit` functions on **one model per section controller**. All of the cells in this design correlate to one top-level "post" object delivered by a server. You want to create a `Post` model that contains all of the information that the cells require.
+
+> A common mistake is to create a single model and section controller for a single cell. In this example, that will create a **very confusing** architecture since the top-level objects will contain a mix and match of user, image, and action, and comment models.
```
BasThomas: `will contain a mix and match of user, image, action, and comment models.` (remove one "and")

Guides/Modeling and Binding.md
```diff
@@ -0,0 +1,584 @@
...
+    self.imageURL = imageURL
+    self.likes = likes
+    self.comments = comments
+  }
+
+}
+```
+
+1. It's best practice to always declare your values as `let` so they cannot be mutated again. The compiler will complain about the `Comment` model, ignore that for now.
+2. Since `IGListKit` is Objective-C compatible, your models must be `class`es which means no free initializer. It's only a little copy & paste!
```
BasThomas: `is compatible with Objective-C` reads a bit better imho

Guides/Modeling and Binding.md
```diff
@@ -0,0 +1,584 @@
...
+
+Now add an `ListDiffable` implementation inside of `Post`:
```
BasThomas: `a`

Guides/Modeling and Binding.md
```diff
@@ -0,0 +1,584 @@
...
+
+In this case, the `username` and `text` **must be equal** by the time two objects are checked for equality.
+
+</p></details>
+
+Using the `Comment` array on a `Post` should make some sense: there are a dynamic number of comments on each post. For each comment, you want to display a cell.
+
+What might be a little bit of a new concept, though, is that you need to create models for the `UserCell`, `ImageCell`, _and_ `ActionCell` as well when working with `IGListBindingSectionController`.
+
+> A binding section controller is almost like a mini-`IGListKit`. It takes an array of view models and turns them into configured cells. Get into the habit of creating a new model for each cell type withing an `IGListBindingSectionController` instance.
```
BasThomas: `withing` -> `within`

cikpis:
Hi @rnystrom, thanks for the guide. I had some time to read it and seems ok-ish :) 
The only thing that still is not answering my questions is: How can I add, in base the content of your guide, a new comment? 

Also, seems you are calling `update(animated: true)`, isn't it reloading the whole/partial Controller instead of reloading just the cell like cell?

heumn:
> The only thing that still is not answering my questions is: How can I add, in base the content of your guide, a new comment?

Something changes (a new json response? user has typed a new comment? and the Post updates (gets re-created) and then you simply call perfor `performUpdates`

skjermbilde 2017-08-22 kl 00 23 25

I even made a demo 😄 

https://www.dropbox.com/s/8v28xbcuff6sjqi/update.mov?dl=0

cikpis:
Hi @heumn, you can find [here](https://github.com/Instagram/IGListKit/issues/799) what I mean about the `update(animated: true)`.

We have always been using `insert, delete, move` in any collection/table views. In fact in that link there is a comment from me with a code of how I achieved it. So, what I would like to know is: Why arem't we using `batchContext.insert(in:, at:)` way which should just update that part of the controller instead of call `update(animated: true)` which is updating the whole/partial controller?

heumn:
> Why arem't we using batchContext.insert(in:, at:) way which should just update that part of the controller instead of call update(animated: true) which is updating the whole/partial controller?

Because why would you _want_ to do that work (figuring out _what_ _where_ and _how_ something changes)? And even if you do, why would you move beyond the realms of just updating the model and call it a day (with calling `performUpdates`)? It seems like you are approaching the problem from the wrong end.

This is why IGListKit is so elegant. You only ever (almost) serve it immutable objects (section objects/viewModels), explain it who should be responsible for them (section controllers), and how they should map them to cells.

I might have misunderstood you but it seems like you are asking "but I want to do it like I did in UICollectionView delegates and datasource previously, why can I not?". First and formost because that is not how IGListKit was designed. But more importantly because that is a difficult task when the project grows and you are dealing with exponential complexity and tons of different models. This is what IGListKit abstracts away for you.

cikpis:
So in base of what you said, `batchContext.insert(in:, at:)` should be not visible to devs, otherwise I don't get why that method is there and we should not use it. Maybe for that I got confused :)

Also, again, if I recall properly, when I call `performUpdates` or `update(animated:)` after user has scrolled the list, the controller load all and the list start to the top again. Is still the same?

heumn:
> So in base of what you said, batchContext.insert(in:, at:) shouldn't be visible to devs, otherwise I don't get why that method is there and we should not use it. Maybe for that I got confused :)

I think it would be more easy for me to propose a solution to your problem if you could provide me with a dummy project (fork an example project?) and explain to me in detail why and why you want to do something :)

> Also, again, if I recall properly, when I call performUpdates or update(animated:) after user has scrolled the list, the controller load all and the list start to the top again. Is still the same?

Good question and I think this is something that is being worked on 
https://github.com/Instagram/IGListKit/issues/242

also if you search a little it has been discussed (with different solutions) around in the repo.

Or you can open a PR for it :)

rnystrom:
@cikpis the update method on the binding controller performs a diff on its view models and then applies insert+delete+move+reload _on the cells_. Also, the binding SC will diff when the top level model changes and you call `adapter.performUpdates()` (so you can append new comments at the view controller or data source).

Instagram has been doing cell-level updates since before IGListKit was open sources, I can guarantee that cell insert+delete+move+reload works great 😉 

I'll update the guide to make it even more clear how easy changing the data is!

rnystrom:
Also +1 @heumn'a suggestion to search for scrolling solutions, that issue has been pretty well discussed. We don't have a first-class solution built in, but that hasn't prevented anyone from doing it.

cikpis:
@rnystrom nice! so you are telling me that calling `adapter.performUpdates()` is going to perform an `insert+delete+move+reload` without affect the scroll? Am I right?

rnystrom:
@cikpis yes and no: calling `performUpdates()` does insert+delete+move+reload, but IGListKit doesn't do any scroll-position-pinning on updates. Check out [this issue](https://github.com/Instagram/IGListKit/issues/644) for an explanation of how we fix scroll positions in Instagram during updates.

cikpis:
What I meant if is the scroll is gonna go up ( not to the new items ) like the whole view is reloaded after calling `performUpdates()`, or if the scroll stay where it is. Scrolling to the new item is not what I'm asking. I wanna make sure that after I call `performUpdates()` all the view/scroll maintain the same status/scroll.

rnystrom:
Yup, that issue is describing exactly how to maintain scroll position, like you're asking. 
If you don't do any special handling, the offset will change when new items are inserted off screen. This is exactly what `UICollectionView`'s default behavior is atm.

cikpis:
Nice! I'll make some test when I'll be back operative :)

Thanks again fro your time, cheers!

PR #922, opened 2017-09-05
Fixed Search controller section in example project

RishabhTayal:
The search bar used to set the text as empty when the user scrolls.

### DIFF
--- a/Examples/Examples-iOS/IGListKitExamples.xcodeproj/project.pbxproj
+++ b/Examples/Examples-iOS/IGListKitExamples.xcodeproj/project.pbxproj
@@ -1062,7 +1062,7 @@
 				"CODE_SIGN_IDENTITY[sdk=iphoneos*]" = "iPhone Developer";
 				DEVELOPMENT_TEAM = "";
 				INFOPLIST_FILE = IGListKitExamples/Info.plist;
-				IPHONEOS_DEPLOYMENT_TARGET = 10.0;
+				IPHONEOS_DEPLOYMENT_TARGET = 8.0;
 				LD_RUNPATH_SEARCH_PATHS = "$(inherited) @executable_path/Frameworks";
 				PRODUCT_BUNDLE_IDENTIFIER = com.instagram.IGListKitExamples;
 				PRODUCT_NAME = "$(TARGET_NAME)";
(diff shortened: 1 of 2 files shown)


Examples/Examples-iOS/IGListKitExamples.xcodeproj/project.pbxproj
```diff
@@ -1062,7 +1062,7 @@
 				"CODE_SIGN_IDENTITY[sdk=iphoneos*]" = "iPhone Developer";
 				DEVELOPMENT_TEAM = "";
 				INFOPLIST_FILE = IGListKitExamples/Info.plist;
-				IPHONEOS_DEPLOYMENT_TARGET = 8.0;
+				IPHONEOS_DEPLOYMENT_TARGET = 10.0;
```
rnystrom: Why change this? We ship to iOS 8, so supporting min 8 in the examples seems ok to me.

Examples/Examples-iOS/IGListKitExamples/SectionControllers/SearchSectionController.swift
```diff
@@ -46,14 +46,13 @@ final class SearchSectionController: ListSectionController, UISearchBarDelegate,
     }
 
     func searchBarTextDidEndEditing(_ searchBar: UISearchBar) {
-        delegate?.searchSectionController(self, didChangeText: "")
+        delegate?.searchSectionController(self, didChangeText: searchBar.text!)
```
rnystrom: Can you give a little detail about why we want to make this change?

RishabhTayal: If the user starts scrolling, the `searchBar` will be resigned. This will trigger the `searchBarTextDidEndEditing` function. Now since we are calling the delegate function with hard coded empty string it will show all the items in the list.

Examples/Examples-iOS/IGListKitExamples.xcodeproj/project.pbxproj
RishabhTayal: I agree. I wasn't sure if we need to change it to iOS 10. But when I compile the example project in Xcode 9 beta 6, it gives me compile error. I can revert this change back if you want.

rnystrom:
Ah ok, cool. Thanks @RishabhTayal! We'll just need you to sign the CLA then we can import this!

RishabhTayal:
Done

PR #930, opened 2017-09-13
Weak reference collection view when queueing update

rnystrom:

Found a "bug" when adding some new unit tests. The queue method holds a strong ref to the collection view, and doesn't release it until that block is executed. The adapter may have already been released, holding onto the collection view is wasteful.

### Checklist

### DIFF
--- a/Source/IGListAdapterUpdater.m
+++ b/Source/IGListAdapterUpdater.m
@@ -380,6 +380,12 @@ void convertReloadToDeleteInsert(NSMutableIndexSet *reloads,
     }
 
     __weak __typeof__(self) weakSelf = self;
+    __weak __typeof__(collectionView) weakCollectionView = collectionView;
+
+    // dispatch_async to give the main queue time to collect more batch updates so that a minimum amount of work
+    // (diffing, etc) is done on main. dispatch_async does not garauntee a full runloop turn will pass though.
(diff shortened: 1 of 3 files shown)


Source/IGListAdapterUpdater.m
```diff
@@ -380,16 +380,22 @@ - (void)queueUpdateWithCollectionView:(UICollectionView *)collectionView {
     }
 
     __weak __typeof__(self) weakSelf = self;
+    __weak __typeof__(collectionView) weakCollectionView = collectionView;
+
+    // dispatch_async to give the main queue time to collect more batch updates so that a minimum amount of work
+    // (diffing, etc) is done on main. dispatch_async does not garauntee a full runloop turn will pass though.
+    // see -performUpdateWithCollectionView:fromObjects:toObjects:animated:]objectTransitionBlock:completion: for more
+    // details on how coalescence is done.
```
rnystrom: Added this b/c I keep getting questions about this

jessesquires:
🤔 shouldn't danger have warned about a missing milestone?

Source/IGListAdapterUpdater.m
weyert: guarantee?

rnystrom: I will never, ever be able to spell this word right on the first try.

Issue #929, opened 2017-09-13
Invalidating layout while in the middle of an update can cause a crash


rnystrom:
I can repro this on my [GitHawk](https://github.com/rnystrom/githawk) app pretty easily, but I'm having trouble recreating the crash in a unit test. The gist of the bug is that you can [invalidate](https://github.com/Instagram/IGListKit/blob/master/Source/IGListAdapter.m#L993-L1012) which calls `performBatchUpdates:completion:` while in the middle of a batch update block already. This needs to be queued and unloaded once the current batch update is finished.

Some ideas:

1. Just call the existing `performBatchAnimated:updates:completion:` API
  - That takes an `animated` flag, which breaks the ability to wrap `invalidateLayoutForSectionController:` in things like the spring API (but not crashing > springs)
2. Add a mechanism to queue completion blocks if inside a batch update block, unloading on outer update completion
  - Same problem, lose animation wrapper if already batch updating
  - However, if not already in update can just do it directly

Any thoughts?

PR #931, opened 2017-09-13
Defer invalidatation until all batch updates are finished

rnystrom:

Added a mechanism to `IGListAdapter` to defer (_queue or execute_) blocks around asking the updater to do batch updates. I discovered a crash where the `_updateCompletionHandler` (private ivar) of `UICollectionView` was not being niled when two batch updates collide. There is an assert in `-[UICollectionView dealloc]` that fires if this block != nil. Took some reverse eng to track down what is happening.

Sadly I cannot reproduce the issue in a unit test _or_ sample app. However I can 100% repro this in my GitHawk app. Applying this patch fixes the problem. I added a unit test that got me _close_ to the state of the crash. Consider this new test future-proofing (also covers the enter/exit/defer stuff).

Also dumping the `UICollectionView` internals confirms that the bad state (`_updateCompletionHandler != nil`) is no longer true.

Issue fixed: #929

Depends on #930

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,8 @@ The changelog for `IGListKit`. Also see the [releases](https://github.com/instag
 
 - Weakly reference the `UICollectionView` in coalescence so that it can be released if the rest of system is destroyed. [Ryan Nystrom](https://github.com/rnystrom) [(#tbd)](https://github.com/Instagram/IGListKit/pull/tbd)
 
+- Avoid crash when invalidating the layout while inside `-[UICollectionView performBatchUpdates:completion:]. [Ryan Nystrom](https://github.com/rnystrom) [(#tbd)](https://github.com/Instagram/IGListKit/pull/tbd)
+
 
 3.1.1
 -----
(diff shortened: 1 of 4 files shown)


weyert:
I am still quite new to iOS development. May I ask what you mean with 'dumping the UICollectionView internals confirms'. How would this work or is it done? Are you dumping the assembly?

rnystrom:
@weyert [chisel](https://github.com/facebook/chisel) for advanced lldb and [Hopper](https://www.hopperapp.com/) for disassembling (assembly to pseudocode).

CHANGELOG.md
```diff
@@ -2,6 +2,16 @@
...
 The changelog for `IGListKit`. Also see the [releases](https://github.com/instagram/IGListKit/releases) on GitHub.
 
+3.2.0 (upcoming release)
+-----
+
+### Fixes
+
+- Weakly reference the `UICollectionView` in coalescence so that it can be released if the rest of system is destroyed. [Ryan Nystrom](https://github.com/rnystrom) [(#tbd)](https://github.com/Instagram/IGListKit/pull/tbd)
+
+- Avoid crash when invalidating the layout while inside `-[UICollectionView performBatchUpdates:completion:]. [Ryan Nystrom](https://github.com/rnystrom) [(#tbd)](https://github.com/Instagram/IGListKit/pull/tbd)
```
jessesquires: missed backtick here `

Issue #932, opened 2017-09-13
Scrolling to object does not work with custom contentInset 


yusufoos:
## New issue checklist

- [YES] I have reviewed the [`README`](https://github.com/Instagram/IGListKit/blob/master/README.md) and [documentation](http://instagram.github.io/IGListKit)
- [YES] I have searched [existing issues](https://github.com/Instagram/IGListKit/issues) and this is not a duplicate

### General information

- `IGListKit` version: 3.0
- iOS version(s): 10
- CocoaPods/Carthage version:
- Xcode version: 8.3
- Devices/Simulators affected: all
- Reproducible in the demo project? (Yes/No): Yes
- Related issues: #563

Issue: 

Modifiying the MixedDataViewController in the examples I've added

`                self.automaticallyAdjustsScrollViewInsets = false
`
`        self.collectionView.contentInset = UIEdgeInsetsMake(200, 0, 0, 0)
`

And then later I call

```
DispatchQueue.main.asyncAfter(deadline: .now() + 1) {
        self.adapter.scroll(to: self.data.last!, supplementaryKinds: nil, scrollDirection: UICollectionViewScrollDirection.vertical, scrollPosition: UICollectionViewScrollPosition.bottom, animated: true)
        }
```

Expected Result: 
It correctly scrolls to the bottom of the last section.

Actual Result:
It scrolls down a little but not completely, it still needs to scroll down by the amount I specified in contentInset to be correct.

Full Code Example:

```
/**
 Copyright (c) 2016-present, Facebook, Inc. All rights reserved.

 The examples provided by Facebook are for non-commercial testing and evaluation
 purposes only. Facebook reserves all rights not expressly granted.

 THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
 IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
 FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL
 FACEBOOK BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN
 ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION
 WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
 */

import UIKit
[...]

PR #940, opened 2017-09-20
Fix scrollToObject: bug when scrolling to bottom with content inset

rnystrom:

This has been around forever, surprised we've never run into it. Instagram doesn't really use the `bottom` position except for one place that is already at the bottom of a view. Looks like this bug has just been hanging around quietly for ages.

`UIScrollView` takes into account the content inset when changing the content offset. Doing math w/ the inset is both useless and incorrect.

Issue fixed: #932

### Checklist

### DIFF
--- a/Source/IGListAdapter.m
+++ b/Source/IGListAdapter.m
@@ -273,7 +273,7 @@
         case UICollectionViewScrollDirectionVertical: {
             switch (scrollPosition) {
                 case UICollectionViewScrollPositionBottom:
-                    contentOffset.y = offsetMax - collectionViewHeight - contentInset.top;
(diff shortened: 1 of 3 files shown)


Source/IGListAdapter.m
```diff
@@ -273,7 +273,7 @@ - (void)scrollToObject:(id)object
         case UICollectionViewScrollDirectionVertical: {
             switch (scrollPosition) {
                 case UICollectionViewScrollPositionBottom:
-                    contentOffset.y = offsetMax - collectionViewHeight - contentInset.top;
```
jessesquires: lol

PR #941, opened 2017-09-20
[iOS11] Update example project files, convert to Swift 4, and fix errors

rnystrom:

Used migration tool with Swift 4, fixed build errors, and killed all warnings. Only non-trivial change was in the macOS example. Will flag that one.

No changelog needed for this, right?

### Checklist


Examples/Examples-tvOS/IGListKitExamples/Views/LabelCell.swift
```diff
@@ -25,7 +25,7 @@ final class LabelCell: UICollectionViewCell {
 
     static func textHeight(_ text: String, width: CGFloat) -> CGFloat {
         let constrainedSize = CGSize(width: width - insets.left - insets.right, height: CGFloat.greatestFiniteMagnitude)
-        let attributes = [ NSFontAttributeName: font ]
+        let attributes = [ NSAttributedStringKey.font: font ]
```
Sherlouk: Should be able to drop enum type here and let it infer it `.font:`

Examples/Examples-iOS/IGListKitMessageExample/MessagesViewController.swift
```diff
@@ -16,6 +16,7 @@ import UIKit
 import Messages
 import IGListKit
 
+@available(iOSApplicationExtension 10.0, *)
```
Sherlouk: Do we know why this is suddenly required? Assuming the target version hasn't been changed?

Examples/Examples-iOS/IGListKitExamples/Views/LabelCell.swift
```diff
@@ -26,7 +26,7 @@ final class LabelCell: UICollectionViewCell {
 
     static func textHeight(_ text: String, width: CGFloat) -> CGFloat {
         let constrainedSize = CGSize(width: width - insets.left - insets.right, height: CGFloat.greatestFiniteMagnitude)
-        let attributes = [ NSFontAttributeName: font ]
+        let attributes = [ NSAttributedStringKey.font: font ]
```
Sherlouk: Infer enum `.font`

Examples/Examples-iOS/IGListKitExamples/Systems/IncrementAnnouncer.swift
```diff
@@ -14,8 +14,7 @@
 
 import UIKit
 
-@objc
-protocol IncrementListener: class {
+@objc protocol IncrementListener: class {
```
Sherlouk: Does this need to have `@objc` for the single ObjC example we have?

Examples/Examples-tvOS/IGListKitExamples/Views/LabelCell.swift
rnystrom: Nice, ya this was the migration tool

Examples/Examples-iOS/IGListKitMessageExample/MessagesViewController.swift
rnystrom: No idea tbh, failed to build so just clicked "fix" haha

Examples/Examples-iOS/IGListKitExamples/Systems/IncrementAnnouncer.swift
rnystrom: It does b/c we're using it in `NSHashTable`, can only put `NSObjectProtocol` stuff in there.

Examples/Examples-iOS/IGListKitExamples/Views/LabelCell.swift
jessesquires: this won't work unless you explicitly add a type annotation:

`let attributes: [NSAttributedStringKey: Any] = [ .font: font]`

"swifty" style is to omit types, though, so probably leaves this as is.

Sherlouk: It would work if the attributes was inline with the actual initialiser which given the length of text probably wouldn't be bad

Not fussed either way

jessesquires: Ah yes, 👍👍👍

rnystrom:
Build failures fixed in #946

PR #943, opened 2017-09-20
[iOS11] Inhibit all warnings from OCMock #trivial

rnystrom:
Removing project warnings from third party libs. Mostly from iOS 11 changes.


Podfile
```diff
@@ -4,9 +4,9 @@ use_frameworks!
 workspace 'IGListKit'
```
Sherlouk: FWIW `inhibit_all_warnings!` once at the top should have the same effect

Issue #870, opened 2017-07-27
[RFC] Provide UIKit bindings for diffs *outside* of the SectionController + ListAdapter flow


jessesquires:
## Provide UIKit bindings for diffs *outside* of the SectionController + ListAdapter flow

One barrier to using IGLK is adopting the `SectionController` + `ListAdapter` architecture, which is difficult for legacy projects, or projects that have adopted other "list" abstractions but now want to implement diffing for those lists.

Currently, we provide a `Diffing` subspec, but this requires clients to write bindings to `UICollectionView` and `UITableView` themselves. IGLK should provide this instead.

### Proposal

Provide extension methods on `UITableView` and `UICollectionView` that will perform updates on the view given the results from the [IGListDiff](https://github.com/Instagram/IGListKit/blob/master/Source/Common/IGListDiff.h) C functions. Essentially, we want to expose the functionality of [IGListAdapterUpdater](https://github.com/Instagram/IGListKit/blob/master/Source/IGListAdapterUpdater.m) in way that allows clients to perform updates *without* using `SectionController`s and `ListAdapter`s.

### Pseudo example

```swift
let old = //...

let new = //....

let diffResults = ListDiff(oldArray:old newArray:new option:IGListDiffEquality)

collectionView.performBatchUpdates(diffResults: diffResults, completion: {
    // completion block
})

// table view follows similarly
```

rnystrom:
So on board with this. Will be trivial.

@jessesquires what do you think about creating a "light" adapter that does:

- Give it an array of diffable models, diffs+applies update
- One section, array powers `numberOfItems`
- `IGListBindable` cells, automatically binds view models

This would solve our `UITableView` issue without refactoring the world (I still think full table support will open a world of hurt).

Maybe we solve "apply diff update" category first then this?

weyert:
Sounds reasonable to my ears, @rnystrom :)

jessesquires:
@rnystrom -- 💯 yeah, let's do the simpler, "least intrusive" solution first

Issue #691, opened 2017-04-23
[AppleBug] UICollectionView batch update heap corruption


rnystrom:
Summary:
When performing a particular set of batch updates on a UICollectionView with a certain number of sections, it appears there is a heap corruption and crash in objc_release in -`[UICollectionViewUpdate _computeItemUpdates]`.

Steps to Reproduce:
1. Setup a UICollectionView with 8 sections, each with 3 items
2. Perform batch updates with:
  a. Insert item at section 1, item 0
  b. Insert section 0
  c. Move section 3 to section 1
3. Make sure data source is properly updated to reflect changes

Expected Results:
UICollectionView performs the queued updates on the view hierarchy.

Actual Results:
UICollectionView crashes with the following trace:

```
#0	0x0000000101a3bd7b in objc_release ()
#1	0x000000010325241c in -[UICollectionViewUpdate _computeItemUpdates] ()
#2	0x0000000103251315 in -[UICollectionViewUpdate initWithCollectionView:updateItems:oldModel:newModel:oldVisibleBounds:newVisibleBounds:] ()
#3	0x00000001032c078b in -[UICollectionView _endItemAnimationsWithInvalidationContext:tentativelyForReordering:] ()
#4	0x00000001032c8497 in -[UICollectionView _performBatchUpdates:completion:invalidationContext:tentativelyForReordering:] ()
#5	0x00000001032c82d5 in -[UICollectionView _performBatchUpdates:completion:invalidationContext:] ()
#6	0x00000001032c8278 in -[UICollectionView performBatchUpdates:completion:] ()
```

Version:
Crash reports from production show issue occurring on iOS 7.1.2, 8.3, 8.4, 9.1, 9.2, 9.2.1, 9.3, and 9.3.1. Locally reproduce on iOS 9.3.1 and Xcode 7.3 simulators.

Notes:
Changing from 8 to 9 sections in the sample project seems to mitigate the issue. The same happens when using only 7 sections.

Configuration:
All iPhones and iOS Simulators

[UICollectionViewHeapCorruption.zip](https://github.com/Instagram/IGListKit/files/949777/UICollectionViewHeapCorruption.zip)

rnystrom:
Please dupe

https://openradar.appspot.com/radar?id=6070670354022400

Issue #690, opened 2017-04-23
[AppleBug] UICollectionView reload displaced section crashes in batch update


rnystrom:
Summary:
When reloading a section that is displaced/shifted inside `-[UICollectionView performBatchUpdates:completion:]`, it is possible to hit an `NSInternalInconsistencyException` stating "attempt to perform an insert and a move to the same section (i)" when the sections used in `insertSections:`, `deleteSections:`, `reloadSections:` , and `moveSection:toSection:` do not intersect.

For instance, a collection view who's data source is updated to reflect the following section mutations will crash:

```
[self.collectionView performBatchUpdates:^{
  [self.collectionView moveSection:2 toSection:4];
  [self.collectionView reloadSections:[NSIndexSet indexSetWithIndex:3]];
  [self.collectionView deleteSections:[NSIndexSet indexSetWithIndex:4]];
  [self.collectionView insertSections:[NSIndexSet indexSetWithIndex:2]];
} completion:nil];
```

Steps to Reproduce:
1. Open attached UICollectionViewReloadDisplacement sample app
2. Hit "Play" bar button
3. Crash

Expected Results:
- Section 2 moves to section 4
- Section 3 is reloaded
- Section 4 is deleted
- A new section is inserted into section 2

Actual Results:
Crash with exception:

```
*** Terminating app due to uncaught exception 'NSInternalInconsistencyException', reason: 'attempt to perform an insert and a move to the same section (4)'
```

Version:
iOS 9.3.4, 8.4, and 10.0 beta 4

Configuration:
iPhone 6, iPhone 6s simulator

rnystrom:
Please dupe

https://openradar.appspot.com/radar?id=5037608161771520

PeterAntonsen:
I think there's something about, deletes and reloads using the index'es of the old-dataset, whereas inserts, uses the index'es of the new-dataset.. so you might want to check up on your relevant indexes as I think this is your issue..

rnystrom:
@PeterAntonsen this is actually a confirmed issue

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

PeterAntonsen:
@rnystrom, fair enough, just stumbled across this, weighing in from memory as it reminded of another thread I saw.. hope it gets solved :)

PR #948, opened 2017-09-22
[WIP] Provide UICollectionView category to apply diff results

rnystrom:

@jessesquires is this the sort of thing you had in mind for #870?

I started taking some notes on how to merge this all together so that there's a single entry point into batch updating a `UICollectionView` so that this can be shared w/ `IGListAdapterUpdater`. That's my preferred approach, but it's a bit complicated. If we're going to get more serious about this, I'll explore that path. 

My biggest fear is that we wont use this category, and that supporting it will be a monster PITA. If I can centralize the codepaths then it'll be supported "for free".

### Checklist

### DIFF
--- a/IGListKit.xcodeproj/project.pbxproj
+++ b/IGListKit.xcodeproj/project.pbxproj
@@ -196,6 +196,8 @@
 		294652BB1EA927750063BDD9 /* IGListSectionMap+DebugDescription.h in Headers */ = {isa = PBXBuildFile; fileRef = 290DF36E1E931457009FE456 /* IGListSectionMap+DebugDescription.h */; settings = {ATTRIBUTES = (Private, ); }; };
 		294652BC1EA927750063BDD9 /* UICollectionView+DebugDescription.h in Headers */ = {isa = PBXBuildFile; fileRef = 290DF3561E930CE2009FE456 /* UICollectionView+DebugDescription.h */; settings = {ATTRIBUTES = (Private, ); }; };
 		294AC6321DDE4C19002FCE5D /* IGListDiffResultTests.m in Sources */ = {isa = PBXBuildFile; fileRef = 294AC6311DDE4C19002FCE5D /* IGListDiffResultTests.m */; };
+		295A77B51F754BB3007BC403 /* UICollectionView+BatchUpdates.h in Headers */ = {isa = PBXBuildFile; fileRef = 295A77B31F754BB3007BC403 /* UICollectionView+BatchUpdates.h */; };
+		295A77B61F754BB3007BC403 /* UICollectionView+BatchUpdates.m in Sources */ = {isa = PBXBuildFile; fileRef = 295A77B41F754BB3007BC403 /* UICollectionView+BatchUpdates.m */; };
 		296AC95C1EA518D3005137E2 /* IGListReloadIndexPath.h in Headers */ = {isa = PBXBuildFile; fileRef = 296AC95A1EA518D3005137E2 /* IGListReloadIndexPath.h */; settings = {ATTRIBUTES = (Private, ); }; };
 		296AC95D1EA518D3005137E2 /* IGListReloadIndexPath.h in Headers */ = {isa = PBXBuildFile; fileRef = 296AC95A1EA518D3005137E2 /* IGListReloadIndexPath.h */; settings = {ATTRIBUTES = (Private, ); }; };
 		296AC95F1EA518D3005137E2 /* IGListReloadIndexPath.m in Sources */ = {isa = PBXBuildFile; fileRef = 296AC95B1EA518D3005137E2 /* IGListReloadIndexPath.m */; };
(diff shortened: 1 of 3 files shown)


rnystrom:
Not much activity, gonna let this die out. Can revive if this becomes more requested.

DreamingInBinary:
Hey hey @rnystrom - totally understand closing this one down. For our use case at Buffer this one is key, so I'm trying to roll it on my own. I've got a few questions, if that's cool.

1) By diving in more and more, it seems any updates should manifest themselves as deletes/inserts in `performBatchUpdates` - is that right? At first, I was allocating a `IGListBatchUpdateData` object for batch updates but hit exceptions are reloading items at index paths.

2) Our use case is diffing multiple sections, never just one. My flow right now is, diff the sections, loop through any updates that comes from it and snag a `IGListIndexPathResult` for each of them. Is that the sane way of thinking about it?

3) Just a quick one, in Apple's sample code I've come across a comment that mentions the "correct" way to update things in batch updates was Delete/Insert/Reload/Move. Though I can't specifically remember if this example was a table view, which is possible. In this code I noticed its Delete/Insert/Move.

I've got the exact code we're rolling with right now (some Buffer specific objects) - wondered if I'm on a good path forward? Totally no rush or expectation to respond, just thought I would pick your brain if possible 😄 

<hr /> 

```
- (void)reloadByDiffingOldUpdates:(NSArray<UpdatesCollectionSectionModel *> *)oldUpdates newUpdates:(NSArray<UpdatesCollectionSectionModel *> *)newUpdates diffCompleted:(void (^)(BOOL))diffCompleted completion:(void (^)(BOOL))completion {
    
    // Represents if there were any changes at all within sections
    IGListIndexSetResult *diffs = IGListDiff(oldUpdates, newUpdates, IGListDiffEquality);

    // Early out
    if (!diffs || !diffs.hasChanges) {
        if (diffCompleted) {
            diffCompleted(YES);
        }
        if(completion){
            completion(YES);
        }
        return;
    }
    
    // Now get the granular changes within each section
    NSMutableIndexSet *inserts = [diffs.inserts mutableCopy];
    NSMutableIndexSet *deletes = [diffs.deletes mutableCopy];
    NSSet *moves = [[NSSet alloc] initWithArray:diffs.moves];
    NSMutableIndexSet *updates = [diffs.updates mutableCopy];
    
    // Changes within the section, if applicable
    NSMutableArray<NSIndexPath *> *itemInserts = [NSMutableArray new];
    NSMutableArray<NSIndexPath *> *itemDeletes = [NSMutableArray new];
    NSMutableArray<IGListMoveIndexPath *> *itemMoves = [NSMutableArray new];
    
    // A section was updated, let's find out what happened
    [updates enumerateIndexesUsingBlock:^(NSUInteger oldIndex, BOOL *stop) {
        __block NSUInteger newIndex = NSNotFound;
        
        // Assign any old sections that were entirely moved and have a new section index? I might be off here
        [moves enumerateObjectsUsingBlock:^(IGListMoveIndex *moveIndex, BOOL *moveStop) {
            if (moveIndex.from == oldIndex) {
                newIndex = moveIndex.to;
                *moveStop = YES;
            }
        }];

        if (newIndex == NSNotFound) newIndex = oldIndex;
        
        IGListIndexPathResult *paths = IGListDiffPaths(oldIndex, newIndex, oldUpdates[oldIndex].updates, newUpdates[newIndex].updates, IGListDiffEquality);
        paths = [paths resultForBatchUpdates];
        [itemInserts addObjectsFromArray:paths.inserts];
        [itemDeletes addObjectsFromArray:paths.deletes];
        [itemMoves addObjectsFromArray:paths.moves];
    }];
    
    diffCompleted(YES);

    [self performBatchUpdates:^{
        
        // Sections
        [self insertSections:inserts];
        [self deleteSections:deletes];
        for (IGListMoveIndex *move in moves) {
            [self moveSection:move.from toSection:move.to];
        }
        
        // Index paths
        [self insertItemsAtIndexPaths:itemInserts];
        [self deleteItemsAtIndexPaths:itemDeletes];
        for (IGListMoveIndexPath *move in itemMoves) {
            [self moveItemAtIndexPath:move.from toIndexPath:move.to];
        }
        
    } completion:^ (BOOL done) {
        completion(YES);
    }];
}
```

rnystrom:
This is awesome! I'm glad someone (esp Buffer ❤️ ) is interested in this. I'll try and be as helpful as I can! Also very happy to make this a part of IGLK.

### 1.

> By diving in more and more, it seems any updates should manifest themselves as deletes/inserts in performBatchUpdates - is that right? At first, I was allocating a IGListBatchUpdateData object for batch updates but hit exceptions around reloading items at index paths (i.e. same index path was both a move and reload).

Yup, deletes+inserts+moves. 

I spent a **lot** of time investigating reloads. Originally reloads were supported, however at Instagram's scale we detected a significant number of crashes with exceptions like:

```
*** Terminating app due to uncaught exception 'NSInternalInconsistencyException', reason: 'attempt to perform an insert and a move to the same section (4)'
```

`UICollectionView`'s documentation is misleading:

> You can use this method in cases where you want to make multiple changes to the collection view in one single animated operation, as opposed to in several separate animations. You might use this method to **insert, delete, reload, or move** cells or use it to change the layout parameters associated with one or more cells.

This is true, but reload transitions are implemented strangely.

To prove the inconsistency in messaging, the delete+insert+move API [docs](https://developer.apple.com/documentation/uikit/uicollectionview/1618097-insertitemsatindexpaths?language=objc) have the line:

> You can also call this method from a block passed to the performBatchUpdates:completion: method ...

This line is curiously missing from [reload's docs](https://developer.apple.com/documentation/uikit/uicollectionview/1618055-reloaditemsatindexpaths?language=objc).

We've opened a radar which you can view (and dupe! 🙏 ) with #690. It includes a very basic sample app that repros this error. With this finding we decided that reloads inside batch updates are considered unstable and shouldn't be used.

#### 2.

> Our use case is diffing multiple sections, never just one. My flow right now is, diff the sections, loop through any updates that comes from it and snag a IGListIndexPathResult for each of them via its resultForBatchUpdates. Is that the sane way of thinking about it?

Yup, and your code is handling when sections are repositioned too, so you should be ok. Keep in mind that this is an _n_<sup>2</sup> operation tho. I've always wanted to write an algo that could take in 2d arrays, flatten to 1d, diff, then output a 2d result. Saving that for a rainy day (or someone else to contribute 😇 ).

But don't get bogged down w/ perf. If its a problem, diff on a bg queue. But as long as you aren't dealing w/ thousands of sections+items, you're fine.

Also, `UICollectionView` changes are orders-of-magnitude more expensive than our diffing algo. We're still testing the improvements from falling back to `reloadData` for monster changesets in 61c1524bb0c59d5eb43f65b02442ad50dfb2d0dc.

#### 3.

Basically see 1.

---

Some notes on the code:

- You probably want to use `resultForBatchUpdates` on the section changes too, those will have reloads as-is
- Some section+item ops have issues. See #691 for another radar. See some of the cleaning in [IGListBatchUpdateData](https://github.com/Instagram/IGListKit/blob/master/Source/Common/IGListBatchUpdateData.mm).
  - I was actually going to update this class to take in reloads as well and do the conversion internally instead of where it is now in `IGListAdapterUpdater`. That way you could use the update data standalone. I'd accept a PR for this!

DreamingInBinary:
Wow @rnystrom this info is gold 💯 - huge thanks as we're now able to land this change in our iPhone X release! Woop woop 😄 

All of that is great insight, and performance has been stellar so for us too as we rarely hit a collection view approaching 1k items within it. Thanks for all of your work with good ol' collection view, I know it can be a temperamental one sometimes haha.

Also, I've got both of those radars duped as well. Looking forward to more IGListKit goodness in the future 😄

PR #957, opened 2017-10-05
Allow IGListKit to be used without -ObjC

kastiglione:
The `-ObjC` linker flag is used to work around a mismatch between how Mach-O
files are linked, and how the Objective-C runtime operates.

With this change, IGListKit can be used without requiring the `-ObjC` linker flag.

This change annotates the headers and sources for categories. Categories are the primary reason for using the `-ObjC` flag. The annotations setup dummy variables which create a relationship between files that `#import` a category, and the file that implements a category. These dummy variables are marked weak, to prevent redundant copies of them. Also, the `-dead_strip` flag will remove the dummy variables from the final binary.

Related, see: "What causes those exceptions?" in https://developer.apple.com/library/content/qa/qa1490/_index.html

#trivial

Issue fixed: #

### Checklist

### DIFF
--- a/Source/Common/IGListMacros.h
+++ b/Source/Common/IGListMacros.h
@@ -7,6 +7,8 @@
  * of patent rights can be found in the PATENTS file in the same directory.
  */
 
+#include <os/base.h>
+
 #ifndef IGLK_SUBCLASSING_RESTRICTED
 #if defined(__has_attribute) && __has_attribute(objc_subclassing_restricted)
 #define IGLK_SUBCLASSING_RESTRICTED __attribute__((objc_subclassing_restricted))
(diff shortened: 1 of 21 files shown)


kastiglione:
It's definitely a subtle issue. The use case is when building app where IGListKit is a static library. This use case is true of any library that includes objc categories.

The way categories get compiled (no global symbols) mean the linker will, in most cases, not include the category in the final binary. The `-ObjC` flag has long been the solution to this, but the `-ObjC` flag causes the linker to include more code than is strictly necessary. The flag causes all files with objc categories to be add, which is legit, but it also causes all files with objc classes to be added, which generally is isn't needed. Plus the linker has to follow the transitive dependencies of each, so whole graphs of unused code can be added because of this flag. These macros address the linking issue around categories. They create a link time dependency between the file that defines a category (Foo+Bar.m) and any file that imports that categorys' header (Foo+Bar.h).

rnystrom:
@kastiglione is there anything special we have to do to prevent regressing this? Or just look at internal lands?

kastiglione:
Does the IGListKit integrate with any linting hooks? Linting was my plan for internal changes. There could be other ways to catch regressions, looking at the symbols and such.

rnystrom:
We use Danger for PR changes but don’t do any code punting with it. We could add a lint step to our CI.

rnystrom:
But internal blockers are ok too, just more burden on me to update.

weyert:
Stupid question, maybe, but does this mean you could bundle up `IGListKit` in a Swift app without it being an external `framework`? I have been looking into that earlier today.

kastiglione:
@rnystrom I'm working on figuring out how to best catch regressions and how to automate the macro annotations.

@weyert Is there something that prevents IGListKit from being used as a static library with swift apps?

rnystrom:
@kastiglione playing catch up, do you want me to just merge this? Or wait until we have a good regression prevention plan?

kastiglione:
Thanks for checking. It could certainly be merged as is, but I'm in no rush for this week if you want to wait until next week. Regression catching work is in progress.

rnystrom:
Great! Thanks for the update. I'll leave this in limbo til I hear back from you, sound good @kastiglione?

kastiglione:
I'll close this. Nothing stopping this from being committed, but I don't want to leave IGListKit with a responsibility that doesn't have someone looking after it.

Issue #291, opened 2016-12-06
Support for interactive cell movement/reordering


rnystrom:
We can add support for moving data, but it will probably be tricky.

For starters, add the `UICollectionViewDataSource` method `collectionView:moveItemAtIndexPath:toIndexPath:`. The biggest problem will be moving items between sections, which we probably will have to disable. The two features we can investigate supporting are:

- Moving items amongst a section
- Rearranging whole sections

rnystrom:
Originally discussed in #290

mobilinked:
I am testing and trying a solution based on IGListKit, to
- Moving amongst a section 
- Moving across sections 
- By gesture

This should be ok if to operation the collectionview directly, what I worrying about is that the app should also to update the datasource, but next time, calling adapter.performUpdates MAY cause problem. 

So if you does not plan to support "moving across sections", would you please help to add a function to adapter to "update the datasource, but without UI updating", so it is possible when special case to directly operate the collectionview and then update the datasource as normal way. 

Thanks

rnystrom:
Since the `IGListAdapter` acts as the source of truth, we could easily modify the data source ourselves and then send a delegate event that a transaction occurred (e.g. "moved object `o` from section `1` to section `3`").

You're right that w/out telling the data owner of the move they might update and then things get screwed up.

mobilinked:
IGListKit now is able to help the "APP" to move cell easily via IGListAdapter to operate data source. 
After this new proposal, IGListKit will help the "USER" to move them cell amongst section.

I'd like to make my preceding comments be clear:
As it's impossible for IGListKit to cover all the possible changes (such as move across sections, or something like move a cell to recyclebin), the developers (the users of IGListKit) may have to directly access CollectionView, and have to directly change the datasource, for these scenarios, IGListKit must be informed that "the datasource be changed, and the UI also be changed by others, so please don't automatically update the UI next time"

jessesquires:
@rnystrom - this is done, right?

rnystrom:
@jessesquires this is about interactive moves. Something that'd be nice to have, but low pri.

jessesquires:
Ahhh 👍

jessesquires:
Pushing this to 3.1.0

jlaws:
Any update on this?  Is anyone working on it?

jverdi:
^ Took a shot at it. A bit of a work in progress, but hopefully it gets the ball rolling.

PR #976, opened 2017-10-24
Interactive Reordering

jverdi:
I had a desire for interactive reordering in a personal project, so here's a first attempt at adding support in IGListKit.

I figured I might as well get a WIP PR up for comments before I continue further as there are a few aspects to interactive reordering that don't interplay perfectly with IGListKit.

As discussed in #291, I went after two prime use cases:
1. Moving items amongst a section
2. Rearranging whole sections

I also "disabled" moving items between sections by having those moves revert, to mimic interactive reordering cancellation as closely as possible.

You can see both in the Mixed Data example. Grid items can be moved within a section, while users can be moved to reorder whole sections. But trying to move a grid item out of a grid or a user item into a grid will auto-revert. The revert animation isn't as tight as it should be. It may be more desirable to disable the animation - though you lose the visual cue.

There is a also a new example, `ReorderableViewController`, that demonstrates 2 in its pure form (likely the most desired use case), where all sections are reorderable single rows.

Happy to take feedback -- this is my first experience working on IGListKit, so I would expect there to be gaps. (Ex. I haven't used `IGListStackedSectionController`, and its tests failed as I hadn't implemented reordering delegates for it. Those are simply stubbed out for now.)

Issue fixed: #291

### Checklist

### Additional TODO

- [x] Proper support in `IGListStackedSectionController`


weyert:
Thank you for this pull request @jverdi. Would this increase the minimal iOS version requirement from 8.0 to 9.0? As I think the interactive movement feature was introduced in iOS 9? The library is targeting 8.0 at the moment but maybe this can be changed @rnystrom if necessary? As Instagram itself seems to be iOS 9 or later.

jverdi:
Hi @weyert - the PR shouldn't _require_ changing the deployment target to iOS 9. The 9+ specific code for interactive reordering only gets implemented in the client apps (and IGListKit example projects). But, to take advantage of this optional feature would require iOS 9.

Common solutions to support interactive reordering on iOS <9 typically impose a specific UICollectionViewLayout like [`LXReorderableCollectionViewFlowLayout`](https://github.com/lxcid/LXReorderableCollectionViewFlowLayout) or [`DraggableCollectionView`](https://github.com/lukescott/DraggableCollectionView). I think this strikes a better balance for a library - but definitely up to the maintainers.

rnystrom:
@jverdi incredible. I’m going to set aside some time to thoroughly review this. Been a feature I and others have wanted for a while. Really excited for this and want to make sure it gets merged!

weyert:
Great thank you, @jverdi. Sounds like a great addition to the project! (and sorry for the dumb question ☺️)

jessesquires:
Thanks @jverdi ! 🙌

Source/IGListAdapterDataSource.h
```diff
@@ -60,7 +60,27 @@ NS_SWIFT_NAME(ListAdapterDataSource)
  adding the background view and maintaining its visibility.
  */
 - (nullable UIView *)emptyViewForListAdapter:(IGListAdapter *)listAdapter;
-
+    
+@optional
```
jessesquires: just noting:

we can keep this for `3.2` to avoid breaking clients and remove for `4.0`.

after this merges, let's be sure to open a new issue to track this.

Source/IGListUpdatingDelegate.h
```diff
@@ -125,6 +125,17 @@ NS_SWIFT_NAME(ListUpdatingDelegate)
                        toIndexPath:(NSIndexPath *)toIndexPath;
 
 /**
+ Tells the delegate to move a section from and to given indexes.
+ 
+ @param collectionView The collection view on which to perform the transition.
+ @param fromIndex The source index of the section to move.
+ @param toIndex The destination index of the section to move.
+ */
+- (void)moveSectionInCollectionView:(UICollectionView *)collectionView
```
rnystrom: _Technically_ this is a breaking change, but I'm not sure if ppl are making their own updaters. @jessesquires thoughts?

Source/IGListAdapterDataSource.h
```diff
@@ -60,7 +60,27 @@ NS_SWIFT_NAME(ListAdapterDataSource)
...
+    
+/**
+ Asks the datasource whether a list object should be movable as the result of interactive reordering.
+ 
+ @param listAdapter The list adapter sending this information.
+ @param sectionIndex The index of the section proposed to be moved.
+ @param index The index of the object proposed to be moved.
+ */
+- (BOOL)listAdapter:(IGListAdapter *)listAdapter canMoveObjectInSection:(NSInteger)sectionIndex atIndex:(NSInteger)index;
```
rnystrom: Is there a way we could make this a property (defaulted to `NO`) on the adapter? Adding another (eventually) required method like this will be a pretty big pain to upgrade.

Another option, (which I like a lot more), can we add a `IGListAdapterMovingDelegate` or something onto `IGListAdapter`? Then we can make all the methods required _and_ not break any upgrades. Plus, if you don't use moving, it makes your data source implementation much cleaner.

Thoughts?

Source/Internal/IGListAdapter+UICollectionView.m
```diff
@@ -57,6 +57,69 @@ - (UICollectionReusableView *)collectionView:(UICollectionView *)collectionView
...
+    id <IGListAdapterDataSource> dataSource = self.dataSource;
+    IGAssert(dataSource != nil, @"Found a nil dataSource when requesting canMoveItemAtIndexPath for interactive reordering");
+    
+    NSInteger sectionIndex = indexPath.section;
+    NSInteger itemIndex = indexPath.item;
+    
+    // for ease of implementation, first we check with the listAdapter
+    if ([dataSource respondsToSelector:@selector(listAdapter:canMoveObjectInSection:atIndex:)]) {
+        return [dataSource listAdapter:self canMoveObjectInSection:sectionIndex atIndex:itemIndex];
+    }
```
rnystrom: I left a note above, but wondering if we can just skip this entirely? Maybe just rely on the section controller deciding if a move is allowed or not?

Source/Internal/IGListAdapter+UICollectionView.m
```diff
@@ -57,6 +57,69 @@ - (UICollectionReusableView *)collectionView:(UICollectionView *)collectionView
 
     return view;
 }
+    
+- (BOOL)collectionView:(UICollectionView *)collectionView canMoveItemAtIndexPath:(NSIndexPath *)indexPath {
+    id <IGListAdapterDataSource> dataSource = self.dataSource;
+    IGAssert(dataSource != nil, @"Found a nil dataSource when requesting canMoveItemAtIndexPath for interactive reordering");
+    
+    NSInteger sectionIndex = indexPath.section;
+    NSInteger itemIndex = indexPath.item;
```
rnystrom: nit: `const` on both

Source/Internal/IGListAdapter+UICollectionView.m
```diff
@@ -57,6 +57,69 @@ - (UICollectionReusableView *)collectionView:(UICollectionView *)collectionView
...
+    // then we fall back to the sectionController method
+    IGListSectionController *sectionController = [self sectionControllerForSection:sectionIndex];
+    return [sectionController canMoveItemAtIndex:itemIndex];
+}
+    
+- (void)collectionView:(UICollectionView *)collectionView moveItemAtIndexPath:(NSIndexPath *)sourceIndexPath toIndexPath:(NSIndexPath *)destinationIndexPath {
+    NSInteger sourceItemIndex = sourceIndexPath.item;
+    NSInteger destinationItemIndex = destinationIndexPath.item;
+    NSInteger sourceSectionIndex = sourceIndexPath.section;
+    NSInteger destinationSectionIndex = destinationIndexPath.section;
```
rnystrom: nit: `const` all of it

Source/Internal/IGListAdapter+UICollectionView.m
```diff
@@ -57,6 +57,69 @@ - (UICollectionReusableView *)collectionView:(UICollectionView *)collectionView
...
+    if (sourceSectionController == destinationSectionController) {
+        [self moveInSectionControllerInteractive:sourceSectionController
+                                       fromIndex:sourceItemIndex
+                                         toIndex:destinationItemIndex];
+        return;
+    }
+    
+    // this is a reordering of sections themselves
+    if ([sourceSectionController numberOfItems] == 1
+        && [destinationSectionController numberOfItems] == 1) {
```
rnystrom: Great, I think this is the right call. Just need to make sure that it is documented heavily that movement between sections is only allowed when `numberOfItems` == 1.

I don't have the answer, but I wonder if we should be asking the section controller or the collection view for # items (e.g. `[collectionView numberOfItemsInSection:section]`). Like what if there is an item visible, but the data has changed....

Will have to think about it more.

Source/Internal/IGListAdapter+UICollectionView.m
```diff
@@ -57,6 +57,69 @@ - (UICollectionReusableView *)collectionView:(UICollectionView *)collectionView
...
+
+        if (destinationItemIndex == 0 && destinationSectionIndex > sourceSectionIndex) {
+            // the "item" representing our section was dropped
+            // into the beginning of a destination section rather than the end
+            // so it really belongs one section before the section where it landed
+            destinationSectionIndex -= 1;
+        }
+        else if (destinationItemIndex == 1 && destinationSectionIndex < sourceSectionIndex) {
```
rnystrom: nit: sameline the `} else if (...) {`

Source/Internal/IGListAdapter+UICollectionView.m
```diff
@@ -57,6 +57,69 @@ - (UICollectionReusableView *)collectionView:(UICollectionView *)collectionView
...
+    
+    // this is a reordering of sections themselves
+    if ([sourceSectionController numberOfItems] == 1
+        && [destinationSectionController numberOfItems] == 1) {
+
+        if (destinationItemIndex == 0 && destinationSectionIndex > sourceSectionIndex) {
+            // the "item" representing our section was dropped
+            // into the beginning of a destination section rather than the end
+            // so it really belongs one section before the section where it landed
+            destinationSectionIndex -= 1;
```
rnystrom: great!

Source/Internal/IGListAdapter+UICollectionView.m
```diff
@@ -57,6 +57,69 @@ - (UICollectionReusableView *)collectionView:(UICollectionView *)collectionView
...
+        // perform view changes in the collection view
+        [self moveSectionControllerInteractive:sourceSectionController
+                                     fromIndex:sourceSectionIndex
+                                       toIndex:destinationSectionIndex];
+        return;
+    }
+    
+    // otherwise this is a move of an _item_ from one section to another section
+    // this is not currently supported, so we need to revert the change as it's too late to cancel
+    [self revertInvalidInteractiveMoveFromIndexPath:sourceIndexPath toIndexPath:destinationIndexPath];
```
rnystrom: You're comments are 💯

Source/IGListSectionController.h
```diff
@@ -107,8 +107,27 @@ NS_SWIFT_NAME(ListSectionController)
  @note The default implementation does nothing. **Calling super is not required.**
  */
 - (void)didUnhighlightItemAtIndex:(NSInteger)index;
+    
+/**
+ Identifies whether an object can be moved through interactive reordering.
+ 
+ @param index The index of the unhighlighted cell.
+ 
+ @note Interactive reordering is supported both for items within a single section, as well as for reordering sections themselves when sections contain only one item. The default implementation returns false.
```
rnystrom: nit: let's newline. I try to wrap comments at 120 chars (we should add a lint)

Also "false" should be "NO" with markdown code ticks

screen shot 2017-10-24 at 3 57 04 pm

Source/IGListSectionController.m
```diff
@@ -89,5 +89,13 @@ - (void)didDeselectItemAtIndex:(NSInteger)index {}
 - (void)didHighlightItemAtIndex:(NSInteger)index {}
 
 - (void)didUnhighlightItemAtIndex:(NSInteger)index {}
+    
+- (BOOL)canMoveItemAtIndex:(NSInteger)index {
+    return NO;
+}
+    
+- (void)moveObjectFromIndex:(NSInteger)sourceIndex toIndex:(NSInteger)destinationIndex {
+    IGFailAssert(@"Section controller %@ must override %s if interactive reordering is enabled.", self, __PRETTY_FUNCTION__);
```
rnystrom: Great!

Source/IGListAdapterDataSource.h
```diff
@@ -60,7 +60,27 @@ NS_SWIFT_NAME(ListAdapterDataSource)
...
+    
+/**
+ Notifies the datasource that a list object should move as the result of interactive reordering.
+ 
+ @param listAdapter The list adapter sending this information.
+ @param sourceIndex The starting index of the object.
+ @param destinationIndex The ending index of the object.
+ */
+- (void)listAdapter:(IGListAdapter *)listAdapter moveSectionAtIndex:(NSInteger)sourceIndex toIndex:(NSInteger)destinationIndex;
```
rnystrom: B/c data sources could be out of sync w/ what is being displayed (ultimately held by `IGListAdapter` internally), I think we should do one of the following:

- Add params for to/from objects
- Add to/from object _arrays_

Source/Internal/IGListAdapter+UICollectionView.m
jverdi: Sure -- I was thinking about whole section reordering here. My motivation was that for section controllers to be reusable, they might be re-orderable or not depending on the context of the collection/adapter they appear in.

But, if the delegates are all to be required eventually, yes, it's worth dropping this to reduce complexity.

Source/IGListAdapterDataSource.h
jverdi: Removing this entirely based on later feedback.

jverdi:
Ok, feedback addressed and tests added. I still have to dive into `IGListStackedSectionController`, but the PR is close I think.

bosbefok:
Super keen to see this PR merged as it's something I'm keen to use in a current project.

jverdi:
Ok, `IGListStackedSectionController` is also tackled now, w/an example and tested. CHANGELOG is also updated.

@rnystrom @jessesquires I think this is ready for a re-review now.

ityutin:
Hey there! I've found one problem in the examples. When you try to reorder cell to the very bottom of the list, the app just crashes. 

 Terminating app due to uncaught exception 'NSInternalInconsistencyException', reason: 'attempting to invalidate an item at an invalid indexPath: <NSIndexPath: 0xc000000000201316> {length = 2, path = 19 - 1} globalIndex: 20 numItems: 20'

It happens only in such case. Actually I am very interested in this pull request and I tried to implement it by myself without changes inside IGListAdapter. And I failed with exactly the same problem. :)

jverdi:
Yikes, good catch @ityutin. I think this is caused by item movements mapping to section movements underneath the collection view. Maybe it's better dealt with at the collection view layout level by implementing this method:
```
- (NSIndexPath *)targetIndexPathForInteractivelyMovingItem:(NSIndexPath *)previousIndexPath withPosition:(CGPoint)position NS_AVAILABLE_IOS(9_0);
```
I've been trying to avoid UICollectionViewLayout since I don't want the framework to impose a layout on the client app.

I'm now wondering if I should hijack the implementation for that method by swizzling it w/the ObjC runtime. It might make some of the reordering implementation in IGListKit more straightforward too.

jverdi:
@ityutin I've updated with the approach I mentioned in my previous comment. That should solve the crash you saw. Let me know if you get a chance to try it.

jverdi:
@rnystrom ok, I've changed section moves to delegate to `IGListAdapterMoveDelegate` as requested.

Also fixed a strong reference that should've been weak in `UICollectionViewLayout+InteractiveReordering.m` (caught by unit tests).

Source/Internal/UICollectionViewLayout+InteractiveReordering.m
```diff
@@ -0,0 +1,202 @@
...
+                [UICollectionViewFlowLayoutInvalidationContext new];
+
+                flowModifiedContext.invalidateFlowLayoutDelegateMetrics =
+                [(UICollectionViewFlowLayoutInvalidationContext *)originalContext invalidateFlowLayoutDelegateMetrics];
+                flowModifiedContext.invalidateFlowLayoutAttributes =
+                [(UICollectionViewFlowLayoutInvalidationContext *)originalContext invalidateFlowLayoutAttributes];
+                modifiedContext = flowModifiedContext;
+            }
+            else {
+                modifiedContext = [UICollectionViewLayoutInvalidationContext new];
```
rnystrom: This needs to be:

```objc
modifiedContext = [[self layoutAttributesClass] new];
```

jverdi: I think this was supposed to be `invalidationContextClass`

Source/Internal/UICollectionViewLayout+InteractiveReordering.m
```diff
@@ -0,0 +1,202 @@
...
+    return [self ig_cleanupInvalidationContext:originalContext];
+}
+
+- (UICollectionViewLayoutInvalidationContext *)ig_cleanupInvalidationContext:(UICollectionViewLayoutInvalidationContext *)originalContext {
+    IGListAdapter *adapter = (IGListAdapter *)objc_getAssociatedObject(self, kIGListAdapterKey);
+    if (adapter == nil || !self.collectionView) {
+        return originalContext;
+    }
+
+    NSInteger numSections = [adapter numberOfSectionsInCollectionView:(UICollectionView * _Nonnull)self.collectionView];
```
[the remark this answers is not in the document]
jverdi: The cast is for nullability. `collectionView` is nullable on `UICollectionViewLayout`, but nonnull on the datasource call.

Source/IGListAdapterUpdater.m
```diff
@@ -518,6 +518,46 @@ - (void)reloadItemInCollectionView:(UICollectionView *)collectionView
...
+    
+    // prevent double-animation from reloadData + reloadSections
+    
+    [CATransaction begin];
+    [CATransaction setDisableActions:YES];
+    [collectionView performBatchUpdates:^{
+        [collectionView reloadSections:visibleSections];
+    } completion:^(BOOL finished) {
+        [CATransaction commit];
+    }];
```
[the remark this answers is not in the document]
jverdi: Switching to `[UIView performWithoutAnimation:]` has the interactive cell animation falling onto a copy of itself when released. I think dropping down to `CATransaction` counteracts some of the animation that the UICollectionView is trying to do simultaneously.

Source/IGListAdapterUpdater.m
```diff
@@ -518,6 +518,46 @@ - (void)reloadItemInCollectionView:(UICollectionView *)collectionView
...
+    // you end up with two items in the section for the drop location,
+    // and zero items in the section originating at the drag location
+    // so, we have to reload data rather than doing a section move
+    
+    [self performReloadDataWithCollectionView:collectionView];
+    
+    // It seems that reloadData called during UICollectionView's moveItemAtIndexPath
+    // delegate call does not reload all cells as intended
+    // So, we further reload all visible sections to make sure none of our cells
+    // are left with data that's out of sync with our dataSource
```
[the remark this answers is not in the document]
jverdi: Me too :) Admittedly my understanding of why these workarounds are necessary is falling a bit short.

Converting the `performReloadDataWithCollectionView` line to `[collectionView reloadData]` works.

Without the reload of visible sections, the collection temporarily shows stale data (despite the prior reloadData).

If you want to look at it, try the `Reorder Cells` demo in the example project and long press and drag cells around with some of the permutations you suggested.

Source/Internal/UICollectionViewLayout+InteractiveReordering.m
```diff
@@ -0,0 +1,202 @@
...
+        SEL overrideEndInvalidationSelector =
+        @selector(ig_invalidationContextForEndingInteractiveMovementOfItemsToFinalIndexPaths:previousIndexPaths:movementCancelled:);
+        Method userEndInvalidationMethod = class_getInstanceMethod(layoutClass, userEndInvalidationSelector);
+        Method overrideEndInvalidationMethod = class_getInstanceMethod(layoutClass, overrideEndInvalidationSelector);
+        method_exchangeImplementations(userEndInvalidationMethod, overrideEndInvalidationMethod);
+    });
+}
+
+- (void)ig_hijackLayoutInteractiveReorderingMethodForAdapter:(IGListAdapter *)adapter {
+    objc_setAssociatedObject(self, kIGListAdapterKey, adapter, OBJC_ASSOCIATION_ASSIGN);
```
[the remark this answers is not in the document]
jverdi: This may have been a misread on my part. I'm not seeing cycles in the memory debugger or leaks in instruments, but with retain, I'm failing 5 tests that all seem to be checking that the adapter gets released. Ex. `test_whenReleasingObjects_thatAssertDoesntFire` 🤔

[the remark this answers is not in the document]
rnystrom: Oh! I see.

Source/Internal/UICollectionViewLayout+InteractiveReordering.m
rnystrom: Errr ya, that one

Source/IGListAdapterUpdater.m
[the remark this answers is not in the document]
rnystrom: @jverdi here's w/ just doing `[collectionView reloadData]` and not doing any of the section reload stuff. Seems to be working just fine?

move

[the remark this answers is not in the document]
jverdi: I think the problems exhibit most easily when you're a little less controlled...

iglistkit-reorder-bug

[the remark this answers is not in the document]
rnystrom: Wo! That's really _really_ weird. Ok, this sounds good then.

rnystrom:
I'm going to hold on importing until Monday since everyone will be back from the holidays. SO excited!

jverdi:
Sounds great! Thanks for all the feedback.

adenisonafifi:
@jverdi I was trying out your interactive reordering functionality and I experienced a reoccurring crash in the following scenario:

UICollectionView with layout: `IGListCollectionViewLayout`
Section controllers with one item per section
Three sections per collection view row like: ( [  ] - [  ] - [   ] )

While dragging to move a section controller it would crash on: `collectionView.updateInteractiveMovementTargetPosition(position)`

I was able to successfully use your PR by switching my data source to use one section controller with many items. Please ignore if it was intended to not function with many section controllers per row.

Source/IGListSectionController.m
```diff
@@ -89,5 +89,17 @@ - (void)didDeselectItemAtIndex:(NSInteger)index {}
 - (void)didHighlightItemAtIndex:(NSInteger)index {}
 
 - (void)didUnhighlightItemAtIndex:(NSInteger)index {}
+    
+- (BOOL)canMoveItemAtIndex:(NSInteger)index {
+    return NO;
+}
+
+- (BOOL)canMoveItemAtIndex:(NSInteger)sourceItemIndex toIndex:(NSInteger)destinationItemIndex {
```
adenisonafifi: Could this be exposed for overriding in order to allow users to prevent moving to specific indices? 
Currently even if `canMoveItemAtIndex:` returns false for a specific index that item can still be moved when you move another item into its place.

rnystrom:
Thanks for the reminder! Going to merge 

Source/IGListStackedSectionController.m
```diff
@@ -135,7 +135,12 @@ - (NSInteger)numberOfItems {
 }
 
 - (CGSize)sizeForItemAtIndex:(NSInteger)index {
-    IGListSectionController *sectionController = [self sectionControllerForObjectIndex:index];
+    // during interactive reordering, its possible for an item to be moved into a section beyond the last section
+    // in that case, just return the size of the current last item in the section
+    const NSInteger maxIndex = [self numberOfItems]-1;
```
rnystrom: h/t to @amonshiz for catching this bug. `numberOfItems` can absolutely return `0`. I'm updating locally before landing.

amonshiz: 😊

jessesquires: lol oops 😳

rnystrom:
Made a gif of it in action 😄 

reorder

jessesquires:
nice! 😎

RamblinWreck77:
This is super cool! I'm really looking forward to 3.3.0 now. We're working on a live data gauge cluster and the ability to move/reorder them like this would be really slick. Great work @jverdi !!!

@rnystrom will this work on an NxN grid as well?

rnystrom:
@RamblinWreck77 3.3.0 WIP right now 😇 

I believe it should work just fine w/ an NxN grid but haven't done that myself. Would be an awesome addition to the demos app!

JGVB:
Why didn't this get merged?  :(

jessesquires:
@JGVB it was merged. f15b167e85d1a866d264040eca5978caa07d49d2

Issue #977, opened 2017-10-24
Negative sizes are not supported in the flow layout crash


svara:
### General information

- Reproducible in the demo project? (Yes/No): Yes
- Related issues: #937.

To reproduce the crash, return a size with NaN values from `(CGSize)sizeForItemAtIndex:(NSInteger)index.`  The assert in IGListAdapter doesn't help in this case. 
To guard against this, you have to check if the returned size contains NaN with isnan(x). 
This crash is not caused by IGListKit, but as suggested by @rnystrom it would be a great addition to the assert.

PR #992, opened 2017-10-27
Negative size assertion

yurybogdanov:

Assertions for item's size added to prevent negative values return from IGListAdapter

Issue fixed: #977 

### Checklist

### DIFF
--- a/Source/Internal/IGListAdapter+UICollectionView.m
+++ b/Source/Internal/IGListAdapter+UICollectionView.m
@@ -178,8 +178,8 @@
     IGAssert(![self.collectionViewDelegate respondsToSelector:_cmd], @"IGListAdapter is consuming method also implemented by the collectionViewDelegate: %@", NSStringFromSelector(_cmd));
     
     CGSize size = [self sizeForItemAtIndexPath:indexPath];
-    IGAssert(!isnan(size.height), @"IGListAdapter returned negative height %f for item at indexPath <%@>", size.height, indexPath);
-    IGAssert(!isnan(size.width), @"IGListAdapter returned negative width %f for item at indexPath <%@>", size.width, indexPath);
+    IGAssert(!isnan(size.height), @"IGListAdapter returned negative height = %f for item at indexPath <%@>", size.height, indexPath);
+    IGAssert(!isnan(size.width), @"IGListAdapter returned negative width = %f for item at indexPath <%@>", size.width, indexPath);
     
     return size;
 }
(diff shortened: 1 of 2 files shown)


Source/Internal/IGListAdapter+UICollectionView.m
```diff
@@ -176,7 +176,12 @@ - (void)collectionView:(UICollectionView *)collectionView didUnhighlightItemAtIn
 
 - (CGSize)collectionView:(UICollectionView *)collectionView layout:(UICollectionViewLayout *)collectionViewLayout sizeForItemAtIndexPath:(NSIndexPath *)indexPath {
     IGAssert(![self.collectionViewDelegate respondsToSelector:_cmd], @"IGListAdapter is consuming method also implemented by the collectionViewDelegate: %@", NSStringFromSelector(_cmd));
-    return [self sizeForItemAtIndexPath:indexPath];
+    
+    CGSize size = [self sizeForItemAtIndexPath:indexPath];
+    IGAssert(!isnan(size.height), @"IGListAdapter returned negative height = %f for item at indexPath <%@>", size.height, indexPath);
+    IGAssert(!isnan(size.width), @"IGListAdapter returned negative width = %f for item at indexPath <%@>", size.width, indexPath);
```
vincethecoder: @yurybogdanov you beat me to it. Would be nice to see a unit test that accounts for this as well in `IGListAdapterTests.m`. Nicely done 👍

yurybogdanov: @vincethecoder thank you for your feedback :)
In fact there is test in the file you mentioned:
`- (void)test_whenSectionControllerReturnsNegativeSize_thatAdapterReturnsZero`

However I will try to add test which covers the case when `NAN` is returned in size. Hope to update PR tomorrow.

vincethecoder: @yurybogdanov Yep, that's correct. Cool man. 🥇

yurybogdanov: So I spent a couple of days trying to solve this problem with test and it appears that we can't force `sizeForItemAtIndexPath` to return size with NAN (I doubt it can even exist, 'cause I've been trying to `CGSizeMake(NAN, NAN)` and got size with zeros), thus test can't be made.
I can be wrong and if I am any help here would be highly appreciated.

jessesquires: you probably want `CGFLOAT_MAX` ?

yurybogdanov: @jessesquires no, it eventually worked with NAN. Just missed one line of code :)
I've updated PR with three tests (when height is NAN and width is positive + vice versa + both width and height are NAN).

Issue #815, opened 2017-06-20
Investigate why Swift objects with duped identifiers can appear more than once


rnystrom:
Followup from #810. Duped identifier models should have just been omitted.

yemodin:
@rnystrom Hello, looks like even NSNumbers can appear 2+ times if the data source has objects with the same diffIdentifier on initial setup. I can see assert for such duplication, but it will check it only on DEBUG and looks like better to trigger "objectsWithDuplicateIdentifiersRemoved" for initial objects, with logs of such duplicates.

PR #993, opened 2017-10-29
Make sure collection have unique ListDiffable objects on initial data…

yemodin:
… source setup too

Issue fixed: #815

### Checklist

### DIFF
--- a/Source/IGListAdapter.m
+++ b/Source/IGListAdapter.m
@@ -15,6 +15,7 @@
 
 #import "IGListSectionControllerInternal.h"
 #import "IGListDebugger.h"
+#import "IGListArrayUtilsInternal.h"
 
 @implementation IGListAdapter {
     NSMapTable<UICollectionReusableView *, IGListSectionController *> *_viewSectionControllerMap;
@@ -563,13 +564,12 @@
     IGParameterAssert(dataSource != nil);
 
 #if DEBUG
-    NSCountedSet *identifiersSet = [NSCountedSet new];
     for (id object in objects) {
-        [identifiersSet addObject:[object diffIdentifier]];
         IGAssert([object isEqualToDiffableObject:object], @"Object instance %@ not equal to itself. This will break infra map tables.", object);
-        IGAssert([identifiersSet countForObject:[object diffIdentifier]] <= 1, @"Diff identifier %@ for object %@ occurs more than once. Identifiers must be unique!", [object diffIdentifier], object);
     }
 #endif
+
+    NSArray *uniqueObjects = objectsWithDuplicateIdentifiersRemoved(objects);
 
     NSMutableArray<IGListSectionController *> *sectionControllers = [NSMutableArray new];
     NSMutableArray *validObjects = [NSMutableArray new];
@@ -583,7 +583,7 @@
     // for IGListSectionController subclasses after calling [super init]
     IGListSectionControllerPushThread(self.viewController, self);
 
-    for (id object in objects) {
+    for (id object in uniqueObjects) {
         // infra checks to see if a controller exists
         IGListSectionController *sectionController = [map sectionControllerForObject:object];
 
--- a/Source/IGListAdapter.m
+++ b/Source/IGListAdapter.m
@@ -144,7 +144,8 @@
 - (void)updateAfterPublicSettingsChange {
     id<IGListAdapterDataSource> dataSource = _dataSource;
     if (_collectionView != nil && dataSource != nil) {
-        [self updateObjects:[[dataSource objectsForListAdapter:self] copy] dataSource:dataSource];
+        NSArray *uniqueObjects = objectsWithDuplicateIdentifiersRemoved([dataSource objectsForListAdapter:self]);
+        [self updateObjects:uniqueObjects dataSource:dataSource];
     }
 }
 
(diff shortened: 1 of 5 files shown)


yemodin:
@rnystrom can we use something like this for #815? Or better handle this at AdapterUpdater?

yemodin:
Not sure about logic behind this test https://github.com/Instagram/IGListKit/blob/a6f264fc12a802bcf54c103ec40202474aa3067b/Tests/IGListDebuggerTests.m#L48

Source/IGListAdapter.m
```diff
@@ -563,14 +564,13 @@ - (void)updateObjects:(NSArray *)objects dataSource:(id<IGListAdapterDataSource>
...
 #if DEBUG
-    NSCountedSet *identifiersSet = [NSCountedSet new];
     for (id object in objects) {
-        [identifiersSet addObject:[object diffIdentifier]];
         IGAssert([object isEqualToDiffableObject:object], @"Object instance %@ not equal to itself. This will break infra map tables.", object);
-        IGAssert([identifiersSet countForObject:[object diffIdentifier]] <= 1, @"Diff identifier %@ for object %@ occurs more than once. Identifiers must be unique!", [object diffIdentifier], object);
     }
 #endif
 
+    NSArray *uniqueObjects = objectsWithDuplicateIdentifiersRemoved(objects);
```
rnystrom: While not opposed to this change, this does make [deduping in the updater](https://github.com/Instagram/IGListKit/blob/master/Source/IGListAdapterUpdater.m#L109) redundant.

Maybe instead the deduping should be done here:

https://github.com/Instagram/IGListKit/blob/master/Source/IGListAdapter.m#L146

yemodin: @rnystrom If I dedupe at https://github.com/Instagram/IGListKit/blob/master/Source/IGListAdapter.m#L146 we still have the same issue reappear on reloadDataWithCompletion for adapter: https://github.com/Instagram/IGListKit/blob/b3ffafc018978c20845a3b5aa7b2edb9db8c7622/Source/IGListAdapter.m#L368

rnystrom: Oooooooh, now I see it! Wo this is an awesome find @yemodin! How about we:

- Move the dedupe to [public settings change](https://github.com/Instagram/IGListKit/blob/master/Source/IGListAdapter.m#L146) **_and_** [reloadData](https://github.com/Instagram/IGListKit/blob/master/Source/IGListAdapter.m#L362)
- We make a followup issue to consolidate and dedupe in one place

I actually want to move object generation entirely so that "to" object arrays are only constructed once. Atm they are built w/ every call to `performUpdatesAnimated:completion:`, even if a bunch of calls are updated into a single batch.

yemodin: Done. Also made deduping before https://github.com/Instagram/IGListKit/blob/b3ffafc018978c20845a3b5aa7b2edb9db8c7622/Source/IGListAdapter.m#L322

rnystrom:
@yemodin would love to import this, but we still need you to sign the CLA, then we're good to go! https://code.facebook.com/cla

Source/IGListAdapter.m
```diff
@@ -319,14 +321,14 @@ - (void)performUpdatesAnimated:(BOOL)animated completion:(IGListUpdaterCompletio
     }
 
     NSArray *fromObjects = self.sectionMap.objects;
-    NSArray *newObjects = [dataSource objectsForListAdapter:self];
+    NSArray *uniqueObjects = objectsWithDuplicateIdentifiersRemoved([dataSource objectsForListAdapter:self]);
```
rnystrom: If we're going to put this here (which we don't need to yet), we should remove it [from here](https://github.com/Instagram/IGListKit/blob/master/Source/IGListAdapterUpdater.m#L109) since that'll be wasteful. I lean towards not messing w/ it here since it'll be extra work (dedupe every update call instead of at coalescence).

I'm actually going to patch that change then import this!

rnystrom:
Internal CI is reporting a test failure for `-[IGListAdapterTests test_whenPassingNonUniqueIdentifiers_adapterUpdatesShouldSkipDuplicates]`. I'll need to figure out what's going on there before we can land.

yemodin:
@rnystrom It fails because "revert dedupe in performUpdatesAnimated". If only set objects twice and perform update on adapter it will pass duplicates too

drinkius:
Is this pull request going to be merged to `master`? Currently in our project we face an issue with the `Identifiers must be unique!` assert, in our setup having several object with the same IDs is expected and having this Assert doesn't bring much value

yemodin:
@rnystrom Should I return dedupe to performUpdatesAnimated? Let me know if we have better options.

rnystrom:
@drinkius it'll be in master momentarily. Bear in mind that this will assert _more_ if your objects aren't unique. The assert isn't going away. In fact, your objects w/ duped IDs are being stripped.


@yemodin no worries! Patching internally then landing 😄

drinkius:
@rnystrom From what I see in the current implementation in @yemodin 's repo it just strips the duplicates (I indeed need this behaviour) but you can continue working well with the stripped list, is it going to change?

rnystrom:
Nope, this PR adds deduping to a few places where it was missing and causing inconsistencies. The assert is staying put tho as a warning to the developer that the data you're putting into IGLK isn't what you're going to see.

rnystrom:
Need to unland and re-evaluate, this caused some regressions internally

rnystrom:
False alarm, something else started crashing in a similar spot at almost the exact same time. Will be reverting my revert this week 😅

jessesquires:
@rnystrom -- re-opened this so we don't loose track. close once you successfully un-revert? 😄

rnystrom:
Added in c7d4dc29a8b86b582e3d08e5531f631eb960301d

PR #1011, opened 2017-11-08
Adds IGListDiffable type information

tapi:
Add type information for variables and methods that accept and/or return `id` when what is meant is `IGListDiffable`

Nothing to test as all that changed is the api as seen from Swift, that being said I did have to update a test to include the NSString+IGListDiffable category to get test to compile.

Changing the API to expose `IGListDiffable` resulted in some changes in how method names are generated for Swift. In order to reduce change overhead I've over-ridden the swift name so that the methods have the same signature (albeit with the new type information)

There was a lot of churn after running  `pod_setup.sh`

Issue fixed: #1077

### Checklist


weyert:
Lovely!

rnystrom:
@jessesquires thoughts on this? This will be a big breaking change now that I think about it. Maybe we hold it until 4.0?

Examples/Examples-macOS/Pods/Manifest.lock
```diff
@@ -17,4 +17,4 @@ SPEC CHECKSUMS:
 
 PODFILE CHECKSUM: bbe5708b3a7f2dd7c70bb14225a7e37f67e1de58
 
-COCOAPODS: 1.4.0.beta.1
+COCOAPODS: 1.3.1
```
jessesquires: please revert these changes. (use CP 1.4.0 beta)

jessesquires:
@rnystrom - yep. This is breaking for ObjC too. Should go to 4.0

Examples/Examples-macOS/Pods/Manifest.lock
tapi: Not that it would have made a difference for this change but should that be reflected in the Gemfile?

rnystrom:
Do we still want to push for this? Been a while and I'm not totally sold if its the right call since this hasn't come up once in feedback.

tapi:
It was more of an API Ergonomics issue, Its always going to be that type so why not use it.
This started off as wanting typing in the `ListAdapterDelegate` methods and it sort of ended up being a rabbit hole.

From a practicality standpoint I guess it's only really useful in reducing the number of casts when forwarding the object to something else expecting a `ListDiffable`. 
I imagine most uses will either return a static section controller or cast to a more concrete type.

PR #1020, opened 2017-11-21
Add new iOS 11 api adjustedContentInset support

yiplee:
adopt adjustedContentInset instead of contentInset on iOS 11

### Checklist


rnystrom:
Awesome! Would love a change log entry and even a unit test for this.

yiplee:
Change log entry and unit test added :)

yiplee:
screen shot 2017-11-22 at 01 37 48

Yes, the new files have been included by pod.

rnystrom:
@yiplee are the examples building for you?

yiplee:
@rnystrom  I agree with you. But the bug will remain exist if people forget to replace ```containerInset``` by ```adjustedContainerInset``` in their codes.  And in my opinion, when people call context's ```containerInset```, they expect context to return the real content inset applied to scroll view.

rnystrom:
@yiplee though even more about this, I think the only way to go is with a separate API. I found a few places that will break in Instagram if we make this change, so that’s a blocker.

Would you be ok making this a separate API? Thanks!!

yiplee:
@rnystrom It's OK. I will update the PR right now. 😄

PR #1025, opened 2017-11-27
Update Travis to Xcode 9.2 #trivial

rnystrom:
Issue fixed: #1009

PR #1041, opened 2017-12-19
[Guides] Fix dead link in migration.md

jessesquires:


rnystrom:
I actually have these all fixed in a new local commit for #1025, pushing it now

jessesquires:
👍

Issue #1060, opened 2018-01-22
[Travis] cannot find cocoapods


jessesquires:
travis seems to be barfing on installing CP. not sure why.

https://travis-ci.org/Instagram/IGListKit/jobs/331557269#L428

```
Could not find proper version of cocoapods (1.4.0) in any of the sources
```

jessesquires:
Ahh, I *think* this is a caching problem.

Travis is caching old bundler versions.

PR #1065, opened 2018-01-29
Remove example code and UI tests from CI #trivial

rnystrom:
I'm over dealing w/ the example projects on CI slowing things down and giving poor signal. Now there's a [SwiftLint error](https://travis-ci.org/Instagram/IGListKit/jobs/332648229) that I don't really understand. At the point where its discouraging me from being more active and keeping PRs moving.

My vote is to burn down examples building on CI, and we can deal w/ breakages as they come. Any objections @jessesquires?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -25,18 +25,18 @@ env:
    - EXAMPLE_SCHEME="IGListKitExamples"
 
    matrix:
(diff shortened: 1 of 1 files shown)


jessesquires:
See discussion in #1067. I think that's the right way to go 😄

jessesquires:
Also, we need to fix #1060 then CI should be ✅

rnystrom:
Agree w/ comments in other issues, appreciate everyone jumping in here!

PR #1067, opened 2018-01-29
Add SwiftLint installation step

Sherlouk:
Alternative to #1065 

From my understanding in the lint.sh script we have a check which is basically saying if the output of `swiftlint version` does not equal "0.24.2" then exit with a failure note

As per this [travis blog](https://blog.travis-ci.com/2017-12-06-xcode-92-has-landed) they currently use "0.24.0" as the pre-installed version therefore this would cause the failure

I'm attempting to add a `brew install` step which _should_ guarantee the version to be something we can deal with!

Would be a shame to just ditch CI on all the example projects!

### DIFF
--- a/scripts/lint.sh
+++ b/scripts/lint.sh
@@ -1,6 +1,8 @@
 #!/bin/bash
 
-VERSION="0.24.2"
+VERSION="0.24.0"
+
+brew switch swiftlint $VERSION
 
 if which swiftlint >/dev/null && [ $(swiftlint version) == $VERSION ]; then
     swiftlint lint --config ../.swiftlint.yml


jessesquires:
Here it looks like travis is complaining about `brew install swiftlint@0.24.2`.

It is probably good to keep travis up-to-date. We should just tell `brew` to upgrade swiftlint.

Sherlouk:
It is a build phase inside of the project, correct, but that build phase still gets executed when run on CI!

So we could totally disable the `exit 1` which would fix it, but then we would never be running SwiftLint and really at that point we're not getting any of the benefits of it if it's just not enforced!

We could remove the version check which would allow SwiftLint to continue to work, but with a version we might not be using locally so violations could be different

As I DMd Ryan on Twitter, I see our options boiling down to:
- We change the locked version in the `lint.sh` to 0.24.0 which is what Travis uses
- or we install a new version to be the one we expect (0.24.2)

As this PR has shown though I tried the second one and Travis seems to not be able to find SwiftLint for likely a different reason entirely which I'll look into ASAP!

jessesquires:
🙌 Thanks so much @Sherlouk !

rnystrom:
Might need a rebase on #1069

PR #1068, opened 2018-01-29
Update .travis.yml and lint.sh, try to fix #1060 and lint errors #trivial

jessesquires:
project:
- fix file target membership issues in framework targets and test targets
- fix private/internal header imports, which shouldn't be `<IGListKit/` apparently
- fix static analyzer errors

travis:
- always install latest swiftlint
- ~~don't cache bundler, attempts to fix #1060~~
- remove markdown link check

swiftlint:
- make script non-failing if *any* version of swiftlint is installed
- warning if incorrect version is installed
- fail if not installed
- remove `scripts/generate_ci_yaml.rb`, we can just set the config file path directly

PR #1070, opened 2018-01-29
SwiftLint CI Changes

Sherlouk:
Replaces #1067 because I suck at rebasing 🥉

### DIFF
--- a/scripts/lint.sh
+++ b/scripts/lint.sh
@@ -1,16 +1,27 @@
 #!/bin/bash
 
-VERSION="0.24.0"
+VERSION="0.24.2"
+FOUND=$(swiftlint version)
 
-brew switch swiftlint $VERSION
-
-if which swiftlint >/dev/null && [ $(swiftlint version) == $VERSION ]; then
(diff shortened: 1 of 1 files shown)


jessesquires:
I think it would be ideal to keep SwiftLint updated, because travis can be slow to update.

What if we do this?

- We check if swiftlint (any version) is installed
    - if correct one is not installed, trigger a warning but run SwiftLint anyway
    - if correct is installed, run it
- If no version is installed at all, `exit 1`

jessesquires:
turns out my other projects had the same problem. lol 😳 

my solution:
https://github.com/jessesquires/JSQDataSourcesKit/pull/109

Sherlouk:
@jessesquires pull your changes over they look perfect!

jessesquires:
pushed updates to this PR 😄

jessesquires:
@Sherlouk  @rnystrom 

Oh -- just realized this is not merging into master. we should update

jessesquires:
actually --- i'll just make all these changes together at #1068

Issue #1071, opened 2018-01-29
Remove unnecessary nil check on transitionDelegate


rnystrom:
Here:

https://github.com/Instagram/IGListKit/blob/master/Source/Internal/IGListAdapter+UICollectionView.m#L221

and here:

https://github.com/Instagram/IGListKit/blob/master/Source/Internal/IGListAdapter+UICollectionView.m#L235

PR #1076, opened 2018-01-31
Removed unnecessary nil check on transitionDelegate

vasilenkoigor:

Issue fixed: #1071

### Checklist


rnystrom:
FYI removed the changelog entries since this is to a feature that hasn't been released yet (the addition of the feature is in the existing changelog already).

Thank you!!

rnystrom:
Turns out this was actually necessary. If the delegate is nil then we return nil initial and final attributes instead of the original. 😬

 I’m reverting this now.

PR #1080, opened 2018-02-02
Bump min macOS version to 10.11 #trivial

rnystrom:
Followup from #1068

The `NSIndexPath` methods we use **are not** available <10.11. In theory this has always been "broken" b/c older macOS versions would crash at runtime when trying to use these methods.

Normally this would be breaking change, but I can't imagine anyone is using IGListKit on the mac targeting 10.10 and shipping a working app.

cc @jessesquires for thoughts

This should be the last remaining item before we get a green badge.

(note that 10.11 was released 9/30/15 which is about 2 weeks before the public IGListKit release)


rnystrom:
THE BUILD IS GREEN 🌮🌮🌮

jessesquires:
Oh one more thing -- I think you need to open up the macOS example project and update all the targets

rnystrom:
@jessesquires good catch yup, updating then landing 

rnystrom:
Gonna wait on green again now that examples and stuff have been updated. Funny enough, the macOS example was already set to 10.11, probably b/c it was crashing w/out it.

rnystrom:
Alright still green, landing

Issue #871, opened 2017-07-27
Including better Swift interfaces/wrapper APIs


jessesquires:
## Proposal

CocoaTouch frameworks can include both Swift and ObjC. IGListKit should expose better Swift interfaces by including some Swift wrapper APIs. 

For example, we could:

- Provide a proper *generic* `SectionController` in Swift by wrapping `IGListBindingSectionController`, since ObjC lightweight generics aren't really useful in Swift.
- Provide default parameter values to initializers 
- Namespace the `IGListDiff` functions and provide defaults for the options

-------

Before, this was blocked by IG internal, but I think this is resolved now? IG internal could re-map the framework to BUCK to *exclude* the `.swift` files (in fact, this should already work by default since the BUCK file should only be including `.h`, `.m`, `.mm`, etc.). Thus, only the OSS version would include Swift.

cc @rnystrom can IG/FB internal handle Swift scale now? 😄

rnystrom:
Should be doable, tho Buck still has some issues with mix-matched libs. I _think_ its all fixed with Xcode 9 tho now that static libs work.

rnystrom:
At the very least we could include this in he public project but just exclude it with Buck. Been thinking about this a lot, have some ideas.

Issue #146, opened 2016-10-31
Better Swift interop for -[IGListDiffable diffIdentifier]?


rnystrom:
B/c the return type of `diffIdentifier` is `id<NSObject>`, we end up with Swift that looks like this:

```swift
func diffIdentifier() -> NSObjectProtocol {
    return messageID
}
```

I'm not crazy about this. Could we change the method to `- (id)diffIdentifier`? I'd imagine we end up w/

```swift
func diffIdentifier() -> AnyObject {
    return messageID
}
```

Right? The `AnyObject` docs say:

> `AnyObject` can also be used as the concrete type for an instance of a type that bridges to an Objective-C class. Many value types in Swift bridge to Objective-C counterparts, like `String` and `Int`.

So if `AnyObject` is meant for interop w/ ObjC, and all ObjC objects respond to `hash` and `isEqual:` (which is [what the diffing algo uses](https://github.com/Instagram/IGListKit/blob/master/Source/IGListDiff.mm#L57-L67)).

cc @jessesquires for thoughts

jessesquires:
Not quite: `id` now bridges to `Any` in Swift, so you would end up with

`func diffIdentifier() -> Any`

Then, I imagine you'd get `does not respond to selector` crashes.

Have you tried this yet?

zhubofei:
Can we rewrite the objective-c++ in swift? Will there be a performance drop?

rnystrom:
@zhubofei no, not for apps that don't want to bundle the Swift runtime (e.g. Instagram.app)

@jessesquires Ah, ok. I'll play around w/ this and see what happens. Really just anything bridgeable to an ObjC object is fine.

rnystrom:
Adding to 2.0.0 milestone to make a decision. I think changing this to `id` will give us `Any` in Swift and that should be better to work w/ I think. As long as its an ObjC object that gets returned so we can call `hash` and `isEqual:` on it.

jessesquires:
> I think changing this to `id` will give us `Any` in Swift 

Correct.

> As long as its an ObjC object that gets returned so we can call `hash` and `isEqual:` on it.

There's no way for us to guarantee this. `Any` in Swift means any *value or reference type*. If clients return a `struct` or `enum`, we'll crash when trying to message `hash` and `isEqual:`.

jessesquires:
I think we're stuck with `id<NSObject>` (swift: `NSObjectProtocol`).

[...]

Issue #35, opened 2016-09-29
[Swift] reference type and value type conformances to IGListDiffable


rnystrom:
The handy `NSObject+IGListDiffable` category doesn't really help with Swift objects like `String` and `Int`. This seems a little tricky because `String` is a `struct` but is convertible to an `NSString`. Doing that conversion gives us the `NSObject<IGListDiffable>` conformance, but that's kind of lame. I wonder if there's a way we can get this to work a little better?

For example, the Swift 3.0 compiler wont allow this:

``` swift
let letters: [IGListDiffable] = ["a", "b", "c"]
```

But you can get this to work a few ways:

``` swift
let strings: [NSString]  = ["a", "b", "c"]
let letters = strings as [IGListDiffable]
// or
let letters = ["a" as IGListDiffable, "b" as IGListDiffable, "c" as IGListDiffable]
// or
let letters: [IGListDiffable] = ["a" as NSString, "b" as NSString, "c" as NSString]
```

Also adding an extension to `String` isn't easy either, since `String` is a struct and `IGListDiffable` is an Objective-C protocol...

> Note that this also stunts using `IGListDiffable` with Swift structs

matthewcheok:
Is there still no way we have a swift struct conform to `IGListDiffable`?

rnystrom:
@matthewcheok nope 😕  when using Swift objects have to be a `class`. I'm sure there's a good pattern by boxing with something like `Box<StructType>` where `Box` is a `class` conforming to `IGListDiffable`.

jessesquires:
Going to close this since there's nothing actionable right now (and for the foreseeable future)

danielgalasko:
Inspired by @jessesquires I will put my solution in here for enabling swift value types to conform to Diffable. I ended up with a new Swift protocol `ItemDiffable` and used `Equatable` as the means for exposing `isEqual(toDiffableObject object)` from IGListKit.

I then wrap diff'ing around a struct called `DiffUtility`

```swift
/**
 A diffable value type that can be used in conjunction with
 `DiffUtility` to perform a diff between two result sets.
 */
public protocol ItemDiffable: Equatable {
    
    /**
     Returns a key that uniquely identifies the object.
     
     - returns: A key that can be used to uniquely identify the object.
     
     - note: Two objects may share the same identifier, but are not equal.
     
     - warning: This value should never be mutated.
     */
    var diffIdentifier: String { get }
}

/**
[...]

PR #1081, opened 2018-02-03
[WIP][RFC] Add Swift-bridge subspec

rnystrom:
# ⚠️ WIP ⚠️  

(_and will be for a while_)

### Goals

- Allow Swift `structs` to be used with IGListKit core mechanics
  - Diffing, `IGListAdapter` data source and APIs, `IGListBindingSectionController`
- Remove need to `guard` and `fatalError()` for optional values on `IGListSectionController`
- Reduce complexity of `IGListAdapterDataSource` and `IGListBindingSectionControllerDataSource`
  - Convert data source methods into a tuple of data and a section-controller generation function
  - Explicit mapping between data and section controller

#### Naming

I'm not sold on naming yet, but I've started exploring `ListSwift*` as a prefix for all Swift-bridge systems. Thoughts?

#### Diffing & structs

Add three new protocols:

- `ListSwiftIdentifiable` to uniquely identify data
  - Return an `Int` for easier interop w/ `Hashable` (below)
  - analogous to `diffIdentifier()`
- `ListSwiftEquatable` to compare data values
  - analogous to `isEqual(toDiffable object:)`
- `ListSwiftDiffable` that is just a union of identifiable and equatable
  - Maybe this could be omitted w/ Swift 4 unions, e.g. `ListSwiftIdentifiable & ListSwiftEquatable`

Can provide default conformance to `Hashable` and `Equatable` via:

```swift
public extension Hashable {
    var identifier: Int {
        return hashValue
    }
}

public extension Equatable {
    func isEqual(to object: ListSwiftEquatable) -> Bool {
        guard let object = object as? Self else { return false }
        return self == object
    }
}
```

- This lets any `struct` that conforms to `Hashable` and `Equatable` automatically satisfy `ListSwiftDiffable`
- Doesn't _require_ you to use hash+equal

An example `struct` conformance would look like this:

```swift
struct Person: Hashable, Equatable {
    let name: String
    let age: Int

    var hashValue: Int {
        return name.hashValue ^ age
    }

    static func ==(lhs: CustomValue, rhs: CustomValue) -> Bool {
        return lhs.name == rhs.name
            && lhs.age == rhs.age
    }
}
extension Person: ListSwiftDiffable { }
```

- [ ] Is everyone ok with returning an `Int` for the identifier instead of `NSObjectProtocol`?
- [ ] Are we ok w/ the default hash + equality mapping?
- [ ] Would anyone prefer we **explicitly** require conformance to `Hashable` and `Equatable`?

#### Boxing

Provide an **internal-only** box object that bridges `ListSwiftDiffable` and `ListDiffable`. Should be as simple as:

```swift
internal final class ListDiffableBox: ListDiffable {

    internal let value: ListSwiftDiffable

    init(value: ListSwiftDiffable) {
        self.value = value
    }

    // MARK: ListDiffable

    func diffIdentifier() -> NSObjectProtocol {
        return value.identifier as NSObjectProtocol
    }

    func isEqual(toDiffableObject object: ListDiffable?) -> Bool {
        if self === object { return true }
        guard let object = object as? ListDiffableBox else { return false }
        return value.isEqual(to: object.value)
    }

}
```

Use this to interact w/ all Objective-C systems (`ListAdapter`, `ListSectionController`, etc)

- [ ] Performance could be a concern w/ allocating tons of boxes. Maybe we have a global reuse pool for boxes?

#### Data Source

`ListAdapterDataSource` has 3 **required** functions:

- Return an array of `ListDiffable` objects
- Given a `ListDiffable` object, return a `ListSectionController
- Return an optional `UIView` for empty state

I have 2 big issues w/ this design in Swift:

- Empty view isn't always needed, boilerplate
- Mapping an object to a section controller is extremely error-prone b/c we cannot rely on the compiler to catch anything
  - Frequently have to `if object is Person { return PersonSectionController() }` which explodes in complexity quickly
  - Easy to miss during refactors
  - Can add new data to the `objects: [ListDiffable]` return value and forget to add the section controller pairing

I propose we slim this down into a **single method** that returns a tuple with:

- A data conforming to `ListSwiftDiffable`
- A **function** that returns a new instance of `ListSectionController`
  - Function b/c we don't create a section controller when one already exists

This required function would look like:

```swift
public protocol ListSwiftAdapterDataSource: class {
  func values(adapter: ListSwiftAdapter) -> [ 
    (
      ListSwiftDiffable, 
      (ListSwiftDiffable) -> (ListSectionController)
    )
  ]
}
```

And an example implementation would be as simple as:

```swift
func values(adapter: ListSwiftAdapter) -> [(ListSwiftDiffable, (ListSwiftDiffable) -> (ListSectionController))] {
    let values: [Person] = // ... get from somewhere
    return values.map { ($0, { _ in PersonSectionController() }) }
}
```

Notice the explicit pairing of `Person` with `PersonSectionController`.

- [ ] Any concerns about returning this type of mapping?
- [ ] The bridge adapter will need to take care to release the closure once executed. Very easy to retain too much stuff in there which could grow unbounded if we're not draining.

## Relevant issues

#35, #146, #871


rnystrom:
Note that this branch is already usable as a subspec if you'd like to experiment w/ it.

```
pod 'IGListKit/Swift', :git => 'https://github.com/Instagram/IGListKit.git', :branch => 'swift'
```

HarshilShah:
This is some great work, and will make using IGListKit with Swift so much better!

Some feedback on the implementation so far:

> Are we ok w/ the default hash + equality mapping?

I’m not entirely certain about this one. To me at least, `diffIdentifier()` seemed explicit about it being a way to specify which properties in particular determine a model’s identity i.e. which properties can be changed to create a new, distinct model as opposed to an updated model.

As with your above example, I think people tend towards including all properties in their `Hashable` implementation. This essentially reduces `Hashable` to basically to `Equatable`, completely removing it’s value in the sense of its use as a `diffIdentifier()` in IGListKit.

Moreover, Swift 4’s automatic `Equatable` and `Hashable` synthesis seems to work this way as well, essentially making `Hashable`. The [Swift Evolution proposal (SE-0185)](https://github.com/apple/swift-evolution/blob/master/proposals/0185-synthesize-equatable-hashable.md "SE-0185: Synthesizing Equatable and Hashable conformance") suggests that all stored properties are used in the `Hashable` derivation (P in the quote is `Hashable`/`Equatable`):

> The compiler synthesizes P's requirements for a struct with one or more stored properties if and only if all of the types of all of its stored properties conform to P.

This can be confirmed just by deleting the explicit Hashable implementation from the example code above.

```swift
struct Person: Hashable {
    let name: String
    let age: Int
}

let personOne = Person(name: "Harshil", age: 22)
let personTwo = Person(name: "Harshil", age: 23)

print(personOne.hashValue == personTwo.hashValue)
/// Outputs `false`
```

In the above example, since the hashValues a.k.a. diffIdentifiers don’t match up, the operation will be treated as a delete + insert, instead of as an update.

This behaviour can be achieved by defining the `hashValue` manually, however that is opt-in as of Swift 4 and so may cause issues with people seeing unexpected behaviour. Moreover this would manifest only in use and not even be caught in many UI tests, let alone throw a compiler error or warning, which would be much more difficult to debug.

For these reasons I think that an explicit `diffIdentifier` API should remain. This way its purpose would be demarcated and documented, and its absence would throw a compiler error. While it would add an extra property to otherwise “vanilla” types, and  I think it’s preferable to relying on users to remember to override `Hashable` with their own implementations, which also creates room for error (`name.hashValue ^ age` would be 1 for all names if the age is 0!) and could be less performant or optimised than compiler generated implementations.

> Would anyone prefer we explicitly require conformance to `Hashable` and `Equatable`?

I’m not sure if requiring `Equatable` would work, since that would open the pandora’s box that of PATs. The current proposed solution seems optimal to me with `ListSwiftEquatable` coming for free with `Equatable`.

> Thinking about it more, the `ListSwiftIdentiable`, `ListSwiftEquatable` dance probably isn't even necessary. We can just have a Swift version of `ListDiffable` in `ListSwiftDiffable` or w/e and call it a day.

Absolutely agree with this. They are never used individually, so it makes a lot of sense this way.

> Mapping an object to a section controller is extremely error-prone b/c we cannot rely on the compiler to catch anything

This is definitely an issue, and one I run into way more frequently than I would like to, however I’m not sure the proposed solution does much to solve that.

```swift
public protocol ListSwiftAdapterDataSource: class {
  func values(adapter: ListSwiftAdapter) -> [ 
    (
      ListSwiftDiffable, 
      (ListSwiftDiffable) -> (ListSectionController)
    )
  ]
}
```

While it does localise the model and section controller declarations to within the same function, they would usually only be a few lines apart in the existing implementation, and perhaps easier to read individually. Additionally, while they are an improvement over Objective-C blocks, Swift closures are a bit difficult to grok, especially when within a tuple within an array.

But keeping aside readability, this doesn't help achieve the actual goal much because there would be no warning or error on making a mistake here. Even for the example code, this would compile just fine if you replaced `PersonSectionController` with any other section controller; it is entirely up to users to find and correct the error when a crash occurs, which is pretty much the same as it is right now.

At first I thought the proposed `ListSwiftSectionController<T: ListSwiftDiffable>` might be the solution here, allowing us to constrain the `ListSectionController` to the `ListSwiftDiffable`, like so:

```swift
public protocol ListSwiftAdapterDataSource: class {
  func values<T: ListSwiftDiffable>(adapter: ListSwiftAdapter) -> [ 
    (
      T, 
      (T) -> (ListSwiftSectionController<T>)
    )
  ]
}
```

While this does indeed provide a type-safe interface which would work for the provided example code with an array of `Person`, it brings up the same issue as `Equatable` above, in that you cannot return a heterogenous array of arbitrary `ListSwiftDiffable` model/closure tuples this way.

I’m not a generics expert so I would probably confirm with someone who is one, but as best as I can tell there’s no type-safe way to enforce the behaviour we’re looking for here.

In the absence of a fix in Swift to better handle such data structures, I'm not sure unifying the model and section controller retrievals helps much by itself. So I would suggest retaining the separate methods for retrieving models and section controllers which might be better in terms of code organisation and legibility. There's probably even some performance gains to be had from not generating and releasing arrays of closures which will never be called every time `adapter.performUpdates(animated:)` is called.

----

To sum up, a few minor qualms aside, this really is a great change, and I can't wait to use it!

Apologies if this was overly lengthy or rambling, this is my first time partaking in the discourse around such a major API design.

rnystrom:
@HarshilShah thanks so much for taking the time to write this up. You make some fantastic points.

I'm pretty convinced about your comments re: hash+equality. I also had the idea today of maybe making the identifier a `String` where we can also internally namespace identifiers to the object type being used.

- So `struct Person { ... }` could return `name` as its identifier, but internally we would convert it to `"\(value.Self)\(value.identifier)" or something, resulting in "Personryan"
- This allows identifier methods to be very very simple and still avoid collisions

I tried **exactly** what you outlined w/ generic pairings in the function. You're right that you don't get any added compiler protection against mismatching your data + section controller w/ the tuple. However you at least explicitly pair them side-by-side in code, which I _think_ would make it harder to screw up.

(even I still mess this up)

Maybe we can pull in some Swift geniuses to weigh in on this. We obviously _have_ to keep the data heterogeneous.

HarshilShah:
The string interpolation is what I'm already doing! Helps ease my mind about potential collisions quite a bit, and I'd be down with `String` identifiers as well.

I'm mostly ambivalent about the `values(adapter:)` design, but I could live with it either way.

I'd definitely consult at least a few geniuses about the generics issue. I *think* type erasure might help a bit here but based on the little reading I've done I'm not sure.

rnystrom:
@HarshilShah got any details or resources on type erasure? Idk what that is.

HarshilShah:
I remember reading this (well, trying to) a while back: https://www.bignerdranch.com/blog/breaking-down-type-erasures-in-swift/

Seems like it might not be the solution needed here though (read the very last paragraph)

rnystrom:
Dang I thought I had it w/ type erasure... @HarshilShah + [this tweet](https://twitter.com/daloog/status/960265522486095872) pointed me in a direction:

```swift
public struct Value {
    let value: ListSwiftDiffable
    let constructor: () -> ListSectionController // base class

    // params are generic
    init<T: ListSwiftDiffable>(value: T, constructor: @escaping () -> ListSwiftSectionController<T>) {
        self.value = value
        self.constructor = constructor
    }
}

public protocol ListSwiftAdapterDataSource: class {
    func values(adapter: ListSwiftAdapter) -> [ Value ]
}
```

This compiles just fine, even when used in `ListSwiftAdapter`. However in my data source implementation:

```swift
// section controller
class PersonSectionController: ListSwiftSectionController<Person> {
    // ...
}

// data source conformance
func values(adapter: ListSwiftAdapter) -> [Value] {
    return values.map { Value(value: $0, constructor: { PersonSectionController() }) }
}
```

I get a compiler error w/

```
Cannot convert value of type '() -> PersonSectionController' to expected argument type '() -> ListSwiftSectionController<_>'
```

Any ideas? Pretty surprised this doesn't work.

sphilipakis:
What about this (note in this case Value has to be a class)
`
public class Value<T> {
    let value: T
    let constructor: () -> ListSectionController
    
    init(value: T, constructor: @escaping () -> ListSectionController) {
        self.value = value
        self.constructor = constructor
    }
}
`
then 
`
public protocol ListSwiftAdapterDataSource: class {
    func values(adapter: ListSwiftAdapter) -> [ Value<ListSwiftDiffable> ]
}
`

rnystrom:
@sphilipakis hmm sorry I don't think I follow. Why would changing to a `class` matter? Just tried it locally, same error.

> Are you sure that Person conforms to ListSwiftDiffable in your tests?

Yup! It compiles and works just fine w/ different API designs.

HarshilShah:
@rnystrom Hmm so that implementation seems to work with a couple of changes.

```swift
class IntSection: ListSwiftSectionController<Int> { }
class BoolSection: ListSwiftSectionController<Bool> { }

func values(adapter: ListSwiftAdapter) -> [Value] {
    return [
        Value(value: true, constructor: { () -> ListSwiftSectionController<Bool> in
            return BoolSection()
        }),
        Value(value: 1, constructor: { () -> ListSwiftSectionController<Int> in
            return IntSection()
        })
    ]
}
```

I ran this in a Playground, and it doesn't throw any warnings 🤔

The simple `.map` won't work here since you need to ensure that the `$0`'s type actually matches that of the `ListSwiftSectionController<T>` in order to form the `Value`.

sphilipakis:
@rnystrom yeah was solving another issue actually :) 
@HarshilShah is right, the `.map` won't work without some explicit typing like:

```swift
func values(adapter: ListSwiftAdapter) -> [Value] {
    return values.flatMap {
        if let person = $0 as? Person {
            return Value(value: person , constructor: { PersonSectionController() })
        } else if let company = $0 as? Company {
            return Value(value: company , constructor: { CompanySectionController() })
        }
        return nil
    }
}
```

HarshilShah:
@sphilipakis Yup, that's the right idea. It's worth noting however that `flatMap`-ing also doesn't help much in catching the lack of a corresponding section controller for a given model, and it really can't.

We're kinda going outside of the compiler's domain here; there's no way to enforce that for every `T: ListSwiftDiffable` you have created and pattern match against a corresponding `ListSwiftSectionController<T>`.

This mostly comes down to best practises recommendations in the library docs/example code, and I see two ways to deal with this:

1. Recommending adding `asserts` so that the `values` and the returned `[Value]` have the same `count`

    ```swift
    func values(adapter: ListSwiftAdapter) -> [Value] {
        let modelValues: [ListSwiftDiffable] = values
        let boxedValues: [Value] = /// flatMap business
        assert(modelValues.count == boxedValues.count) 
        return boxedValues
    }
    ```

1. This one is more opinionated and requires some extra code so I'm not sure it would work for everyone, but it's close to something I've done in my own projects so I thought I'd include it.

   We could wrap all the possible models for a given view controller inside an enum declared within the same VC, like so:

    ```swift
    enum DiffableValues: ListSwiftDiffable {
        case integer(Int)
        case boolean(Bool)
        
        func boxed() -> Value {
            switch self {
            case .integer(let value): return Value(value: value, constructor: { return IntSection() })
            case .boolean(let value): return Value(value: value, constructor: { return BoolSection() })
            }
        }
    }
    ```

    All possible sections you can display are grouped together this way, and we obtain a clear, non-failing mapping between model values and section controllers. Any new additions require adding a new enum case, at which point you also need to add the corresponding `boxed()` implementation.

    The only additional work aside from this is that instead of an array of vanilla `ListSwiftDiffable`s, users must wrap them in the appropriate `DiffableValue` to create an array of those instead.

    The call site can now be modified as follows:

    ```swift
    func values(adapter: ListSwiftAdapter) -> [Value] {
        let values: [DiffableValues] = /// D.Y.T.
        return values.map { $0.boxed() }
    }
    ```

   With this you get the assurance that every single model value will be displayed with the appropriate section controller.

    Again, I recognise that this isn't as simple as the original, but I just thought it was worth including given the safety gains.

rnystrom:
@HarshilShah ha! We arrived at the same conclusion, that's great. I got this working basically the same way:

```swift
func values(adapter: ListSwiftAdapter) -> [Value] {
  let constructor: () -> (ListSwiftSectionController<Person>) = { PersonSectionController() }
  return values.map { Value(value: $0, constructor: constructor) }
}
```

I see your point about pattern matching. If instead of assembling your data in the data source call:

```swift
func values(adapter: ListSwiftAdapter) -> [Value] {
  var values = [Value]()
  // append based on different states and stuff
  return values
}
```

([real example](https://github.com/rnystrom/GitHawk/blob/master/Classes/Issues/IssuesViewController.swift#L352-L402))

Then even if we return the value+controller tuple, you still have to pattern match.

```swift
let values: [ListSwiftDiffable] = // ...

func values(adapter: ListSwiftAdapter) -> [Value] {
  return values.map {
    switch $0 {
      case is String: return Value($0, { StringSectionController() }
      case is Int: // ...
  }
}
```

In that case, I see a couple choices for moving forward:

1. Keep drilling w/ the goal of making a comprehensive and type-safe API at the expense of API complexity
2. Mimic the `IGListAdapterDataSource` API by splitting `values()` and `sectionController(for value:)` into two functions
3. Keep the tuple pattern, just w/out any type safety

I lean towards 3 since that explicitly pairs the data type w/ the section controller. What does everyone else think?

HarshilShah:
Personally I'd say if the two methods are going to be coupled into one, it's better to be somewhat type-safe at least in terms of ensuring that the sectionController is compatible with the corresponding `ListSwiftDiffable`.

It could be explained as a composition of the existing two `objects(adapter:)` and `sectionController(object:)` methods.

(I'm using Xcode 9.3/Swift 4.1 so `flatMap` is renamed to `compactMap` in the code below)

```swift
func values(adapter: ListSwiftAdapter) -> [Value] {
        
    // Add stuff to `values` as in the existing `objects(adapter:)` method
    let values = [ListSwiftDiffable]()
    
    // Mapping, type-safe, and with little change vs. `sectionController(object:)` method
    return values.compactMap { value in
        if let value = value as? Bool {
            return Value(value: value, constructor: { return BoolSection() })
        } else if let value = value as? Int {
            return Value(value: value, constructor: { return IntSection() })
        }
        
        print("Couldn't find a match for \(value)! Better check this")
        return nil
    }
}
```

This ensures that users never end up with incompatible `object`/`sectionController` pairs, while also leaving a clear point for them to log/debug any holes in their own pattern matching (the `print` and `return nil` at the end).

Absolutely agreed about the enum bit; it's too complex for most and there's no way it could even be enforced. I just suggested it to show that the generics `Value` implementation allows for more type-safe code with relative ease.

rnystrom:
cc @jessesquires on some brainstorm ideas ☝️

rnystrom:
Last real TODO IMO is bridge `IGListBindingSectionController`. I'd like to use the same constructs w/ pairing and strong typing to bind models to cells. Might be trickier, but I'll try.

HarshilShah:
This is looking really nice!

Yeah map/flatMap were confusing at first until I realised flatMap "flattens" the input *and* output as well. `b = a.map { }` will always produce an array of the same size as `a` and make it an array of optionals if you return nil even once, whereas `flatMap` will remove all nil values entirely and return an array of unwrapped<T>.

Re naming: `ListSwiftData<something>` might work well, perhaps? Especially since the dataSource is `ListSwiftDataSource`.

alexito4:
FYI that confusion why flatMap is being renamed ;) https://github.com/apple/swift-evolution/blob/master/proposals/0187-introduce-filtermap.md

rnystrom:
Another proposal re: binding...

**What if we eliminate `ListSwiftSectionController` in favor of _requiring_ binding?**

Simply, we enforce the IGListKit+Swift bridge to use a Swift-compatible `IGListBindingSectionController` and nothing else, so you _have_ to decompose your model into view models, and you have to make your cells view-model-binding.

The inspiration for this is part laziness on me. Making `ListSwiftData` strong typing work w/ `ListSwiftSectionController` **and** `ListSwiftBindingSectionController` might be impossible w/out rewriting the binding SC to subclass `ListSwiftSectionController`. At that point, we're getting pretty complicated.

Maybe not? Would love some thoughts.

I kind of think making the Swift-bridge take a more opinionated approach (similar to the `ListSwiftData`+typing API) would be better, so long as we don't restrict what people can do today w/ IGListKit.

btw @HarshilShah I dig the name `ListSwiftData`, will update.

HarshilShah:
Hmm, that's a pretty big decision, but I'm all for binding controllers; those are the only ones I've written since they were made available.

One potential problem: This would reintroduce the same issue as discussed for `Hashable` above, however, in that the default implementation of `isEqual:` which falls back to `Equatable` isn't what's needed; [ListBindingSectionController](https://instagram.github.io/IGListKit/Classes/IGListBindingSectionController.html) requires `isEqual:` to always be `true`.

Perhaps there could be a separate protocol instead to handle top-level types (`ListSwiftSection`, maybe?) which only required `diffIdentifier()` for diffing?

One potential benefit: The diffing algorithm could also probably be made way more efficient if it never called `isEqual:` (imagine all the saved branches!) and just used the identifier.

rnystrom:
@HarshilShah ya we could enforce that the "top level" models are just identifiable (bringing back `ListSwiftIdentifiable` + `ListSwiftEquatable`). That would actually be pretty great, b/c then you can't screw it up. We can have two different boxes:

- One that wraps identifiable and always returns `true` for `isEqual(...)`
- One that wraps identifiable+equal to do equality-based diffing

I'm actually struggling to see how we can make a world where `ListSwiftSectionController<T>` will ever work cleanly w/ a binding section controller...

rnystrom:
Update after a bunch of hacking:

- I totally rebuilt `IGListBindingSectionController` in Swift
- I reinstated `ListSwiftIdentifiable` and `ListSwiftEquatable` so that binding SC models only implement identifiable bits (equal is always `true`)
- API is now all overrides so that we can take more advantage of generics (vs protocols can't use generics, and assoc type is redundant)
- Instead of returning an array of view models, dequeued cells, and sizes independently, you now return a data object w/ a model, cell type, and size function

The great part about this is that section controller implementation boils down to just this

```swift
class PersonSectionController: ListSwiftSectionController<Person> {

    override func createViewModelData(value: Person) -> [BindingData] {
        return [
            bindingData(value, cellType: PersonCell.self, size: { (context) -> CGSize in
                return CGSize(width: context.collectionContext.containerSize.width, height: 55)
            })
        ]
    }

}
```

One downside is that you still have to check the model type in the cell:

```swift
func bind(value: ListSwiftDiffable) {
    guard let value = value as? Person else { return }
    label.text = value.name.capitalized
}
```

Not sure I'm too offended by that.

HarshilShah:
Nice, really like that the `dequeue` logic is all abstracted away as well!

Not sure if the binding checks can be enforced on the cell anyway so that's okay.

rnystrom:
Ya I dig it. We'll need to provide a `didDequeue(cell: UICollectionViewCell & ListSwiftBindable)` hook so that ppl can do extra configuration, add delegates, etc. But I'm really feeling this direction.

Going to keep playing w/ ways to not have to rebuild `IGListBindingSectionController`. Or maybe at its core, share most of the logic in there. Otherwise the two controllers could diverge and maintenance will be a nightmare.

Bear in mind Instagram still doesn't use Swift, so any of this stuff we're building wont get the stress-testing of our hundreds of millions of people.

AndrewSB:
I just integrated with this, three things I didn’t love:
1. Conforming to all the protocols with `Swift` in the name felt like the swift version was tacked on top of the library, instead of carefully thought out and built. What would you think of hiding all the IG prefixed symbols and only renaming all the Swift symbols to non-prefixed versions, with no `Swift` in the name either.
2. I got burned by the default implementations of Equatable and Hashable - I created 3 instances of the same struct (with all the instance variables set to identical values), and the adapter decided to only render one, it assumed they were all the same. They were mocked models (my real models will probably have UUIDs, from the server).
Not sure if this is a swift specific problem, or just IGListKit in general, since this is my first time using it
3. I wasn’t a huge fan of the returning pairs syntax, but I couldn’t think of better pattern

rnystrom:
@AndrewSB **thank you** for taking the time to test this out and provide us with feedback!

1. I totally agree. I think we're pass the point where we can restore the `IG` prefix in Swift. Also you'll still interact w/ other core IGListKit components (unless/until everything is converted). I'll workshop some naming schemas and report back.
2. Good to know, I think I might totally remove those and just let people supply their own identifiers and equals methods. Though the concept of using 3 of the _exact same_ models wont work since they will be mapped to the same section controller. This is a longstanding constraint that I don't have a great answer to.
3. It's a little different, but I'll take the safety over syntax. I'm going to do some building w/ this over the weekend, might come up w/ some new ideas.

AndrewSB:
This is also my first time using IGListKit - so re: 2, if I had used objc IGListKit maybe I would have already been aware of that. 

Sounds good @rnystrom 😄

AndrewSB:
@rnystrom I noticed an issue with this API: It's not possible for `ListSwiftSectionController`'s to conform to `FooterSupplementaryViewSource`, since the section controller is generic (and `FooterSupplementaryViewSource` requires @objc classes, which cannot be swift generics).

I had issues with supplementary views in general, I'm not sure if the issue is related to the new swift classes, or something I'm doing wrong, tracking in #1130.

AndrewSB:
Not tracking in #1130 anymore 😛
Let me know if I can provide any more details!

MarvinNazari:
@AndrewSB: `supplementaryViewSource` is `weak` so you need to have a strong reference to `FooterSupplementaryViewSource`.

This should fix your issue:
```swift
    var footerSource: FooterSupplementaryViewSource?
    
    override init() {
        super.init()
        footerSource = FooterSupplementaryViewSource(collectionContext: collectionContext!, section: self)
        self.supplementaryViewSource = footerSource
    }
```

AndrewSB:
I see, that makes sense. Should a section's `supplementaryViewSource` be weak? That diverged from the UIKit norm I'm used to

AndrewSB:
I’m not a fan of the new master - from my cursory look it doesn’t work with Nibs/Storyboards

rnystrom:
@AndrewSB great point, I’ll make sure that’s covered in the first release 

rnystrom:
[Huge update](https://github.com/Instagram/IGListKit/pull/1081/commits/79689b76fe4e18c2f749cfadd7fb33b3ebedc75b). I'm really excited about some recent developments! You can download a [playground app here](https://github.com/Instagram/IGListKit/files/1981078/IGListKitExperimental.zip).

### Optional Binder Methods

I added some optional, **typed** closures to the binder models so you can do:

- Cell configuration
- Selection/deselection
- Highlight/unhighlight

Why? Previously to handle cell selection you'd have to do something like this:

```swift
override func didSelectItem(at index: Int) {
  if let model = viewModels[index] as? Person {
    didSelect(person: model)
  } else if model = viewModels[index] as? Animal {
    didSelect(animal: model)
  }
}
```

This lets bugs creep in as you add new models, refactor, etc. Not to mention the confusing separation between model+cell and selection.

Now, when you create a binder you can provide a selection _closure_ that has type information.

```swift
override func createBinders(from value: Person) -> [ListBinder] {
  return [
    binder(value, cellType: ListCellType.class(LabelCell.self), size: { (context) -> CGSize in
      return CGSize(width: context.collection.containerSize.width, height: 55)
    }, didSelect: { (context) in
      // note that the context.value is typed!
      print("Selected first cell with name \(context.value.name)")
    })
  ]
}
```

I'd encourage people to write methods like this:

```swift
func  didSelect(person: Person) {
  // ...
}

// in binder creation:
didSelect: { [weak self] (context) in
  self?.didSelect(person: context.value)
}
```

Pretty cool!

### Removing `ListSwiftBindable`

`IGListBindingSectionController` has a concept of auto cell-binding. While this is great, similar to selection above, you must check the type of the model before binding. It also forces your views to know about the type of the model. IMO this isn't a bad pattern (esp. w/ immutable models), but its a little too opinionated of a design for some.

Instead, we can use the binder models and generics to know the model _and_ cell type. This is all provided for free since the cell type information is required with the `cellType` param.

```swift
binder(value, cellType: ListCellType.class(LabelCell.self), size: { (context) -> CGSize in
  return CGSize(width: context.collection.containerSize.width, height: 55)
}, configure: { (cell, context) -> Void in
  // NOTE: both the cell and model are already typed!
  cell.label.text = context.value.name.capitalized
})
```

You would also use this method to pass any extra state to the cell, or wire up things like delegates.

```swift
configure: { [weak self] (cell, context) -> Void in
  cell.label.text = context.value.name.capitalized
  cell.delegate = self
}
```

### Other Stuff

- Provide a `ListCellType` to drive where to dequeue the cell from.
- Cleaned up dead code
- Some renaming in `ListSwiftSectionController` so that all view-model data is now wrapped up in "binder" models.
- `fatalError()`s now have descriptions

----

# Finalizing the API

Some things that I want to do before finalizing this API. **I need your help!**

- [ ] Finish and ship GitHawkApp/GitHawk#1743 using this branch
- [ ] Get more signoff on naming
- [ ] Document unsupported features at first launch (e.g. supplementary views)
- [ ] Unit tests
- [ ] Header documentation
  - Or maybe not? Swift is pretty self-documenting. Thoughts here?

### Naming

I've seen some discussion around the naming here. I'm not a fan of the `ListSwift*` either. However, since all of the `IGListKit` classes will naturally be available in this `IGListKit+Swift` extension, I'm not sure of a great alternative. Does anyone have any suggestions?

Here's the current list of types included in this extension:

- `ListDiffableBox` internal class
- `ListSwiftAdapter` class
- `ListSwiftAdapterDataSource` protocol
- `ListSwiftAdapterEmptyViewSource` protocol
- `ListSwiftIdentifiable` protocol
- `ListSwiftEquatable` protocol
- `ListSwiftDiffable` protocol
- `ListSwiftPair` class
- `ListCellType` enum
- `ListBinder` struct
- `ListSwiftSectionController` class
  - `State` private enum
  - `Context` struct

rnystrom:
Maybe a more experimental API, but this lets you just use the `Context` object to access a typed cell, and call select/deselect methods w/out all the params. Open to feedback on this one.

Using this in GitHawk and it lets me do stuff like this in the `didSelect` closure:

```swift
didSelect: { [weak self] context in
  guard let strongSelf = self else { return }
  strongSelf.selected = !strongSelf.selected

  context.deselect() // no params
  context.cell?.setSelected(strongSelf.selected) // cell is typed!
}
```

rnystrom:
One more thing I'm trying to find a way around is the missing type in the equality method. Currently you have to check the type:

```swift
func isEqual(to value: ListSwiftDiffable) -> Bool {
  guard let value = value as? MyModel else { return false }
  return name == value.name
}
```

jbouaziz:
Could you pass a `Self` value in the equality protocol?

Something in the lines of:
```func isEqual(to value: Self) -> Bool```

I haven’t tested it though.


Jonathan Bouaziz
________________________________

SirensOfTitan:
This is pretty wonderful, would love to see this land into master sometime soon.  

One thing I noticed while playing: it would be nice if the box functions were not `internal` so that we could port over existing applications progressively (i.e. not all at once).

rnystrom:
@SirensOfTitan great idea!

At this point, I'm feeling pretty comfortable about this change. We shipped it in GitHawk and things are going great. The biggest issue now is documentation and tests (there are none 😭). Might take a bit for us to get those added.

I'll rebase this branch so its up to date. You should be able to start using this branch to see how it works for you!

SirensOfTitan:
@rnystrom: Cool!  Happy to dogfood and potentially write documentation as we integrate the swift interop into our codebase.  Will start once the rebase happens. 👍

koenpunt:
This looks promising. One though I had, can't you just use `ListSwiftDiffable` as `ListDiffable`, and remove the `NS_SWIFT_NAME` macro from `IGListDiffable`? Same applies other types of course.

GitMAM:
Thanks very much! I am using it with Apollo GraphQL structs and has been working great so far. No issues!

rnystrom:
Quick update: I’m leaning towards making this an _**unofficial**_ library instead of being part of IGListKit core. Reason being that I’m more and more concerned about official support since we won’t use this in Instagram anytime soon. 

It’s an awesome layer that makes IGListKit in Swift _even better_, but I want to set the appropriate expectations.

SirensOfTitan:
> Quick update: I’m leaning towards making this an unofficial library instead of being part of IGListKit core. Reason being that I’m more and more concerned about official support since we won’t use this in Instagram anytime soon.

As I mentioned, happy to help maintain.  I'm also former FB of like 5 years so you can probably trust me 😛 .  Obviously pursuant to my availability, but I'm definitely willing to help support this add-on as a community member with others as I have time.   Really just need to get this to a good place for integration into our app (ListSwiftDiffable is ListDiffable comformant, public boxes, rebase) and then can start playing around with docs and bugs and all that.

rnystrom:
@SirensOfTitan appreciate that! I’ll give it more thought as I travel home this weekend and send another update. If you’re willing to help maintain I’m definitely more encouraged to make it an official part of the library 😍

ermik:
image
but `LabelCell` inherits from `UICollectionViewCell`! Why is this happening to me?

ermik:
I was playing with AsyncDisplayKit and it seems a bit hard to work with the swift methods and guess where in `createBinders` the call into ASDK's `cellForItemAtIndex` method (meant for IGListKit's needs) would go. I guess when it comes to integrations we're better off using Objective-C API?

rnystrom:
@ermik definitely since this branch isn’t really battle tested

RamblinWreck77:
Just found this! @rnystrom is it still your intention to keep this outside the main library? This stuff looks awesome!

ManWithBear:
Any update so far on Swift version?

koenpunt:
I think the future of IGListKit for swift is not so sure Or necessary anymore, now that as of Xcode 11/iOS 13 UICollectionView natively supports diffable datasources.

rnystrom:
Closing this as I've been experimenting w/ an all-swift rewrite that is a _much_ better direction. Hoping we can open source this soon!

levi:
Tho closed, still tremendously valuable. For anyone looking to use this in their own implementation, I have discovered a significant bug in `SwiftListSectionController`. Newly diffed updates to cells fail to update the correct cell. This is because `func update(animated: _)` enumerates through the indexes of the diffed result, not the actual indexes in the `NSIndexSet`.

Line 160 of `Source/Swift/ListSwiftSectionController.swift` should be changed from this:
```for (i, _) in result.updates.enumerated() {```
to this:
``` for i in result.updates {```

yardiboy:
Good

yardiboy:
Good

Marcocanc:
> Closing this as I've been experimenting w/ an all-swift rewrite that is a _much_ better direction. Hoping we can open source this soon!

Hey all, very excited to announce that we have just open sourced [SectionKit](https://github.com/traderepublic/SectionKit) which is an all-swift implementation of the same concept that IGListKit follows 🙌🏻

ermik:
Apparently I've been procrastinating on developing my app so long that an entire library has been deprecated

PR #1085, opened 2018-02-06
Generate 3.2 docs

rnystrom:
Does this need to wait for #1086?


docs/undocumented.json
```diff
@@ -1,12 +1,47 @@
 {
   "warnings": [
```
jessesquires: Can we add these missing docs to get 100%?

(or add `/// :nodoc:`)

rnystrom: On it

Issue #999, opened 2017-11-01
IGListBindingSectionController updateAnimated have copy BUG


YasinZhou:
IGListBindingSectionController.m
- (void)updateAnimated:(BOOL)animated completion:(void (^)(BOOL))completion{}
```
oldViewModels = self.viewModels;
```
and late
```
self.viewModels = [self.dataSource sectionController:self viewModelsForObject:object];
result = IGListDiff(oldViewModels, self.viewModels, IGListDiffEquality);
```

if `self.viewModels` is a NSMutableArray and just removeAllObjects and change the all value,
`result = IGListDiff(oldViewModels, self.viewModels, IGListDiffEquality);` is NOT right
so i think , oldViewModels need copy the self.viewModels([self.viewModels copy]) or self.viewModels's property replaces strong with copy.

and i think another ` updateAnimated ` have same BUG maybe

rnystrom:
Thanks for investigating @YasinZhou! We [deduplicate](https://github.com/Instagram/IGListKit/blob/master/Source/IGListBindingSectionController.m#L64-L66) the models returned from the data source, so even if a mutable array is returned, it should be totally safe.

You made me double check everything though, and I did come across [this](https://github.com/Instagram/IGListKit/blob/master/Source/IGListBindingSectionController.m#L117), and you're right that if the data source returns an `NSMutableArray`, they could mutate the array and corrupt the models stored inside the binding section controller.

That'd be a really easy and useful fix if you want to submit a PR!

rnystrom:
Good bug, but not world-ending. Leaving as a great starter task for someone!

KashishGoel:
@rnystrom I decided to take care of this. However, I'm not sure if this is something we need to write a test for. I don't think we need to write one for this change. Any thoughts on this?

PR #1109, opened 2018-02-22
Fix mutating array datasource bug

KashishGoel:
Copy objects when retrieving from datasource to account for the edge case where the returned data is a mutable array. 

Issue fixed: #999 

### Checklist


rnystrom:
@KashishGoel responding to the Test question here. Actually think a Test would be awesome for the longevity of this fix! Would be really easy:

- Create a mutable array
- Return array in binding SC method
- Mutate array
- Assert view models isn’t mutated

You can init a new SC without setting up the rest of the IGLK infra even.

KashishGoel:
@rnystrom Added a test, should be enough to ensure we don't run into this problem again.

rnystrom:
@KashishGoel its perfect 😇

Issue #1117, opened 2018-03-01
Section is invisible when numberOfItems is 0


aelam:
## New issue checklist

- [ x ] I have reviewed the [`README`](https://github.com/Instagram/IGListKit/blob/master/README.md) and [documentation](http://instagram.github.io/IGListKit)
- [ x ] I have searched [existing issues](https://github.com/Instagram/IGListKit/issues) and this is not a duplicate

### General information

- `IGListKit` version: 3.2.0
- iOS version(s): 11.0
- CocoaPods/Carthage version:  1.3.1
- Xcode version: 9.2
- Devices/Simulators affected: 
- Reproducible in the demo project? (Yes/No): Yes
- Related issues:

### Issue

I want to keep the header and footer while the numberOfItems is 0, But I've found that nothing shows when  numberOfItems is 0
Then I got this   **do not add headers if there are no items**  in `-layoutAttributesForElementsInRect:`

then I remove the condition `itemCount > 0`, it works just the same as it should be 
I also tried `UICollectionViewFlowLayout`,  work the same as the `IGListCollectionViewLayout` I modified 

So what's the reason of not showing the section when the number of items is 0

```objc
- (NSArray<UICollectionViewLayoutAttributes *> *)layoutAttributesForElementsInRect:(CGRect)rect {

        // ...
        // do not add headers if there are no items
//        if (itemCount > 0) {
            for (NSString *elementKind in _supplementaryAttributesCache.allKeys) {
                NSIndexPath *indexPath = indexPathForSection(section);
                UICollectionViewLayoutAttributes *attributes = [self layoutAttributesForSupplementaryViewOfKind:elementKind
                                                                                                    atIndexPath:indexPath];
                // do not add zero height headers/footers or headers/footers that are outside the rect
                const CGRect frame = attributes.frame;
                const CGRect intersection = CGRectIntersection(frame, rect);
                if (!CGRectIsEmpty(intersection) && CGRectGetLengthInDirection(frame, self.scrollDirection) > 0.0) {
                    [result addObject:attributes];
                }
            }
//        }

        // ...

}
```

rnystrom:
@aelam this is some behavior we have in Instagram to hide items. It's possible that we could revisit this though especially if that's how `UICollectionViewFlowLayout` works. I'd be happy to take a look at a PR!

marcuswu0814:
[...]

PR #1094, opened 2018-02-09
fixed Footer is sticky to the top

aelam:
if `stickyHeader` is true, the footer could be sticky to the top

## Changes in this pull request

Issue fixed: #1093

### Checklist

- [x] All tests pass. Demo project builds and runs.
- [x] I added tests, an experiment, or detailed why my change isn't tested.
- [x] I added an entry to the `CHANGELOG.md` for any breaking changes, enhancements, or bug fixes.
- [x] I have reviewed the [contributing guide](https://github.com/Instagram/IGListKit/blob/master/.github/CONTRIBUTING.md)

PR #1129, opened 2018-03-18
Show header when section item empty 

marcuswu0814:

Issue fixed: #1117

I adding a new constructor for making a `IGListCollectionViewLayout` instance that can always show sticky header although section data is empty.

It's working well and [this is demo project](https://github.com/marcuswu0814/IGListKit_ShowStickyHeaderWhenDataEmpty).

Bug I'm not sure is any good way to let this much more readability. Is there any good advice?

```objc
const CGRect headerBounds = (self.scrollDirection == UICollectionViewScrollDirectionVertical) ?
                CGRectMake(insets.left,
                        (itemCount == 0) ? CGRectGetMaxY(rollingSectionBounds) : CGRectGetMinY(rollingSectionBounds) - headerSize.height,
                        paddedLengthInFixedDirection,
                        headerSize.height) :
                CGRectMake((itemCount == 0) ? CGRectGetMaxX(rollingSectionBounds) : CGRectGetMinX(rollingSectionBounds) - headerSize.width,
                        insets.top,
                        headerSize.width,
                        paddedLengthInFixedDirection);
```

### Checklist

### DIFF
--- a/Source/IGListCollectionViewLayout.h
+++ b/Source/IGListCollectionViewLayout.h
@@ -129,6 +129,21 @@ NS_SWIFT_NAME(ListCollectionViewLayout)
                         stretchToEdge:(BOOL)stretchToEdge;
 
 /**
+ Create and return a new vertically scrolling collection view layout.
+ 
+ @param stickyHeaders Set to `YES` to stick section headers to the top of the bounds while scrolling.
+ @param showHeaderWhenEmpty Set to `YES` to show sticky header when a section had no item
+ @param topContentInset The top content inset used to offset the sticky headers. Ignored if stickyHeaders is `NO`.
+ @param stretchToEdge Specifies whether to stretch width of last item to right edge when distance from last item to right edge < epsilon(1)
+ 
+ @return A new collection view layout.
+ */
+- (instancetype)initWithStickyHeaders:(BOOL)stickyHeaders
+                  showHeaderWhenEmpty:(BOOL)showHeaderWhenEmpty
+                      topContentInset:(CGFloat)topContentInset
+                        stretchToEdge:(BOOL)stretchToEdge;
+
+/**
  :nodoc:
  */
 - (instancetype)init NS_UNAVAILABLE;
--- a/Source/IGListCollectionViewLayout.mm
+++ b/Source/IGListCollectionViewLayout.mm
@@ -150,6 +150,7 @@ static void adjustZIndexForAttributes(UICollectionViewLayoutAttributes *attribut
 @property (nonatomic, assign, readonly) BOOL stickyHeaders;
 @property (nonatomic, assign, readonly) CGFloat topContentInset;
 @property (nonatomic, assign, readonly) BOOL stretchToEdge;
+@property (nonatomic, assign, readonly) BOOL showHeaderWhenEmpty;
 
 @end
 
@@ -174,10 +175,22 @@ static void adjustZIndexForAttributes(UICollectionViewLayoutAttributes *attribut
 }
 
 - (instancetype)initWithStickyHeaders:(BOOL)stickyHeaders
+                  showHeaderWhenEmpty:(BOOL)showHeaderWhenEmpty
+                      topContentInset:(CGFloat)topContentInset
+                        stretchToEdge:(BOOL)stretchToEdge {
+    _showHeaderWhenEmpty = showHeaderWhenEmpty;
+    
+    return [self initWithStickyHeaders:stickyHeaders
+                       scrollDirection:UICollectionViewScrollDirectionVertical
+                       topContentInset:topContentInset
+                         stretchToEdge:stretchToEdge];
+}
+
+- (instancetype)initWithStickyHeaders:(BOOL)stickyHeaders
                       topContentInset:(CGFloat)topContentInset
                         stretchToEdge:(BOOL)stretchToEdge {
     return [self initWithStickyHeaders:stickyHeaders
-                       scrollDirection:UICollectionViewScrollDirectionVertical
+                   showHeaderWhenEmpty:NO
                        topContentInset:topContentInset
                          stretchToEdge:stretchToEdge];
 }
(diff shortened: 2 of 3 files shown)


Source/IGListCollectionViewLayout.h
```diff
@@ -128,6 +128,21 @@ NS_SWIFT_NAME(ListCollectionViewLayout)
...
+ @param showHeaderWhenEmpty Set to `YES` to show sticky header when a section had no item
+ @param topContentInset The top content inset used to offset the sticky headers. Ignored if stickyHeaders is `NO`.
+ @param stretchToEdge Specifies whether to stretch width of last item to right edge when distance from last item to right edge < epsilon(1)
+ 
+ @return A new collection view layout.
+ */
+- (instancetype)initWithStickyHeaders:(BOOL)stickyHeaders
+                  showHeaderWhenEmpty:(BOOL)showHeaderWhenEmpty
+                      topContentInset:(CGFloat)topContentInset
+                        stretchToEdge:(BOOL)stretchToEdge;
```
rnystrom: Could we actually just include this as a mutable `@property BOOL showHeaderWhenEmpty`? Trying to prevent some initializer creep here, ha!

Tests/IGListCollectionViewLayoutTests.m
```diff
@@ -86,6 +95,54 @@ - (void)test_whenEmptyData_thatContentSizeZero {
...
+                            [[IGLayoutTestSection alloc] initWithInsets:UIEdgeInsetsZero
+                                                            lineSpacing:0
+                                                       interitemSpacing:0
+                                                           headerHeight:20
+                                                           footerHeight:0
+                                                                  items:nil]]];
+    
+    IGAssertEqualFrame([self headerForSection:0].frame, 0, 0, 0, 0);
+    IGAssertEqualFrame([self headerForSection:1].frame, 0, 0, 0, 0);
+}
```
rnystrom: ❤️ tests!

Source/IGListCollectionViewLayout.mm
```diff
@@ -577,19 +590,23 @@ - (void)calculateLayoutIfNeeded {
...
-                CGRectMake(CGRectGetMinX(rollingSectionBounds) - headerSize.width,
+                CGRectMake((itemCount == 0) ? CGRectGetMaxX(rollingSectionBounds) : CGRectGetMinX(rollingSectionBounds) - headerSize.width,
                         insets.top,
                         headerSize.width,
                         paddedLengthInFixedDirection);
 
         _sectionData[section].headerBounds = headerBounds;
-
+        
+        if (itemCount == 0) {
```
rnystrom: Could we capture `itemCount == 0` in a variable?
```objc
const BOOL itemsEmpty = itemCount == 0;
```

Source/IGListCollectionViewLayout.mm
```diff
@@ -577,19 +590,23 @@ - (void)calculateLayoutIfNeeded {
                 rollingSectionBounds = CGRectUnion(rollingSectionBounds, frame);
             }
         }
-
+       
         const CGRect headerBounds = (self.scrollDirection == UICollectionViewScrollDirectionVertical) ?
                 CGRectMake(insets.left,
-                        CGRectGetMinY(rollingSectionBounds) - headerSize.height,
+                        (itemCount == 0) ? CGRectGetMaxY(rollingSectionBounds) : CGRectGetMinY(rollingSectionBounds) - headerSize.height,
```
rnystrom: What happens if `itemCount == 0` but `showHeaderWhenEmpty` is `NO`?

Source/IGListCollectionViewLayout.mm
marcuswu0814: Sure. 👍

Tests/IGListCollectionViewLayoutTests.m
marcuswu0814: 😄

Source/IGListCollectionViewLayout.h
marcuswu0814: Sure, I will roll it back and extract the property `showHeaderWhenEmpty ` to public.

Source/IGListCollectionViewLayout.mm
marcuswu0814: Nothing special, the `headerBounds` will same as `itemCount == 0`.

Before I added this condition, the `headerBounds`'s y position is out off screen when `itemCount == 0`. 

I think this changed did not affect any function because in method 

```objc
- (NSArray<UICollectionViewLayoutAttributes *> *)layoutAttributesForElementsInRect:(CGRect)rect {
        ...
        if (itemCount > 0 || self.showHeaderWhenEmpty) {
            for (NSString *elementKind in _supplementaryAttributesCache.allKeys) {
                NSIndexPath *indexPath = indexPathForSection(section);
                UICollectionViewLayoutAttributes *attributes = [self layoutAttributesForSupplementaryViewOfKind:elementKind
                                                                                                    atIndexPath:indexPath];
                // do not add zero height headers/footers or headers/footers that are outside the rect
                const CGRect frame = attributes.frame;
                const CGRect intersection = CGRectIntersection(frame, rect);
                if (!CGRectIsEmpty(intersection) && CGRectGetLengthInDirection(frame, self.scrollDirection) > 0.0) {
                    [result addObject:attributes];
                }
            }
        }
        ...
}
```

 the header will not be added into the result array.

So, should I modified this into `itemsEmpty && self.showHeaderWhenEmpty`? 😅)

marcuswu0814: Hum... I think I'm wrong, the UI will getting wrong layout. 

I will take a look later.
In my change and set `showHeaderWhenEmpty` to `NO`, try to set a section item count to zero and some error happed.

marcuswu0814:
I think the bug is not cause by my change.

The following demo is using [this project](https://github.com/marcuswu0814/IGListKit_ShowStickyHeaderWhenDataEmpty).

### In branch `IG/release`
Before any thing change, used `release` version install by current pod latest version:
ig release

**Both header and footer had some bug.**

### In branch `IG/master`

ig master

**The footer work well**, maybe fixed it #1094, but header got some bug like release version.

### In my change, branch `marcuswu0814/feature`

marcus0814 feature

If set property `showHeaderWhenEmpty` to `NO` and the header position will getting wrong.

So, In current `master` branch, if section item is empty and using sticky header, sticky header position will getting some bug.

marcuswu0814:
I fixed this bug on my branch latest commit, and the old test case didn't cover this bug, I already modified the test case and pass it. Now, avoid this bug! 😎

bugfix

Need open another issue to tracking this?
And if having any new issue, feel free to tell ms, thanks. 😅

Source/IGListCollectionViewLayout.mm
marcuswu0814: When I trying this, found another bug and note it below.

marcuswu0814:
Hi @rnystrom, will this possible be merged in 3.3.0 release?

PR #1138, opened 2018-03-26
Swift: Allow cells to define their own size

Marcocanc:
I thought it would be nice to have protocols that allow cells to provide their own size, so I introduced the following Protocols and a new initializer for `BindingData`:
- `ListSwiftSizeProviding`: Protocol for cells that can provide their own size based on context and value.
- `ListSwiftFixedHeightProviding`: Protocol for full width, fixed height cells.


Marcocanc:
That makes sense. Also it doesn't really bridge any IGListKit feature, so it probably shouldn't be there anyway.

PR #1205, opened 2018-06-29
Change paths to path

zhubofei:
Should map be a map of `IndexPath`s instead of arrays of `IndexPath`s? Also, we might need some unit tests on this part. I don't think it's covered. 

### Checklist


zhubofei:
@rnystrom Done

zhubofei:
@rnystrom @jessesquires After adding the test, I find this bug isn't that trivial😆. Should I add a changelog for it?

rnystrom:
@zhubofei yes please! I need to do a deeper audit to make sure this doesn't break anything haha.

lorixx:
@zhubofei Thanks for the fix, what I dont understand is that it seems like the current production code is not affected by this bug at all, do u have any theory why this is the case?

PR #1271, opened 2018-10-30
Update bundler to fix nokogiri vulnerability warning

rnystrom:
Ran `bundler update`. @jessesquires is this all that's required to clear vulnerability warnings?


BasThomas:
If it has resolved `nokogiri` to a non-vulnerable version, that should clear it (and it seems like it did).

PR #1330, opened 2019-06-13
added boundable

StevenArmandLee:

Issue fixed: #

problem background: 
majority of people who use IGListkit always have their sectionModel and sectionController depends on each other. 
- the need to have duplicated keys when differentSection controllers being used in different screens
- This will increase cyclomatic complexity as the more types of sectionControllers being used.
- prevent crash when forgetting to add the sectionController at func objects(for listAdapter: ListAdapter) -> [ListDiffable]

solution:
letting the sectionModel to return its sectionController

in depth example:
https://www.youtube.com/watch?v=lu-oSyEVWaA&t=187s 

### Checklist

- [x ] All tests pass. Demo project builds and runs.
- [x ] I have reviewed the [contributing guide](https://github.com/Instagram/IGListKit/blob/master/.github/CONTRIBUTING.md)


lorixx:
Hi Steven, thanks for adding the example! I have read the code thoroughly and the major change is by adding the < IGListBoundable> protocol. 

I am not sure that's the right direction we want to encourage people on how to use IGListKit, because one of the key principal is to encourage separation of concern: Section Model should just be a pure model object; and Section Controller can do other business logic and action handling etc. We shouldn't let Model to own the Section Controller. 

As a matter of fact, the number of Section Controller matches the number of section model, they are 1:1 mapping and managed by the IGListSectionMap data structure internally; besides, we can query the section controller through public API from IGListAdapter.

StevenArmandLee:
@lorixx I definitely agree with you. However after keep getting crashes and huge if else statements that result in cyclomatic complexing, one one to reduce this is by doing this. 

However I think of sectionModel as viewModel where you can put some logic in there instead of pure Model.

lorixx:
> However I think of sectionModel as viewModel where you can put some logic in there instead of pure Model.

I think there are other interpretation for ViewModel too: in MVVM pattern world, view model is also a pure class just containing the view-only data, like color or attributedString or layoutInfo. The business logic however should be handled by the SectionController which is mapping 1:1 to the ViewModel(from your use case). 

Would be great to learn where your crashes came from and how you ended up with giant if-else check and the cyclomatic complexity here, do you mind share some sample code? thanks

StevenArmandLee:
@lorixx 
here is some just some example of the code
`func listAdapter(_ listAdapter: ListAdapter, sectionControllerFor object: Any) -> ListSectionController {
//        if object is LabelSectionModel {
//            return LabelSectionController()
//        } else if object is DemoItem {
//            return DemoSectionController()
//        } else if object is EmptySectionModel {
//            return EmptySectionController()
//        } else if object is ActionLabelSectionModel {
//            return ActionLabelSectionController()
//        }
        if let object = object as? ListBoundable {
            return object.boundedSectionController()
        }
        return TitleValueSectionController()
    }`

so example in the commented sections, the list will grow as the more section controller need to be used. if we forget to add it in this function, it will go to the last section controller and where it will crash at `override func didUpdate(to object: Any) {
        //gonna crash here due to the wrong casting
    }`

so the top explains what case it will crash, to answer the cyclomatic complexity.

the more variety of section controller being used at one view controller, the bigger the cyclomatic complexity becomes, and there might be a need to copy paste this to other view controllers since the section model is reused through out the app.

the video here https://www.youtube.com/watch?v=lu-oSyEVWaA&t=187s might make it clearer on why I'm using this way.

StevenArmandLee:
@lorixx I assume this PR is rejected?

lorixx:
@StevenArmandLee Thanks for your detailed explanation and the talk! 
One thing you did raise a point on the cyclomatic complexity, which we also have a big if/elseif/else statement in inspecting the SectionModel, we solve the SectionController matching by adding an assertion to end of the checking. So if there is any missing type, the assertion will catch the mis-use. 

I think the main drawback here for `-boundedSectionController` is that it would encourage people to write mutable states to the ViewModel(a.k.a the SectionModel here). Most of the time, we want the view models to be immutable, which unlocks thread safety and we can generate ViewModels in a background queue, but if we add a `boundedSectionController` dependencies to the ViewModel, which not only adds more complexities to the ViewModel, but it makes generating ViewModels in background queue much unsafe.

Sorry to take a bit more time to review this, I definitely appreciate you are trying to help and improve the library. Let me know if adding the assertion in the SectionController mapping should practically solve your concern, thanks!

Issue #1341, opened 2019-07-16
Crash when interative reordering and destroying a collectionview at the same time


PhilCai1993:
## New issue checklist

- [x] I have reviewed the [`README`](https://github.com/Instagram/IGListKit/blob/master/README.md) and [documentation](http://instagram.github.io/IGListKit)
- [x] I have searched [existing issues](https://github.com/Instagram/IGListKit/issues) and this is not a duplicate
- [ ] I have attempted to reproduce the issue and include an example project.

### General information

- `IGListKit` version: 3.4.0
- iOS version(s): Found in iOS 12 device, guess it happens at all versions

For example, a collectionview is in a presented viewcontroller. When drag a cell, and dismiss the vc (the cell is still dragging) at the same time, it crash 100% at`-[UICollectionViewLayout+InteractiveReordering ig_cleanupInvalidationContext:]`. 

![image](https://user-images.githubusercontent.com/6828315/61277160-2b28ff80-a7e4-11e9-8ec8-6782642f92e1.png)

Because `UICollectionViewLayout` uses `OBJC_ASSOCIATION_ASSIGN` to hold `IGListAdapter`, and the real adapter is dealloced, EXC_BAD_ACCESS would happen.

One solution is using an intermediate object to hold the adapter, and `UICollectionViewLayout` hold a strong reference to the intermediate object.
![image](https://user-images.githubusercontent.com/6828315/61277520-ed78a680-a7e4-11e9-958f-c628f9869650.png)

![image](https://user-images.githubusercontent.com/6828315/61277508-e782c580-a7e4-11e9-985c-ed3adb2b2248.png)

Ziewvater:
Can you repro this issue? Without more info about the crash we can't address it. I'm going to close for now, but feel free to reopen the issue when you have a demo app that reproduces this crash.

I haven't seen any other cases where this seems to be an issue, so it seems to me more like that something else in your app that is causing this crash. It seems like you could just prevent the crash by preventing the user from dismissing the view controller while they're reordering items in the collection view.

PhilCai1993:
> Can you repro this issue? Without more info about the crash we can't address it. I'm going to close for now, but feel free to reopen the issue when you have a demo app that reproduces this crash.
> 
> I haven't seen any other cases where this seems to be an issue, so it seems to me more like that something else in your app that is causing this crash. It seems like you could just prevent the crash by preventing the user from dismissing the view controller while they're reordering items in the collection view.
[...]

PR #1342, opened 2019-07-16
Fix interactive reordering and destroying collection view at the same time

PhilCai1993:
Fix interactive reordering and destroying collection view at the same time

Issue fixed: https://github.com/Instagram/IGListKit/issues/1341#issuecomment-514055500

### Checklist

### DIFF
--- a/Source/Internal/UICollectionViewLayout+InteractiveReordering.m
+++ b/Source/Internal/UICollectionViewLayout+InteractiveReordering.m
@@ -12,6 +12,7 @@
 #import <IGListKit/IGListAssert.h>
 #import <IGListKit/IGListSectionController.h>
 #import <IGListKit/IGSystemVersion.h>
+#import <IGListKit/IGWeakObject.h>
 
 #import <objc/runtime.h>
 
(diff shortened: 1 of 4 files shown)


PhilCai1993:
Fix this #1341

PhilCai1993:
@bdotdub

PhilCai1993:
@lorixx https://github.com/Instagram/IGListKit/issues/1341#issuecomment-514055500 here is the explanation. It's hard to write unit test for this case because it requires mock gesture events, but it's easy to reproduce.

Source/Internal/UICollectionViewLayout+InteractiveReordering.m
```diff
@@ -60,7 +61,9 @@ + (void)load
 }
 
 - (void)ig_hijackLayoutInteractiveReorderingMethodForAdapter:(IGListAdapter *)adapter {
-    objc_setAssociatedObject(self, kIGListAdapterKey, adapter, OBJC_ASSOCIATION_ASSIGN);
```
lorixx: Maybe the problem here is the `OBJC_ASSOCIATION_ASSIGN`, can you try using the `OBJC_ASSOCIATION_RETAIN_NONATOMIC` ?  

IGListAdapter holds a weak reference to UICollectionView, so maybe we are fine here. Just make sure we dont get into a retain cycle issue, then we can avoid using a IGWeakObject here.

TimOliver:
Hey folks! I'm going through all the PRs to see which ones we can merge for the next version of IGListKit.

I agree with @lorixx's assessment that we can probably set the adapter to be strongly associated with the collection view since the adapter itself weakly references the collection view.

I'll try making the change and doing some memory profiling to confirm this is safe to do.

TimOliver:
Screenshot 2023-04-27 at 12 07 04

Okay, looks like strongly retaining the adapter from the collection view is fine. I was able to confirm in the sample apps that the collection views are properly deallocated once they leave the screen (Same for the list adapters too).

TimOliver:
Um. Okay. I just tried running the test suite on it and a variety of tests testing that the collection view and list adapter get destroyed without issue started failing. So, I just tested the repro steps listed in #1341 and it was working fine.

Since this PR is before the deallocation tests were added, I'm going to assume this has actually been resolved over time, and close the PR now.

@PhilCai1993 If you're still around, can you please test the latest version of IGListKit and see if the issue is still visible for you? If so, can you please open a new task with a video of the crash? Thanks a lot! Sorry we weren't able to get to your PR before.

PR #1386, opened 2019-11-17
use cocoapods cdn

koenpunt:

This updates the Podfile to use the specs CDN instead of the git checkout, which is more performant (and the default with CocoaPods 1.8 and higher).

### Checklist


Pods/Target Support Files/Pods-IGListKit-tvOSTests/Pods-IGListKit-tvOSTests-frameworks.sh
```diff
@@ -94,7 +94,7 @@ install_dsym() {
     binary="${DERIVED_FILES_DIR}/${basename}.framework.dSYM/Contents/Resources/DWARF/${basename}"
 
     # Strip invalid architectures so "fat" simulator / device frameworks work on device
-    if [[ "$(file "$binary")" == *"Mach-O dSYM companion"* ]]; then
+    if [[ "$(file "$binary")" == *"Mach-O "*"dSYM companion"* ]]; then
```
lorixx: What's the reason for all these script changes?

koenpunt: Since these are generated I guess that's because of the CocoaPods update.

koenpunt: Well, I'm not guessing, I know for sure.

lorixx: gotcha, so it's Pod generated script, looks good to me then!

PR #1398, opened 2019-11-22
Support Project Catalyst

jeffersonsetiawan:

Add support when compile Catalyst.
Fix this error: `Cannot define category for undefined class 'IGListAdapter'`
Screenshot 2019-11-22 at 22 35 31

## Step To Reproduce
In the Main target, check the Mac under Deployment Info, more Info in [here](https://developer.apple.com/documentation/uikit/mac_catalyst?language=objc)

### Checklist

### DIFF
--- a/Source/IGListKit/Internal/IGListAdapter+DebugDescription.h
+++ b/Source/IGListKit/Internal/IGListAdapter+DebugDescription.h
@@ -5,6 +5,7 @@
  * LICENSE file in the root directory of this source tree.
  */
 
+#import <IGListKit/IGListKit.h>
 #import <IGListKit/IGListAdapter.h>
 
 @interface IGListAdapter (DebugDescription)
--- a/Source/IGListKit/Internal/IGListAdapter+DebugDescription.h
+++ b/Source/IGListKit/Internal/IGListAdapter+DebugDescription.h
@@ -6,6 +6,7 @@
  */
 
 #import <IGListKit/IGListKit.h>
+#import <IGListKit/IGListAdapter.h>
 
 @interface IGListAdapter (DebugDescription)
 
--- a/Source/IGListKit/Internal/IGListAdapter+DebugDescription.h
+++ b/Source/IGListKit/Internal/IGListAdapter+DebugDescription.h
@@ -5,7 +5,6 @@
  * LICENSE file in the root directory of this source tree.
  */
 
-#import <IGListKit/IGListKit.h>
 #import <IGListKit/IGListAdapter.h>
 
 @interface IGListAdapter (DebugDescription)
--- a/Source/IGListKit/Internal/IGListAdapterUpdater+DebugDescription.h
+++ b/Source/IGListKit/Internal/IGListAdapterUpdater+DebugDescription.h
@@ -5,7 +5,6 @@
  * LICENSE file in the root directory of this source tree.
  */
 
-#import <IGListKit/IGListKit.h>
 #import <IGListKit/IGListAdapterUpdater.h>
 
 @interface IGListAdapterUpdater (DebugDescription)
--- a/Source/IGListKit/Internal/IGListAdapterUpdater+DebugDescription.h
+++ b/Source/IGListKit/Internal/IGListAdapterUpdater+DebugDescription.h
@@ -6,6 +6,7 @@
  */
 
 #import <IGListKit/IGListKit.h>
+#import <IGListKit/IGListAdapterUpdater.h>
 
 @interface IGListAdapterUpdater (DebugDescription)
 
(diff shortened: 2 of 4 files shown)


lorixx:
Thanks for adding this PR! @jeffersonsetiawan Do you mind update the description on how did u get that error? Wanna test out locally too, thanks!

lorixx:
Ah I am able to repro the compile error following this guide: https://developer.apple.com/documentation/uikit/mac_catalyst?language=objc

Source/IGListKit/Internal/IGListAdapter+DebugDescription.h
```diff
@@ -6,6 +6,7 @@
  */
 
 #import <IGListKit/IGListKit.h>
```
lorixx: Can you remove this line?

Source/IGListKit/Internal/IGListAdapterUpdater+DebugDescription.h
```diff
@@ -6,6 +6,7 @@
  */
 
 #import <IGListKit/IGListKit.h>
```
lorixx: ditto

Source/IGListKit/Internal/IGListBindingSectionController+DebugDescription.h
```diff
@@ -6,6 +6,7 @@
  */
 
 #import <IGListKit/IGListKit.h>
```
lorixx: ditto

lorixx:
Can you also remove the #import <IGListKit/IGListKit.h> in IGListBatchUpdateData+DebugDescription.h too? 

The reason the compilation failed is because we dont have IGListKit support for Mac OS, which doesn't have access to UIKit. The real solution might be to make IGListAdapter/SectionController to work with NSCollectionView etc. But that's probably a bigger project.

Keep in mind that MacOS can only access to IGListDiffKit for now: https://github.com/Instagram/IGListKit/blob/master/Source/IGListKit/IGListKit.h#L53

Source/IGListKit/Internal/IGListBindingSectionController+DebugDescription.h
jeffersonsetiawan: done

Source/IGListKit/Internal/IGListAdapterUpdater+DebugDescription.h
jeffersonsetiawan: done

Source/IGListKit/Internal/IGListAdapter+DebugDescription.h
jeffersonsetiawan: done

jeffersonsetiawan:
> Can you also remove the #import <IGListKit/IGListKit.h> in IGListBatchUpdateData+DebugDescription.h too?
> 
> The reason the compilation failed is because we dont have IGListKit support for Mac OS, which doesn't have access to UIKit. The real solution might be to make IGListAdapter/SectionController to work with NSCollectionView etc. But that's probably a bigger project.
> 
> Keep in mind that MacOS can only access to IGListDiffKit for now: https://github.com/Instagram/IGListKit/blob/master/Source/IGListKit/IGListKit.h#L53

Sure, I've update `IGListBatchUpdateData+DebugDescription.h` too.

Is the Catalyst that translate the UIKit world to the AppKit world? I don't know for sure, but in my project, it worked well.

lorixx:
Looks like the catalyst project was meant to make UIKit work on Mac, which is the whole point, I guess that should solve our problem!

jeffersonsetiawan:
The CI is failing, is it because of the removal of `#import <IGListKit/IGListKit.h>`?

lorixx:
Umm it seems very possible, check this: https://travis-ci.org/Instagram/IGListKit/jobs/615830215?utm_medium=notification&utm_source=github_status

Looks like unit test fails, u need to add back more imports.

jeffersonsetiawan:
> Umm it seems very possible, check this: https://travis-ci.org/Instagram/IGListKit/jobs/615830215?utm_medium=notification&utm_source=github_status
> 
> Looks like unit test fails, u need to add back more imports.

done :)

lorixx:
Sorry I didn't mean that you should add back the <IGListKit/IGListKit.h>, quite the opposite, can you add the missing imports in the test files which are failing? Again, we want to avoid using <IGListKit/IGListKit.h> internally which is causing issue. Thanks!

jeffersonsetiawan:
It back to green after I added `UIKit.h`

jeffersonsetiawan:
Yeayy, you’re welcome @lorixx !

Issue #1395, opened 2019-11-20
CI Improvements 


koenpunt:
The current CI configuration doesn't run the tests for all of the different platforms; it doesn't run tests for tvOS, it only builds the example application.

Running the tvOS tests locally from Xcode actually results in failures.

So I'd like to suggest to make the CI setup complete by fixing the tests for all platforms, and to run them all on CI.

Additionally, as suggested earlier in [a comment](https://github.com/Instagram/IGListKit/pull/1388#issuecomment-555523225), I'd see benefit in using a different CI provider than Travis, because Travis is slow to spin up VMs for example.
I did mention GitHub Actions, but I've just found out they only provide Xcode 11, which might be an issue, since the current suite runs with Xcode 10. CircleCI however does have Xcode 10.
With a move from Travis to CircleCI the project wouldn't be the first; the React Native project migrated to CircleCI a long time ago already.

WDYT?

lorixx:
> Running the tvOS tests locally from Xcode actually results in failures.

good catch! I think the tvOS example def needs some love.
Interested to learn more on the CircleCI stuff, is it faster than the Travis? I should at the very least, we should start looking into fixing the broken tvOS tests and use CI to run them for PRs, thanks for the suggestion here! cc @natestedman @bdotdub @Ziewvater @iperry90 too

Ziewvater:
How important is Xcode 10 support for us going forward? Xcode version support != iOS version support. Seems to me like we could probably get away with Github Actions if we wanted to.

The fact that other teams at FB have been using CircleCI probably means there's better internal support for it though, I'm going to look into that

koenpunt:
For reference; I already experimented a bit with GitHub actions on my fork (where you can see it failing because of an invalid destination, which is not present with Xcode 11); https://github.com/koenpunt/IGListKit/pull/1

But now that I mention it, it might actually be possible to pull in additional simulators if necessary.

koenpunt:
> But now that I mention it, it might actually be possible to pull in additional simulators if necessary.

It is possible to create additional simulator configurations, but we're still limited in SDK versions.

I've now also played around with [a configuration for CircleCI](https://github.com/koenpunt/IGListKit/pull/2), but there the problems is that I can't really test, since macOS runners are not enabled by default.

lorixx:
Hi @koenpunt , thanks again for flagging this issue! I think I can def feel the pain points for the queue time gets long whenever publishing new PRs. 

[...]

PR #1401, opened 2019-11-24
Run unit tests for tvOS

koenpunt:

As mentioned before in #1395, the unit tests are not executed for tvOS on the CI.

As you will see, they won't pass either.

### Checklist

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -122,6 +122,10 @@ jobs:
       name: "Tests: iOS 11.3"
       env: DESTINATION="$IOS_DESTINATION_11" WORKSPACE="IGListKit.xcworkspace" SDK="$IOS_SDK" SCHEME="$IOS_SCHEME"
 
+    - <<: *unit-tests
+      name: "Tests: tvOS 11.3"
+      env: DESTINATION="$TVOS_DESTINATION_11" WORKSPACE="IGListKit.xcworkspace" SDK="$TVOS_SDK" SCHEME="IGListKit-tvOS"
+
     # Danger
     - stage: Danger
       script: bundle exec danger --verbose


Tests/IGListSingleSectionControllerTests.m
```diff
@@ -75,14 +75,14 @@ - (void)test_whenItemUpdated_thatCellIsConfigured {
                              genTestObject(@3, @"Baz"),
                              ]];
     self.dataSource.objects = @[
-                                genTestObject(@1, @"Foo"),
-                                genTestObject(@2, @"Qux"), // new value
+                                genTestObject(@1, @"Qux"), // new value
+                                genTestObject(@2, @"Bar"),
                                 genTestObject(@3, @"Baz"),
```
koenpunt: I changed this so that the first object is updated instead of the second, because on tvOS `cellForItemAtIndexPath` doesn't return a cell for the second section, presumably because it's out of range.

Tests/IGListAdapterTests.m
```diff
@@ -41,7 +41,7 @@ - (void)setUp {
     self.adapter.collectionView = self.collectionView;
     self.adapter.dataSource = self.dataSource;
 
-    if (@available(iOS 11.0, *)) {
+    if (@available(iOS 11.0, tvOS 11.0, *)) {
```
lorixx: ummm, this is not ideal as this Test is highly optimized for testing for iOS. And the 2 test cases you moved are actually better coupled/co-located here within this file.

Plus, it adds a lot more maintenance cost in the future if people did not pay attention to some tests need to be put into the "XXX_IB" test files.

Tests/IGListSingleSectionControllerTests.m
lorixx: > because on tvOS cellForItemAtIndexPath doesn't return a cell for the second section, presumably because it's out of range.

Ummm this seems weird, it should return the right section data if the indexPath is correct, can you investigate a bit more why the test doesn't work for tvOS?

koenpunt:
> We'd better create a different set of Test files for tvOS.

How do you envision this? That we copy relevant test files to a `TVOSTests` folder or something? But then again, almost all tests pass for tvOS, so why would you want to duplicate tests?

Another thing I noticed, there are quite a few tests prefixed with `DISABLED_`, do we have to keep them?

lorixx:
> DISABLED_, do we have to keep them?

I think those are flaky tests and we have internal CI to auto-prefix with `DISABLED_` to disable the test, the solution should actually fix the broken fix and remove that prefix here.

> That we copy relevant test files to a TVOSTests folder or something? 

Nah, def not duplicate the test files here. If all the tests work for iOS target, we should have pretty good confidence that the code works for tv_os. I would rather we only add tv_os specific test cases instead of de-colocate tests in `IGListAdapterTests.m`.

Also from the optimization perspective, we should def focus more on iOS instead of tv_os, which had way less usage than the iOS use cases.  

Can you try add @available(iOS 11.0) for those test cases inline inside `IGListAdapterTests.m` ?

koenpunt:
> Can you try add @available(iOS 11.0) for those two test cases inline inside `IGListAdapterTests.m` ?

`@available` didn't work, but using the `TARGET_OS_TV` macro does.

Tests/IGListSingleSectionControllerTests.m
lorixx: @koenpunt did u figure out why this has to change?

Tests/IGListAdapterTests.m
```diff
@@ -284,6 +284,7 @@ - (void)DISABLED_test_whenCellsExtendBeyondBounds_thatVisibleSectionControllersA
     XCTAssertTrue([visibleSectionControllers containsObject:[self.adapter sectionControllerForObject:@4]]);
 }
 
+#if !TARGET_OS_TV
```
lorixx: _like_

Tests/IGListSingleSectionControllerTests.m
koenpunt: Not yet, I'll dive into this a bit more tomorrow.

koenpunt: The behavior of `cellForItemAtIndexPath` is just different for tvOS. When I change the frame of the collection view to be bigger than 100x100, even only slightly (100x105), it works. I yet have to see how this behaves when running in the tvOS simulator, but maybe that will make it clear why this is happening.

TimOliver:
Lemme just pipe up in here to say I'm looking at this. I'm actually surprised we didn't have the tests running on tvOS before.

There's some additional hurdles to handle now (Such as the test XIB files needing to be duplicated and modified for tvOS support), but I'd definitely like to get this set up when we can!

0jvi0 (approved):

TimOliver:
Hey folks! I imported a fresh copy of this branch in #1649 and landed it last night. I tested the whole suite as-is, without any of the special treatment changes here, and then gated out the ones that failed on tvOS.

Most of the ones I gated out involved a Storyboard or a NIB, since it'll be necessary to recreate tvOS-compatible versions of those. If anyone feels like they want those test cases for tvOS covered, please file a new PR.

The only non-NIB related test I had to gate was [this one](https://github.com/Instagram/IGListKit/blob/5784f6db5118566c4ce17f13d64c4490e5a4492c/Tests/IGListAdapterE2ETests.m#L592) as that one was throwing an exception stating the `UICollectionView` had entered a layout loop. This test works fine on iOS, so it seems to be caused by an implicit difference in behavior of UICollectionView on tvOS. If anyone would like to explore that one further, please also consider opening a new PR for that.

In any case, now that the test suites all run on iOS, macOS, and now tvOS, I think we can finally close this PR. Thanks again so much for your help @koenpunt!

PR #1408, opened 2019-12-16
Fix macOS 10.15 defaults cmd

RbBtSn0w:
On macOS Catalina when pod install on Example.
have an error at podspce file when run scrip "Version.sh"

defaults read is fail on macOS 10.15. not find Info

    filepath  Domains may also be specified as a path to an arbitrary plist file, with or without the '.plist' extension. For example:

                     defaults read ~/Library/Containers/com.apple.TextEdit/Data/Library/Preferences/com.apple.TextEdit.plist

               normally gives the same result as the two previous examples.  In the following example:

                     defaults write ~/Desktop/TestFile foo bar

               will write the key 'foo' with the value 'bar' into the plist file 'TestFile.plist' that is on the user's desktop. If the file does not exist, it will
               be created. If it does exist, the key-value pair will be added, overwriting the value of 'foo' if it already existed.

               WARNING: The defaults command will be changed in an upcoming major release to only operate on preferences domains. General plist manipulation utili-
               ties will be folded into a different command-line program.

So update to other cmd. Use plutil
Issue fixed: #

### Checklist

- [x] All tests pass. The demo project builds and runs.


RbBtSn0w:
same issues case at : https://github.com/flutter/flutter/issues/33909

lorixx:
Thanks for working on this!
Looks like CI is still failing here, do u think it's related to your change?

RbBtSn0w:
> change
No. the master branch still running fail.
it's was testing on macOS 10.15.

plutil API is long-time service on macOS

lorixx:
@RbBtSn0w Can you rebase onto master and update the PR again? Let's see if CI passed, I am happy to merge! Thanks again

RbBtSn0w:
rebase and push it.

TimOliver:
Thanks for the PR! The build script was updated in the interim to resolve this issue, so we can close this one off now. :)

PR #1430, opened 2020-02-25
Provided support for iOS 13 Context Menus

jjbourdev:

Issue fixed: Provided support for Context Menus

### Checklist


jjbourdev:
Not sure about what to do with Travis failing, as per the following PR [Fix the podlint error #1428](https://github.com/Instagram/IGListKit/pull/1428) it seems it's a known issue.

lorixx:
hey @jjbourdev , thanks for the PR! mind rebase from master and trigger the build again? thx!

Source/IGListKit/IGListSectionController.h
```diff
@@ -111,6 +111,18 @@ NS_SWIFT_NAME(ListSectionController)
...
+ Tells the section controller that the cell has requested a menu configuration.
+ 
+ @param index The index of the cell that requested the menu.
+ @param point The point of the tap on the cell.
+ 
+ @return An object that conforms to `UIContextMenuConfiguration`
+ 
+ @note The default implementation does nothing. **Calling super is not required.**
+ */
+- (nullable UIContextMenuConfiguration *)contextMenuConfigurationForItemAtIndex:(NSInteger)index point:(CGPoint)point API_AVAILABLE(ios(13.0)) API_UNAVAILABLE(tvos);
```
lorixx: Does this actually compile or work? I saw you have some lint error here: https://travis-ci.org/Instagram/IGListKit/jobs/655541844?utm_medium=notification&utm_source=github_status which is related to this file change.

jjbourdev: Yes it does, I don't really get why it complains here, it seems it wants the nullable to be apart from the return type.
I double checked on all samples, deintegrating / reinstalling all pods, it works pretty well, on multiple versions.

jjbourdev: Here's the local execution of the lint: 

```
$ machine:IGListKit jbourdev$ bundle exec pod lib lint IGListSwiftKit.podspec --allow-warnings "--include-podspecs=*.podspec"

 -> IGListSwiftKit (4.1.0)
    - NOTE  | xcodebuild:  note: Using new build system
    - NOTE  | xcodebuild:  note: Planning build
    - NOTE  | xcodebuild:  note: Constructing build description
    - NOTE  | xcodebuild:  warning: Skipping code signing because the target does not have an Info.plist file and one is not being generated automatically. (in target 'App' from project 'App')

IGListSwiftKit passed validation.
```

jjbourdev: Travis builds on Xcode 10.2.1 - 10E1001 while I locally build on Xcode 11.3.1 - 11C504. It shouldn't be an issue though.

jjbourdev: Same result if I do my setup the same as the CI... @lorixx any idea? :/

jjbourdev: @lorixx The error is different from a CI run to another... It compiles properly locally, on all targets, and for each example app. How could I solve the issue so this PR gets accepted?

Flatout73:
Hey, what about merge?)

jjbourdev:
I wish I can :)

Source/IGListKit/Internal/IGListAdapter+UICollectionView.m
```diff
@@ -235,6 +235,19 @@ - (void)collectionView:(UICollectionView *)collectionView didUnhighlightItemAtIn
     [sectionController didUnhighlightItemAtIndex:indexPath.item];
 }
 
+#if !TARGET_OS_TV
+- (UIContextMenuConfiguration *)collectionView:(UICollectionView *)collectionView contextMenuConfigurationForItemAtIndexPath:(NSIndexPath *)indexPath point:(CGPoint)point API_AVAILABLE(ios(13.0)) {
+  // forward this method to the delegate b/c this implementation will steal the message from the proxy
+  id<UICollectionViewDelegate> collectionViewDelegate = self.collectionViewDelegate;
+  if ([collectionViewDelegate respondsToSelector:@selector(collectionView:contextMenuConfigurationForItemAtIndexPath:point:)]) {
+      [collectionViewDelegate collectionView:collectionView contextMenuConfigurationForItemAtIndexPath:indexPath point:point];
```
AYastrebov, line 243: I think here should be `return [collectionViewDelegate collectionView:collectionView contextMenuConfigurationForItemAtIndexPath:indexPath point:point];`

AYastrebov (commented):
What about merging it to the master?
@lorixx any ETA?

Source/IGListKit/Internal/IGListAdapter+UICollectionView.m
jjbourdev: Nope it's a call to the delegate. 
I don't see anything blocking in this PR, apart from the fact it's getting old now :)

sjchmiela: My 2 cents: I too think this line is missing a `return` in the beginning. As far as I'm aware `collectionView:contextMenuConfigurationForItemAtIndexPath:point:`'s purpose is to **return** a context menu configuration, not simply to notify the delegate (see [docs](https://developer.apple.com/documentation/uikit/uicollectionviewdelegate/3295917-collectionview)).

If you do `return` here you would give the developer option to either add context menu configuration to `collectionViewDelegate` or to `sectionController`. If you do not return here there's no reason to implement this method in `collectionViewDelegate` since it wouldn't do anything.

At least I would have been surprised if I had implemented `collectionView:contextMenuConfigurationForItemAtIndexPath:point:` in `collectionViewDelegate`, returned proper configuration and it wouldn't cause context menu to show up.

Source/IGListKit/IGListBindingSectionControllerSelectionDelegate.h
```diff
@@ -61,6 +61,21 @@ NS_SWIFT_NAME(ListBindingSectionControllerSelectionDelegate)
...
+ Tells the delegate that a cell has requested a menu configuration.
+
+ @param sectionController The section controller the request of a menu configuration occurred in.
+ @param index The index of the cell that is being longed tap.
+ @param point The point of the tap on the cell.
+ @param viewModel The view model that was bound to the cell.
+ 
+ @return An object that conforms to `UIContextMenuConfiguration`.
+ */
+- (UIContextMenuConfiguration * _Nullable)sectionController:(IGListBindingSectionController *)sectionController
```
sjchmiela, line 74: I'm also not super familiar with `IGListKit`, but this seems like an unnecessary forwarding of the call — after all it's `IGListBindingSectionController**Selection**Delegate`, not a "general delegate" — why would a "delegate that receives selection events from cells" be interested in context menu events?

amalenduk:
Hi, Any update on this

TimOliver:
Hi folks! I'm so sorry for the amount of time it's taken to get back to this PR. I had a play with the PR in the sample app, and I gotta say this is a pretty dope feature!

I tried importing it into the Instagram codebase, and after a bit of review, I think we need to tweak it a bit. Namely, that new method in `IGListBindingSectionControllerSelectionDelegate` implicitly adds UIKit as a required import for the protocol, which might have ramifications down the line.

We might need to have a think about this, but it might be more appropriate to add a new property/protocol to `IGListBindingSectionController` to handle context menu configurations in its own silo separate from cell configuration. This also gives us a better space to add more of the context menu protocol methods down the line.

What do you think?

amalenduk:
Hi @TimOliver , Any update on this?

Asmodeusxxx85 (approved):

PR #1462, opened 2020-09-22
Github actions and Carthage users.

3a4oT:

- Fixed tv-OS build on Xcode 11.7 by adding  `IGListTransitionData.m` to compiled sources of tvOS target, `IGListTransitionData.h` , `IGListAdapterUpdaterCompatible.h` to project headers for tvOS target.

- Added GitHub workflow which can automatically create GitHub release(when the new tag was pushed) and attach a binary framework for Carthage users so they can pull it instead of building from source. (For example, you pushed new tag 5.0.0 - this step will automatically create a GitHub release and attach a binary framework.) . Command `carthage archive` - produce the artifact "IGListKitDiff.zip" (I guess because of sorting), but archive contains both binaries.

- Added Github workflow which verifies that Carthage is able to build IGListKit

- Workflows use Xcode 11.7

### Checklist


lorixx:
I think we are good with the Github Action now, thanks @3a4oT !

PR #1520, opened 2021-07-24
fix: sectionController never fresh like a deadlock

zyc666jdi:
when you create a subclass of IGListBindingSectionController,then fill with data and use 'update(animated: true, completion: nil)' when this secontroller 'listAdapter(_ listAdapter: ListAdapter, sectionControllerFor object: Any) -> ListSectionController' never return .  then 'listAdapter(_ listAdapter: ListAdapter, sectionControllerFor object: Any) -> ListSectionController' return this sectionController , and this sectionContrloller use 'update(animated: true' ,but it will never refresh because it's state is IGListDiffingSectionStateUpdateQueued and never update to IGListDiffingSectionStateIdle . so it's refresh method is blocked and never refresh like a deadlock

Issue fixed: #

### Checklist

### DIFF
--- a/Source/IGListKit/IGListBindingSectionController.m
+++ b/Source/IGListKit/IGListBindingSectionController.m
@@ -46,6 +46,9 @@ typedef NS_ENUM(NSInteger, IGListDiffingSectionState) {
     __block NSArray<id<IGListDiffable>> *oldViewModels = nil;
 
     id<IGListCollectionContext> collectionContext = self.collectionContext;
+    if (self.collectionContext == nil) {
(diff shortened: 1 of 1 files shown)


lorixx:
Thanks @zyc666jdi for working on this! Do you mind sign the CLA?

lorixx:
Can you write a unit test to demonstrate the issue that you describe here? Your IGList DataSource `IGListAdapterDataSource` supposed to return the right section controller here.

lorixx (changes requested):
Back to your queue to add unit test for the bug, thanks @zyc666jdi !

